PR #3, opened 2015-03-31
Justin/maven

sul3n3t:
R: @riyazdf


riyazdf:
:+1: looks good, Travis failures only due to MSHADE-183 (verified locally)

PR #4, opened 2015-03-31
Add Travis CI

riyazdf:
R: @sul3n3t 

Tests both the java code and ui through Travis.  There are a few changes and concerns:

- I had to slightly change a karma conf, because Travis is a little harder to setup with Chrome instead of Firefox.  Inspiration from: http://stackoverflow.com/questions/19255976/how-to-make-travis-execute-angular-tests-on-chrome-please-set-env-variable-chr

- I've temporarily reverted the shade plugin to a previously functioning version (before MSHADE-183).  On version 2.3, Travis fails the mvn install stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,15 @@ jdk:
 env:
   - UI_DIR=ui
 
-before_script:
-  - cd $UI_DIR && npm install -g grunt-cli bower && npm install && bower install && grunt build
+before_install:
+  - export CHROME_BIN=chromium-browser
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+  - npm install karma && npm install karma-jasmine karma-chrome-launcher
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,9 +6,16 @@ jdk:
 env:
   - UI_DIR=ui
 
-before_script:
-  - cd $UI_DIR && npm install -g grunt-cli bower && npm install && bower install && grunt build
+before_install:
+  - export CHROME_BIN=chromium-browser
+  - export DISPLAY=:99.0
```
sul3n3t: Can you comment in the config where the setup came from? Thanks for adding it to the PR, but it will be hard to find later.

riyazdf: Good idea, I'll add that in the next commit.

sul3n3t:
:+1:

PR #5, opened 2015-03-31
add metadata column for secrets_content

riyazdf:
R: @sul3n3t 

To be filled later, so that metadata is tied to individual secret versions instead of being shared across all versions (the metadata column in the secrets table will be dropped later)


sul3n3t:
:+1:

PR #6, opened 2015-04-01
Justin/maven

sul3n3t:
R: @riyazdf 

The attempted release failed because every module was getting shaded. For example, the keywhiz-parent module failed with an empty shaded jar.


riyazdf:
:+1:

PR #7, opened 2015-04-01
move all references for metadata from series to content

riyazdf:
R: @sul3n3t 

**Before merging**: will have to run db command to copy pre-existing metadata from secrets to secrets_content table, matching by secret ID: `UPDATE secrets_content SET metadata = secrets.metadata FROM secrets WHERE secrets.id = secretId;`

Changes:
- metadata column in secrets table is now nullable
- all reads and writes of metadata are done on the metadata column in secrets_content
- remove references to plaintext secret in SecretBuilder (column was dropped previously)

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/SecretController.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretController.java
@@ -115,7 +115,7 @@ public class SecretController {
     checkArgument(!secret.isEmpty());
     checkArgument(!creator.isEmpty());
     String encryptedSecret = cryptographer.encryptionKeyDerivedFrom(name).encrypt(secret);
-    return new SecretBuilder(transformer, secretDAO, name, secret, encryptedSecret, creator);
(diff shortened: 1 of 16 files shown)


server/src/main/java/keywhiz/service/daos/SecretController.java
```diff
@@ -136,16 +135,14 @@ public SecretBuilder builder(String name, String secret, String creator) {
      * @param transformer
      * @param secretDAO
      * @param name of secret series.
-     * @param secret content of secret version.
      * @param encryptedSecret encrypted content of secret version
      * @param creator username responsible for creating this secret version.
      */
-    private SecretBuilder(SecretTransformer transformer, SecretDAO secretDAO, String name, String secret, String encryptedSecret,
```
sul3n3t: Nice catch

sul3n3t:
:+1:

PR #8, opened 2015-04-01
remove last references before dropping secrets metadata column

riyazdf:
R: @sul3n3t 

Removes last references to the metadata column in the secrets table (unfortunately I missed them).

After this change, there will be no other references to secrets.metadata, and we can add the migration `ALTER TABLE secrets DROP COLUMN IF EXISTS metadata;` on the next version.


server/src/test/java/keywhiz/service/daos/SecretContentDAOTest.java
```diff
@@ -48,8 +48,8 @@ public void setUp() throws Exception {
     secretContentDAO = testDBRule.getDbi().onDemand(SecretContentDAO.class);
 
     testDBRule.jooqContext().delete(SECRETS).execute();
-    testDBRule.jooqContext().insertInto(SECRETS, SECRETS.ID, SECRETS.NAME, SECRETS.METADATA)
-        .values((int) secretContent1.secretSeriesId(), "secretName", "{}")
```
sul3n3t: Why no corresponding add to the content table?

riyazdf: Good catch, I'll go ahead and add one in the next commit.

sul3n3t:
:+1:

PR #9, opened 2015-04-01
Drop metadata column from secrets table

riyazdf:
R: @sul3n3t 

We can tag for a new release just before this commit -- this will finally drop the metadata column from the secret series table, thus binding metadata to individual versions of secrets.


sul3n3t:
:+1: I already cut 0.7.1

riyazdf:
Awesome, thanks!

PR #15, opened 2015-04-07
AutomationGroupResource returns all groups when name query missing

sul3n3t:
Updates AutomationGroupResource to match the javadoc comment and handle missing name query parameter.

R: @riyazdf


riyazdf:
:+1:

PR #17, opened 2015-04-07
update automation endpoint to list all secrets, update javadocs and test data

riyazdf:
R: @sul3n3t 

automation/secrets now lists all secrets regardless of ownership, cleaned up javadocs and test data (single versioned secret)


sul3n3t:
LGTM

PR #18, opened 2015-04-07
speedup keywhiz.cli list action

riyazdf:
R: @sul3n3t 

Before this change, keywhiz.cli was making unnecessary api calls when no options were provided.  This change explicitly removes options from the keywhiz.cli list command.


sul3n3t:
LGTM

PR #19, opened 2015-04-07
Alok/random fixes

alokmenghrajani:
* Fixes Security Headers
* Fixes dev config
* Adds default username/password to https://square.github.io/keywhiz/

### DIFF
--- a/server/src/main/java/keywhiz/KeywhizService.java
+++ b/server/src/main/java/keywhiz/KeywhizService.java
@@ -29,9 +29,11 @@ import io.dropwizard.jdbi.bundles.DBIExceptionsBundle;
 import io.dropwizard.jersey.setup.JerseyEnvironment;
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
+import java.util.EnumSet;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
+import javax.servlet.DispatcherType;
 import keywhiz.auth.mutualssl.ClientCertificateFilter;
 import keywhiz.auth.xsrf.XsrfServletFilter;
 import keywhiz.commands.MigrateCommand;
@@ -156,11 +158,14 @@ public class KeywhizService extends Application<KeywhizConfig> {
     logger.debug("Registering resource filters");
     jersey.register(injector.getInstance(ClientCertificateFilter.class));
 
-    logger.debug("Registering response filters");
-    jersey.register(injector.getInstance(SecurityHeadersFilter.class));
+    logger.debug("Registering servlet filters");
+    environment.servlets().addFilter("security-headers-filter", injector.getInstance(SecurityHeadersFilter.class))
+        .addMappingForUrlPatterns(
+            null, /* Default is for requests */
+            false /* Can be after other filters */,
+            "/*" /* Every request */);
     jersey.register(injector.getInstance(CookieRenewingFilter.class));
 
-    logger.debug("Registering servlet filters");
     environment.servlets().addFilter("xsrf-filter", injector.getInstance(XsrfServletFilter.class))
         .addMappingForUrlPatterns(
             null /* Default is for requests */,
--- a/server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
+++ b/server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
@@ -18,10 +18,13 @@ package keywhiz.service.filters;
 import com.google.common.net.HttpHeaders;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
-import javax.ws.rs.container.ContainerRequestContext;
-import javax.ws.rs.container.ContainerResponseContext;
-import javax.ws.rs.container.ContainerResponseFilter;
-import javax.ws.rs.core.MultivaluedMap;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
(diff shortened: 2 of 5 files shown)


alokmenghrajani:
I'll try to write some tests to make sure we don't have a regression in the future.

server/src/main/java/keywhiz/KeywhizService.java
```diff
@@ -157,7 +159,8 @@ public void addSecretGenerator(String name, SecretGenerator<?> secretGenerator)
     jersey.register(injector.getInstance(ClientCertificateFilter.class));
 
     logger.debug("Registering response filters");
-    jersey.register(injector.getInstance(SecurityHeadersFilter.class));
+    environment.servlets().addFilter("meh", injector.getInstance(SecurityHeadersFilter.class))
```
riyazdf: rename to "security-filter" or similar?

server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
```diff
@@ -36,24 +38,35 @@
  *   IE X-Content-Type-Options: http://msdn.microsoft.com/en-us/library/ie/gg622941(v=vs.85).aspx
  *   IE X-XSS-Protection: http://blogs.msdn.com/b/ie/archive/2008/07/02/ie8-security-part-iv-the-xss-filter.aspx
  */
-public class SecurityHeadersFilter implements ContainerResponseFilter {
+public class SecurityHeadersFilter implements javax.servlet.Filter {
```
riyazdf: include as import above instead of a static importing here?

website/index.html
```diff
@@ -170,7 +170,8 @@ <h3 id="setup">Setup</h3>
             <pre class="prettyprint">$ java -jar server/target/keywhiz-server-*-SNAPSHOT-shaded.jar server server/src/main/resources/keywhiz-development.yaml</pre>
 
             <p>Access the UI by opening
-                <a href="https://localhost:4444/">https://localhost:4444/</a> in a browser.</p>
+                <a href="https://localhost:4444/">https://localhost:4444/</a> in a browser. The development
```
riyazdf: Good idea to include this here! :+1:

sul3n3t:
Shoot, I thought there was an integration test for these headers. Take a look at `UiAssetsBundleTest` for a similar test setup.

server/src/main/java/keywhiz/KeywhizService.java
alokmenghrajani: fixed

server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
alokmenghrajani: Fixed

server/src/main/java/keywhiz/KeywhizService.java
```diff
@@ -157,7 +159,8 @@ public void addSecretGenerator(String name, SecretGenerator<?> secretGenerator)
     jersey.register(injector.getInstance(ClientCertificateFilter.class));
 
     logger.debug("Registering response filters");
-    jersey.register(injector.getInstance(SecurityHeadersFilter.class));
+    environment.servlets().addFilter("security-headers-filter", injector.getInstance(SecurityHeadersFilter.class))
+        .addMappingForUrlPatterns(EnumSet.of(DispatcherType.REQUEST), true, "/*");
```
sul3n3t: Should this be set for ERROR dispatch types too? Could you move this down under "Registering servlet filters" and use inline comments like XsrfServletFilter does?

server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
```diff
@@ -36,24 +39,35 @@
...
-    headers.add("Frame-Options", "DENY");
-    headers.add(HttpHeaders.X_FRAME_OPTIONS, "DENY");
+      // Protection against click jacking
+      r.addHeader("Frame-Options", "DENY"); // Who uses this?
+      r.addHeader(HttpHeaders.X_FRAME_OPTIONS, "DENY");
+
+      // https-all-the-time
+      r.addHeader(HttpHeaders.STRICT_TRANSPORT_SECURITY, format("max-age=%d; includeSubDomains", YEAR_OF_SECONDS));
+
+      chain.doFilter(request, response);
```
sul3n3t: Move this after the if condition

server/src/test/java/keywhiz/UiAssetsBundleTest.java
```diff
@@ -36,6 +36,14 @@
     client = TestClients.unauthenticatedClient();
   }
 
+  private void checkSecurityHeaders(Response response) {
```
sul3n3t: I guess I was thinking this would be a test called SecurityHeadersFilterTest just in the same style as UiAssetsBundleTest. That would make it discoverable when changes are made to SecurityHeadersFilter.

sul3n3t:
LGTM after the comments I just added. Looks like a rebase is needed to make the branch mergeable.

server/src/main/java/keywhiz/KeywhizService.java
alokmenghrajani: Right now, it triggers on 404 but not on 401. I tried adding ERROR and 401s did not get the header.

server/src/main/java/keywhiz/service/filters/SecurityHeadersFilter.java
alokmenghrajani: Fixed. https://dropwizard.github.io/dropwizard/manual/core.html#servlet-filters does the chain.doFilter inside the if statement.

PR #25, opened 2015-04-09
rename for clarity on apidocs

riyazdf:
R: @sul3n3t 

Slight renaming to make the apidocs clearer for deleting a secret on the automation endpoint by name.


sul3n3t:
:+1:

PR #27, opened 2015-04-09
add explicit exclusions to remove duplicate jars, remove warnings

riyazdf:
R: @sul3n3t 

Add exclusions to the pom so that we don't pick up duplicate jars, this removes the warnings from running mvn package


sul3n3t:
:+1:

PR #28, opened 2015-04-10
API for versioned secrets

riyazdf:
R: @sul3n3t 

Endpoints for versioned secrets:

GET /v2/secrets - gets a list of all secret names (will only list each name once, regardless of versions)
GET /v2/secrets/{name} - gets a list of all version names for this secret name
GET /v2/secrets/{name}/{version} - gets a SecretDetailResponse for this secret name/version combination.  Using "LATEST" as the version field will fetch the latest version of this secret

POST /v2/secrets - creates a versioned secret using the passed-in JSON

DELETE /v2/secrets/{name} - deletes all versions of this secret (the entire secret series)
DELETE /v2/secrets/{name}/{version} - just deletes this version of the secret


riyazdf:
this is addressed by #136

PR #29, opened 2015-04-13
add google group, change maven artifact search to https

riyazdf:
R: @sul3n3t 

Add google group link, also change maven link to https


sul3n3t:
:+1: after comment

PR #35, opened 2015-04-14
Makes sure we are testing json serialization/deserialization for Map<String, String>

alokmenghrajani:
Some Map<String, String> fields are stored as json by the DAO layer. When working on the alok/jooq branch, I noticed the tests were passing even though I hadn't implemented the json serialization/deserialization so I figured it might be worth testing for.


sul3n3t:
:+1:

riyazdf:
:+1:

PR #50, opened 2015-04-20
Prefer char[] for passwords on client

sul3n3t:
Using a char[] allows zeroing the password after use, whereas a String
change would just create a copy. There is still no guarantee the JVM has
not made copies of the char[], but it should stick around less.

R: @alokmenghrajani


alokmenghrajani:
I would make the change on the client side only or use a BCrypt library which won't require converting the char[] to a String.

sul3n3t:
@alokmenghrajani Update to only change client code, server tests using client code, or an API object from the client.

alokmenghrajani:
:+1:

PR #53, opened 2015-04-21
I think the default requires keywhizdb_test

barthoda:


alokmenghrajani:
keywhizdb_test is used by the tests. If you want some data to be populated when you hit https://localhost:4444/, you want to have the data in keywhizdb_development.

sul3n3t:
Yep, we consider the development environment to be where you can add data, play around with new features, etc and have it persisted between server starts/stops.

The test environment is just used when running tests. Tests are free to clear tables, rerun migrations, and load data. The `keywhizdb_test` database should be created and migrated automatically as part of building the `keywhiz-model` module.

barthoda:
I've been through this process only one time. I'm going to move into production with this in the next month or so. When I rebuild the environment a second time I'll let you know what i see,

PR #55, opened 2015-04-22
Changing IV to 12 bytes instead of 16 per NIST recommendation, renaming ...

joelthompson:
...it to a nonce

NIST recommendations at http://csrc.nist.gov/publications/nistpubs/800-38D/SP-800-38D.pdf, specifically printed page 8, states, "For IVs, it is recommended that implementations restrict support to the length of 96 bits, to promote interoperability, efficiency, and simplicity of design." This implements that recommendation. It also mostly renames "IV" to "nonce" to better align with the meaning. auth/cookie/GCMEncryptor.java already implements both these suggestions (calling it a nonce and making it 12 bytes).

### DIFF
--- a/server/src/main/java/keywhiz/service/crypto/ContentCryptographer.java
+++ b/server/src/main/java/keywhiz/service/crypto/ContentCryptographer.java
@@ -65,7 +65,7 @@ public class ContentCryptographer {
   private static final String ENCRYPTION_ALGORITHM = "AES/GCM/NoPadding";
   private static final String KEY_ALGORITHM = "AES";
   private static final int TAG_BITS = 128;
-  private static final int IV_BYTES = 16;
+  private static final int NONCE_BYTES = 16;
   private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
 
   private final SecretKey key;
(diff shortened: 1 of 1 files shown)


alokmenghrajani:
96 bits is what most other AES-GCM libraries use (e.g. our go-jose and js-jose libraries). As long as the content currently encrypted with 128 bits continues to decrypt this change seems fine.

I'm not a big fan of changing the wording from IV to nonce since most docs refer to this data as IV.

joelthompson:
Yes, 96 bits is what most other AES-GCM libraries use, because that's what NIST recommends :)

As far as continuing to support content encrypted with 16-byte IVs, I don't know the codebase well enough to comment thoroughly, but based on my reading of the code, IV_BYTES/NONCE_BYTES seems to ONLY be used in the encrypt function. I believe BouncyCastle just encodes the IV/nonce in the blob it returns when you encrypt with GCM so it shouldn't matter.

With regard to calling it a nonce/IV, given the criticality of not reusing an IV/nonce with a given encryption key when using GCM, I think the phrase "nonce" is much more clear about the dangers of re-using it. If you're familiar enough with crypto to understand the dangers of reusing an IV/nonce in GCM mode, then calling it a nonce shouldn't confuse you. And if you're not familiar enough with the subject matter, then you probably shouldn't be messing around with crypto algorithms at this low of a layer (IMHO).

sul3n3t:
Sorry for taking so long with this PR. I think this is a good change, thanks!

sul3n3t:
@joelthompson Would you mind reviewing and signing the CLA in [CONTRIBUTING.md](https://github.com/square/keywhiz/blob/master/CONTRIBUTING.md)? I will have to revert this PR otherwise.

joelthompson:
@sul3n3t -- set.

PR #48, opened 2015-04-17
adding setup notes for creating keystores

sherzberg:
this isn't ready to merge as a few examples still dont work.

* `keywhiz.cli login` followed by a `add secret` gives:

```bash
$ keywhiz.cli add secret --name testSecret < secret.txt 
Exception in thread "main" java.io.IOException: Please login by running a command without piping.
For example: keywhiz.cli login
        at keywhiz.cli.ClientUtils.readPassword(ClientUtils.java:191)
        at keywhiz.cli.CommandExecutor.executeCommand(CommandExecutor.java:118)
        at keywhiz.cli.CliMain.main(CliMain.java:68)
```

* `keywhiz.cli list` works just fine

Thoughts on these changes?

PR #56, opened 2015-04-23
Revamp dev/test certificates

sul3n3t:
- Longer-lived test certificates and CRL
- Server cert contains localhost and 127.0.0.1
- Removes old, unused openssl files
- Simplify test client setup
- Configure empty CRL so revocation check does not fail

R: @alokmenghrajani


alokmenghrajani:
You should probably comment on https://github.com/square/keywhiz/pull/48 with info on how you created these files.
Also, don't forget to update keywhiz-fs.

alokmenghrajani:
:+1:

PR #57, opened 2015-04-23
Cleanup dbi

alokmenghrajani:


server/src/main/java/keywhiz/ServiceModule.java
```diff
@@ -146,79 +128,7 @@ public ServiceModule(KeywhizConfig config, Environment environment) {
     return DSL.using(dataSource.getConnection());
   }
 
-  // DBI
-
-  @Provides ObjectMapper configuredObjectMapper(Environment environment) {
```
sul3n3t: Don't we still need the object mapper provides?

server/pom.xml
```diff
@@ -93,10 +93,6 @@
       <groupId>io.dropwizard.modules</groupId>
       <artifactId>dropwizard-java8-auth</artifactId>
     </dependency>
-    <dependency>
```
sul3n3t: There's a matching entry w/ version in the root pom.xml

sul3n3t:
LGTM aside from my two comments.

server/src/main/java/keywhiz/ServiceModule.java
alokmenghrajani: I moved it up a little.

sul3n3t:
:+1:

PR #58, opened 2015-04-23
Improves mappers

alokmenghrajani:


server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -55,17 +55,17 @@ public void deleteGroup(Group group) {
...
   public Optional<Group> getGroupById(long id) {
     GroupsRecord r = dslContext.fetchOne(GROUPS, GROUPS.ID.eq(Math.toIntExact(id)));
     return Optional.ofNullable(r).map(
-        (rec) -> rec.map(new GroupMapper()));
+        rec -> new GroupMapper().map(rec));
   }
 
   public ImmutableSet<Group> getGroups() {
-    List<Group> r = dslContext.select().from(GROUPS).fetch().map(new GroupMapper());
+    List<Group> r = dslContext.selectFrom(GROUPS).fetch().map(new GroupMapper());
```
sul3n3t: Could use `selectFrom(...)` in the other DAOs to be consistent.

sul3n3t: Nevermind, I see the problem

sul3n3t:
:+1:

server/src/main/java/keywhiz/service/daos/GroupDAO.java
alokmenghrajani: You can't use `.join()` when doing `.selectFrom()` but you can do `.fetchInto()` to get the same result.

PR #60, opened 2015-04-24
AES key generation command

sul3n3t:
Runnable process/documentation for deriving a derivation key. The
context around using this will be documented on the wiki.

R: @alokmenghrajani

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,7 +209,7 @@
       <dependency>
         <groupId>org.assertj</groupId>
         <artifactId>assertj-core</artifactId>
-        <version>1.7.1</version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -209,7 +209,7 @@
       <dependency>
         <groupId>org.assertj</groupId>
         <artifactId>assertj-core</artifactId>
-        <version>1.7.1</version>
+        <version>3.0.0</version>
```
sul3n3t: The new version of assertj has richer matchers including for java 8 types. In this case, I update to assert on Path instances.

alokmenghrajani:
lgtm

PR #61, opened 2015-04-24
Turns down logging to reasonable levels

sul3n3t:
Leaves a commented out logging configuration for when you want
verbosity.

R: @alokmenghrajani


alokmenghrajani:
:+1:

PR #71, opened 2015-05-01
Replaces exec-maven-plugin with sql-maven-plugin.

alokmenghrajani:
When we switch to mysql, we could write the shell command which is equivalent to createdb. I however think sql-maven-plugin is more database agnostic compared to createdb.


sul3n3t:
LGTM pending green build

PR #77, opened 2015-05-19
Wraps configuration in new DSLContexts (for transactions).

alokmenghrajani:
This is the right way to use jOOQ's transactions
(see http://www.jooq.org/doc/3.6/manual/sql-execution/transaction-management/).

Authoring this PR makes me wonder if all the DAO methods should take a context
as a parameter instead of injecting it. We should make it hard to accidentally
use the incorrect DSLContext.

### DIFF
--- a/server/src/main/java/keywhiz/auth/bcrypt/BcryptAuthenticator.java
+++ b/server/src/main/java/keywhiz/auth/bcrypt/BcryptAuthenticator.java
@@ -22,6 +22,7 @@ import io.dropwizard.java8.auth.Authenticator;
 import java.util.Optional;
 import keywhiz.auth.User;
 import keywhiz.service.daos.UserDAO;
+import org.jooq.DSLContext;
 import org.mindrot.jbcrypt.BCrypt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,9 +31,12 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 public class BcryptAuthenticator implements Authenticator<BasicCredentials, User> {
   private static final Logger logger = LoggerFactory.getLogger(BcryptAuthenticator.class);
+
+  private final DSLContext dslContext;
   private final UserDAO userDAO;
 
-  public BcryptAuthenticator(UserDAO userDAO) {
+  public BcryptAuthenticator(DSLContext dslContext, UserDAO userDAO) {
+    this.dslContext = dslContext;
     this.userDAO = checkNotNull(userDAO);
   }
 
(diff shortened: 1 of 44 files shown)


alokmenghrajani:
[ ] @dgalling 
[ ] @sul3n3t

server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -49,25 +50,20 @@
...
-    this.clientDAO = clientDAO;
-    this.groupDAO = groupDAO;
-    this.secretContentDAO = secretContentDAO;
-    this.secretSeriesDAO = secretSeriesDAO;
     this.mapper = mapper;
   }
 
   public void findAndAllowAccess(long secretId, long groupId) {
     dslContext.transaction(configuration -> {
+      DSLContext context = DSL.using(configuration);
```
sul3n3t: Should give a better name for the DSLContext variables so the wrong one isn't used.

alokmenghrajani: How about `innerJooqContext`?

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -24,17 +24,13 @@
 import keywhiz.jooq.tables.records.ClientsRecord;
 import org.jooq.DSLContext;
 
+import static jersey.repackaged.com.google.common.base.Preconditions.checkNotNull;
```
alokmenghrajani: todo: fix this

server/src/main/java/keywhiz/auth/bcrypt/BcryptAuthenticator.java
```diff
@@ -48,7 +52,7 @@ public BcryptAuthenticator(UserDAO userDAO) {
     String password = credentials.getPassword();
 
     // Get hashed password column from BCrypt table by username
-    Optional<String> optionalHashedPwForUser = userDAO.getHashedPassword(username);
+    Optional<String> optionalHashedPwForUser = userDAO.getHashedPassword(dslContext, username);
```
sul3n3t: This example makes me think changing the API like this may be a code smell. UserDAO requires DSLContext for its lookup (`getHashedPassword`) and its not different based on the lookup. The DSLContext is really a dependency of UserDAO not BcryptAuthenticator, so it really doesn't belong here.

Formerly with JDBI, there were still transaction scopes but they were bound to a Handle object. There was some magic using the (SQL Object API)[http://jdbi.org/sql_object_overview/], which made it transparent. What happened is that instantiated DAO objects had the handle passed in as a dependency. To handle the case of a transaction wrapping multiple calls across classes, CreateSqlObject would pass the current handle to the other DAO instead of a new one. Ultimately, code outside of DAOs wouldn't know about Handle objects, and similarly maybe they shouldn't know about DSLContext objects.

alokmenghrajani: I didn't even address `SecretController.java` which makes things worse. I'm very unclear how we should move forward.

alokmenghrajani:
See https://github.com/square/keywhiz/pull/102

PR #79, opened 2015-05-26
Changes default CLI url to use localhost

sul3n3t:
If you execute the CLI without specifying --url, it now assumes
https://localhost:4444/. This is what would be expected for development
and testing environments and should reduce friction.

In staging, production, or other, the --url argument should really be
specified. I have some concern people may create server TLS certificates
for 'localhost' (which would be terrible), but I think it's not likely
due merely to this change and the friction savings would be worthwhile.

R: @alokmenghrajani


alokmenghrajani:
Should we care about potential ipv6 issues when localhost != 127.0.0.1? lgtm otherwise.

sul3n3t:
Localhost resolving to an IPv6 address should be fine. We include both `localhost` and `127.0.0.1` in the dev/test server certificate, but really adding IP addresses isn't a great approach. Using localhost will just require localhost to be a SAN in the server cert, regardless of IPv4/IPv6.

PR #81, opened 2015-05-26
Example using file for cookie key

sul3n3t:
Changes the development configuration to use a file to hold the cookie
key. This is an example of how Keywhiz server should be configured for
staging or production (no secrets in configuration).

R: @alokmenghrajani 

Tested with:
```
$ mvn package
$ java -jar server/target/keywhiz-server-0.7.4-SNAPSHOT-shaded.jar migrate server/src/main/resources/keywhiz-development.yaml
$ java -jar server/target/keywhiz-server-0.7.4-SNAPSHOT-shaded.jar db-seed server/src/main/resources/keywhiz-development.yaml
$ java -jar server/target/keywhiz-server-0.7.4-SNAPSHOT-shaded.jar server server/src/main/resources/keywhiz-development.yaml
```

Then, I logged into the UI successfully.


server/src/main/resources/keywhiz-development.yaml
```diff
@@ -82,7 +82,7 @@ userAuth:
 # alternateUiPath: ui/app/
 
 # Base64 of "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
```
alokmenghrajani: This comment probably needs to be updated. Do you want to recommend a way to generate the the file? `head -c 32 /dev/urandom | base64`?

alokmenghrajani:
:+1:

server/src/main/resources/keywhiz-development.yaml
sul3n3t: Good idea, added

PR #85, opened 2015-05-27
CLI cookie version 1

sul3n3t:
Finally figured out why the CLI is having issues in development with default settings.

R: @alokmenghrajani or @dgalling 

    Always sets HttpCookie's version to 1

    If HttpCookie determines the version is zero, the CLI may stop working.
    The serialization from HttpCookie is different and drops important
    fields, sometimes resulting in the cookie not even being sent by the
    client.


alokmenghrajani:
:+1:

PR #90, opened 2015-05-29
Minor changes to support MySql 5.5

alokmenghrajani:
Travis seems to be using MySQL 5.5 by default and I didn't
figure out a way to get tests to run with MySQL 5.6. This commit
therefore prepares us to run on the older version of mysql.


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -250,10 +251,14 @@ public void findAndEvictClient(long clientId, long groupId) {
   }
 
   protected void allowAccess(long secretId, long groupId) {
+    OffsetDateTime now = OffsetDateTime.now();
```
sul3n3t: Should specify UTC or it will use the system default.

alokmenghrajani: Totally. Fixed.

sul3n3t:
LGTM

PR #98, opened 2015-06-03
Enable and fixup findbugs

sul3n3t:
R: @alokmenghrajani 

- Enables findbugs in CI
- Changes LoginRequest to use AutoValue so char[] is not exposed
  - Subtle change makes the response type 400 when field missing, which
    is arguably more correct.
- Make some fields final
- Suppress TinyIntConverter warning where null should be returned
- Suppress checking jOOQ-generated files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,5 @@ script:
   - mvn test -P postgres
   - mvn test -P mysql
   - mvn test -P h2
+  - "mvn compile findbugs:check"
   - cd $UI_DIR && grunt test
--- a/model/src/main/java/keywhiz/model/TinyIntConverter.java
+++ b/model/src/main/java/keywhiz/model/TinyIntConverter.java
@@ -16,9 +16,11 @@
 
 package keywhiz.model;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.jooq.Converter;
 
 public class TinyIntConverter implements Converter<Byte, Boolean> {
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -21,4 +21,5 @@ script:
   - mvn test -P postgres
   - mvn test -P mysql
   - mvn test -P h2
+  - "mvn compile findbugs:check"
```
alokmenghrajani: why are you using quotes

model/src/main/java/keywhiz/model/TinyIntConverter.java
```diff
@@ -16,9 +16,11 @@
 
 package keywhiz.model;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.jooq.Converter;
 
 public class TinyIntConverter implements Converter<Byte, Boolean> {
+  @SuppressFBWarnings(value = "NP_BOOLEAN_RETURN_NULL", justification = "null is a valid DB value")
   @Override public Boolean from(Byte aByte) {
```
alokmenghrajani: We could write this as `Converter<Byte, Optional<Boolean>>` and force callers to handle the `Optional.empty` case?

alokmenghrajani:
:+1:

.travis.yml
sul3n3t: `:` looks like a yaml key/value

model/src/main/java/keywhiz/model/TinyIntConverter.java
sul3n3t: That didn't seem to work very well on CI.

PR #102, opened 2015-06-04
Nested transactions

sul3n3t:
R: @alokmenghrajani 

* Use DAOFactory's to create DAOs
 
 * DAOFactory reduces passing around DSLContexts everywhere by allowing a
    factory to be injected and the DAO with correct DB backing to be
    instantiated. Particularly during a nested, transaction, the
    instantiated DAO will use the same underlying connection.
 * This makes a bunch of changes to dependencies (DAO factories instead of
    DAOs) and makes it more clear which constructors are `@Inject`'ed and which
    ones are manually called. Lots of DAO entries from ServiceModule are no
    longer needed.
 * Some tests that needed updating but were complex, like AclDAOTest, were
    simplified.

* Limit test database connections

 * Allowing fewer database connections in testing will hopefully bring
    forward more bugs where a queries is happening outside, but during
    another transaction. The read-write pool needs 2 connections for flyway
    migrations.

* Update jOOQ and guice

 * Make guice-testlib available to server.

* Consistent use of localhost in all dev configs

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -16,7 +16,6 @@
 
 package keywhiz.service.daos;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableSet;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
@@ -33,7 +32,14 @@ import keywhiz.api.model.SecretContent;
 import keywhiz.api.model.SecretSeries;
 import keywhiz.api.model.SecretSeriesAndContent;
 import keywhiz.jooq.tables.records.SecretsRecord;
+import keywhiz.service.config.Readonly;
+import keywhiz.service.daos.ClientDAO.ClientDAOFactory;
+import keywhiz.service.daos.GroupDAO.GroupDAOFactory;
+import keywhiz.service.daos.SecretContentDAO.SecretContentDAOFactory;
+import keywhiz.service.daos.SecretSeriesDAO.SecretSeriesDAOFactory;
+import org.jooq.Configuration;
 import org.jooq.DSLContext;
+import org.jooq.impl.DSL;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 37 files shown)


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -153,14 +176,29 @@ public void findAndEvictClient(long clientId, long groupId) {
...
     ImmutableSet.Builder<SanitizedSecret> set = ImmutableSet.builder();
 
-    for (SecretSeries series : getSecretSeriesFor(group)) {
-      for (SecretContent content : secretContentDAO.getSecretContentsBySecretId(series.getId())) {
-        SecretSeriesAndContent seriesAndContent = SecretSeriesAndContent.of(series, content);
-        set.add(SanitizedSecret.fromSecretSeriesAndContent(seriesAndContent));
+    return dslContext.transactionResult(configuration -> {
+      SecretContentDAO secretContentDAO = secretContentDAOFactory.using(configuration);
+
+      List<SecretSeries> serieses = DSL.using(configuration)
```
alokmenghrajani: why don't you call `getSecretSeriesFor(configuration, group)` here?

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -217,4 +246,33 @@ public void deleteSecretByNameAndVersion(String name, String version) {
...
+
+    @Override public SecretDAO readwrite() {
+      return new SecretDAO(jooq, secretContentDAOFactory, secretSeriesDAOFactory);
+    }
+
+    @Override public SecretDAO readonly() {
+      return new SecretDAO(readonlyJooq, secretContentDAOFactory, secretSeriesDAOFactory);
+    }
+
+    @Override public SecretDAO using(Configuration configuration) {
```
alokmenghrajani: It looks like the code is sometimes doing `dslContext` → `dslContext.configuration()` → `DSL.using(configuration)`. What is the side effect of doing that? What if we had `using(Configuration config)` and `using(DSLContext context)`?

server/src/main/java/keywhiz/service/providers/AutomationClientAuthFactory.java
```diff
@@ -39,7 +41,11 @@
 public class AutomationClientAuthFactory {
   private final Authenticator<String, AutomationClient> authenticator;
 
-  @Inject public AutomationClientAuthFactory(ClientDAO clientDAO) {
+  @Inject public AutomationClientAuthFactory(ClientDAOFactory clientDAOFactory) {
+    this.authenticator = new MyAuthenticator(clientDAOFactory.readwrite());
```
alokmenghrajani: should this be `clientDAOFactory.read()`?

server/src/main/java/keywhiz/service/daos/SecretDAO.java
sul3n3t: Configuration contains settings and a DataSource (internally a Connection). Performing DSL.using(configuration) doesn't change any state, but just creates a wrapper object around the Connection. Similarly, `DSLContext.configuration()` just exposed the wrapped configuration and does not create one.

I think we should have only one of the classes (`using(Configuration)` or `using(DSLContext)`) but not both. My PR uses Configuration because that is the direct object you get from a jOOQ transaction and that's what `using` is for.

Unfortunately, the cases where we execute `dslContext.configuration()` are where we want a transaction to be used except for the MySQL/read-only/jOOQ bug. The other case is the non-public methods are being unit-tested.

server/src/main/java/keywhiz/service/daos/AclDAO.java
sul3n3t: No idea how that ended up that way. Fixed to use that method.

server/src/main/java/keywhiz/service/providers/AutomationClientAuthFactory.java
sul3n3t: Sounds good, switched to readonly

server/src/main/java/keywhiz/service/daos/SecretDAO.java
alokmenghrajani: I'm not sure if the fact that `DSL.using(configuration)` and `DSLContext.configuration()` don't change state is something we can rely on in the long term. What would be the downsides of having only `using(DSLContext)`?

alokmenghrajani:
Besides my last comment (which can be addressed now or at a later point in time) this looks great! Your solution looks way more elegant than my initial attempt, thanks for cleaning this up. Feel free to merge.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
sul3n3t: When you're calling a DAO from another DAO in a transaction, the caller would have to use `SomethingDAOFactory.using(DSL.using(configuration))` since you get a Configuration from jOOQ in a transaction.

sul3n3t:
Thanks, open to more discussion on DSLContext vs Configuration, but it'd be good to merge this so far.

PR #103, opened 2015-06-08
Fix use of Optional for null

sul3n3t:
R: @alokmenghrajani


alokmenghrajani:
lgtm.

The cli has ~50% of test coverage, should we create an issue to eventually add more tests?

PR #109, opened 2015-06-09
Fixes bug in AclDAO.getSecretSeriesFor

alokmenghrajani:
If a client is allowed to access a secret via multiple groups, a bug in the code would throw an Exception. The issue was introduced when converting JDBI's @SingleValueResult into jOOQ's fetchOneInto.

I looked at all the other instances of fetchOneInto and it seems only one method needs this fix.


sul3n3t:
:+1:

PR #116, opened 2015-06-10
Block bad requests sooner with preconditions

sul3n3t:
R: @alokmenghrajani 

* Adds precondition checks to KeywhizClient, to block bad requests sooner.
* Move base64'ing content into KeywhizClient instead of CLI.


alokmenghrajani:
:+1:

PR #126, opened 2015-06-17
Fix docker container usage example

calavera:
Use `sh -c` otherwise `*` is expanded in the host and fails with something like:

```
no matches found: server/target/keywhiz-server-*-SNAPSHOT-shaded.jar
```

/cc @jfrazelle, because she said that you could trust her :stuck_out_tongue_winking_eye:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 #   docker build --rm --force-rm -t square/keywhiz .
 #
 # Example usage:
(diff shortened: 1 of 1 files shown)


alokmenghrajani:
lgtm.

Would you mind reviewing and signing the CLA in CONTRIBUTING.md?

calavera:
@alokmenghrajani CLA signed.

PR #127, opened 2015-06-18
Lists fields in DSLContext.select()

alokmenghrajani:
This improves performance. I initially expected fetchInto() to take care of
this but looking at sql logs reveals that we are overfetching data.


sul3n3t:
:+1:

PR #129, opened 2015-07-02
Quote db entities as is

richo:
I was unable to bring up a keywhiz on postgres, because the db-seed step tries to quote all db entities in uppercase, whereas the migrations bring them up lowercase.

It appears that this was an issue with h2 at some point as well. I don't have a mysql to test on, but I don't see any reason that this would break anything so I think we just want this behaviour for everything.

### DIFF
--- a/server/src/main/java/keywhiz/utility/DSLContexts.java
+++ b/server/src/main/java/keywhiz/utility/DSLContexts.java
@@ -40,12 +40,8 @@ public class DSLContexts {
     try (Connection conn = dataSource.getConnection()) {
       dialect = dialect(conn);
     }
-    if (dialect == SQLDialect.H2) {
-      return DSL.using(dataSource, SQLDialect.H2,
-            new Settings()
-                .withRenderSchema(false)
-                .withRenderNameStyle(RenderNameStyle.AS_IS));
+    return DSL.using(dataSource, dialect, new Settings()
+                                          .withRenderSchema(false)
+                                          .withRenderNameStyle(RenderNameStyle.AS_IS));
     }
-    return DSL.using(dataSource, dialect);
-  }
 }
--- a/server/src/main/java/keywhiz/utility/DSLContexts.java
+++ b/server/src/main/java/keywhiz/utility/DSLContexts.java
@@ -40,8 +40,9 @@ public class DSLContexts {
     try (Connection conn = dataSource.getConnection()) {
       dialect = dialect(conn);
     }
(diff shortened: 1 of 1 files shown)


server/src/main/java/keywhiz/utility/DSLContexts.java
```diff
@@ -40,12 +40,9 @@ public static DSLContext databaseAgnostic(DataSource dataSource) throws SQLExcep
     try (Connection conn = dataSource.getConnection()) {
```
alokmenghrajani: the try block goes away

server/src/main/java/keywhiz/utility/DSLContexts.java
```diff
@@ -40,12 +40,9 @@ public static DSLContext databaseAgnostic(DataSource dataSource) throws SQLExcep
...
       dialect = dialect(conn);
     }
-    if (dialect == SQLDialect.H2) {
-      return DSL.using(dataSource, SQLDialect.H2,
+    return DSL.using(dataSource, SQLDialect.H2,
```
alokmenghrajani: you don't want to use SQLDialect.H2 for non H2 connections. We want DSL.using(dataSource, new Settings()...) instead.

alokmenghrajani:
Thanks for doing this, I agree it's better this way. The PR still needs some minor work, see my inline comments.

server/src/main/java/keywhiz/utility/DSLContexts.java
richo: Oof, I goofed during the refactor. This was originally meant to be consume the `dialect` collected above- is `using` clever enough to just work it out for me?

alokmenghrajani: afaik, if you don't give `using` a dialect it calls `dialect(conn)` which "guesses" the dialect from the connection string.

richo:
Thanks for the feedback @alokmenghrajani.

I've updated the PR- there's isn't a variant of `using` that doesn't take a dialect, so I kept the original code for sniffing it. The tests are green now :)

alokmenghrajani:
Right, http://www.jooq.org/javadoc/3.6.x/org/jooq/impl/DSL.html#using-java.sql.Connection-org.jooq.conf.Settings- takes a Connection instead of a DataSource.

We might want to rename databaseAgnostic to something else, since the code is no longer special casing H2.

PR #130, opened 2015-07-10
Replace postgres JDBC driver with pgjdbc-ng

sul3n3t:
@alokmenghrajani 

pgjdbc-ng is a more up-to-date driver than the official postgres driver.
In particular, prepared statements can be cached.


sul3n3t:
@alokmenghrajani WDYT?

alokmenghrajani:
:+1:

PR #134, opened 2015-09-15
Doclet update

### LINKED ISSUE #54: Swagger docs don't understand ZonedDateTime fields
[Swagger](http://swagger.io/) auto-generates the API documentation for Keywhiz. When it generates model JSON, it doesn't seem to understand datetime fields (using ZonedDateTime) and expands the object further and further. What we would like it to do is stop at any `ZonedDateTime` and treat it as a ISO8691 datetime string.

sul3n3t:
Should fix #54 

R: @mcpherrinm, @alokmenghrajani


alokmenghrajani:
:+1:

Issue #89, opened 2015-05-29
Automation API style inconsistent for deleting a secret


sul3n3t:
To delete a secret via automation API, `DELETE /automation/secrets/<xyz>` takes a secret name as `<xyz>`. All of the other automation APIs use an id for deletion.

nealharris:
Fixed in https://github.com/square/keywhiz/pull/136

PR #136, opened 2015-09-29
Automation API V2

sul3n3t:
@alokmenghrajani @mcpherrinm @csstaub 

Implements V2 API for automation:
* Endpoints are currently still inefficient but API is spec'd out
* Some tests are @Ignored, pending additional feature implementation
* NotImplementedException is used twice where DAO changes are necessary

I intend to follow this PR with another that 1) adds missing DAO methods and 2) makes DAO operations more efficient, specifically for `/automation/v2/`.

Definitely please critique the API organization, requests, methods, etc. too.


sul3n3t:
Looks like it failed the travis CI build on findbugs. That's because there's some unused variables for partially implemented methods. Will resolve.

api/src/main/java/keywhiz/api/automation/v2/CreateClientRequestV2.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @JsonProperty("name") public abstract String name();
+  @JsonProperty("groups") public abstract ImmutableSet<String> groups();
+  @JsonIgnore public abstract Optional<String> description();
+
+  /* Raw getter for proper API docs generation. */
+  @SuppressWarnings("unused")
+  @JsonProperty("description") @Nullable public String rawDescription() {
+    return description().orElse(null);
```
alokmenghrajani: why not return an empty string in the absent case (`orElse("")`) here and then not have the `@Nullable` annotation above?

api/src/main/java/keywhiz/api/automation/v2/ClientDetailResponseV2.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  /**
+   * Static factory method used by Jackson for deserialization
+   */
+  @SuppressWarnings("unused")
+  @JsonCreator public static ClientDetailResponseV2 fromParts(
+      @JsonProperty("name") String name,
+      @JsonProperty("description") String description,
+      @JsonProperty("creationDate") OffsetDateTime creationDate,
```
alokmenghrajani: we talked about it offline but I'll mention it here. I feel using int + milliseconds since epoch helps interoperability.

server/src/main/java/keywhiz/service/resources/automation/v2/ClientResource.java
```diff
@@ -0,0 +1,269 @@
...
+   *
+   * @responseMessage 200 Listing succeeded
+   * @responseMessage 404 Client not found
+   */
+  @GET
+  @Path("{name}/groups")
+  @Produces(APPLICATION_JSON)
+  public Iterable<String> clientGroupsListing(@Auth AutomationClient automationClient,
+      @PathParam("name") String name) {
+    Client client = clientDAO.getClient(name)
```
alokmenghrajani: do you think it's worth having `clientDAO.getClientEx` which would be the same as `clientDAO.getClient` but throws an exception when missing?

alokmenghrajani:
lgtm besides my minor comments. We need to document why we are creating this new API and how to migrate from the old api to this new one.

server/src/main/java/keywhiz/service/resources/automation/v2/ClientResource.java
sul3n3t: I wouldn't want to throw a web exception in the DAO, so it would mean catching and re-throwing a different exception. For a method that can return something or nothing, Optional<Something> seems like a fit.

api/src/main/java/keywhiz/api/automation/v2/CreateClientRequestV2.java
sul3n3t: Good point. I think I'll take a pass at some of the string fields like this and just use empty string instead of Nullable/Optional.

alokmenghrajani: It's also going to be more consistent with some of the new builder code (e.g. https://github.com/square/keywhiz/pull/136/files#diff-7d09372918e0bec1a40c9e65ce08c8cdR21)

api/src/main/java/keywhiz/api/automation/v2/ClientDetailResponseV2.java
```diff
@@ -0,0 +1,42 @@
...
+      @JsonProperty("updateDate") OffsetDateTime updateDate,
+      @JsonProperty("createdBy") String createdBy,
+      @JsonProperty("updatedBy") String updatedBy) {
+    return new AutoValue_ClientDetailResponseV2(name, description, creationDate, updateDate, createdBy, updatedBy);
+  }
+
+  @JsonProperty("name") public abstract String name();
+  @JsonProperty("description") public abstract String description();
+  @JsonProperty("creationDate") public abstract OffsetDateTime creationDate();
```
alokmenghrajani: We could change this to `createdAt`.

sul3n3t:
This should also address #89

api/src/main/java/keywhiz/api/automation/v2/ClientDetailResponseV2.java
sul3n3t: I'm still planning on adding the integer timestamp, but I think I'll put it in the next PR to avoid blocking things.

PR #137, opened 2015-10-06
Uses empty string for missing descriptions

sul3n3t:
There has been no value in representing descriptions as fields that may
or may not be present (i.e. Optional), so this PR removed any Optional
treatment of descriptions. Similarly, null descriptions are not a useful
value either. As just a display/infomative metadata, an empty string
provides more benefit than null. This PR moves null descriptions to
empty strings.

R: @alokmenghrajani


sul3n3t:
I made this PR to remove `Optional` descriptions, separate from #136. I plan to rebase that PR after this is merged.

alokmenghrajani:
:+1:

PR #139, opened 2015-10-09
Wrap OffsetDateTime in ApiDate & update Jackson.

mcpherrinm:
Changes:

Update Jackson to the newest stable release (2.6.1).

We wrap OffsetDateTime in an ApiDate that provides a thin wrapper around it
and custom Json de/serializers.  It also has static now() and parse() methods
that are heavily used in tests, and equality & hashcode checks implented in the
obvious way.

There's a lot of mundane changes to pipe the type throughout the API, and to
update tests appropriately.

Remove the now unused options to set up the date format in Jackson: This makes
it easier to find buggy serialization, as they get written as a number instead.

- Background & Alternatives:

Keywhiz uses Jackson for JSON de/serialization.  When OffsetDateTime is
serialized with Jackson's JSR310 module, which omits the milliseconds if
they're all zero.  While this is unlikely to happen, as most times in the
keywhiz world are timestamps and the milliseconds are effectively random, it
does make Keywhiz-fs unhappy, and unable to list secrets, or read the secret
with the unsupported date format.

We set the jackson date format explicitly, but that's a red herring: The JSR310
date module doesn't use this.  We could explicitly set the pattern with
@JsonPattern to be yyyy-MM-dd'T'mm-HH-ss.SSS'Z' which serializes dates properly,
but then won't deserialize because OffsetDateTime's constructor doesn't know
what timezone it's in.  If we used the unquoted 'Z', the timezone will serialize
as +0000, which is undesired too, but it deserializes properly.

We could load a custom module into Jackson that provides custom de/serializers
for OffsetDateTime, but we need to load the Jdk8 module, which provides
conflicting de/serializers for OffsetDateTime.  You could probably get that to
work with the right ordering, but that feels awfully fragile to me (and I
didn't get it to work).

R? @sul3n3t @alokmenghrajani


sul3n3t:
I'd like to get #136 merged first, if you don't mind. A little rebase will be required.

mcpherrinm:
Yeah, that's fine by me.

mcpherrinm:
Rebased onto master after #136 landed and made tests pass again.  Ran tests in IntelliJ and travis is green.

api/src/main/java/keywhiz/api/ApiDate.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+import java.io.IOException;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * This is a wrapper for the date class used in API responses so we can have a custom JSON ser/deser implementation
```
alokmenghrajani: serializer/deserializer is easier to grep for than "ser/deser".

alokmenghrajani:
:+1:

Do you have merge rights on this repo?

api/src/main/java/keywhiz/api/ApiDate.java
mcpherrinm: Yeah, that's reasonable.  I'll change it.

mcpherrinm:
Yeah, I have merge rights.

PR #140, opened 2015-10-19
Use seconds instead of OffsetDateTime in api v2.

alokmenghrajani:
Seconds since epoch are much nicer to deal with in an API. It avoids the
need to correctly format and parse date/time strings across different
programming languages. It also takes less space on the wire.

note: JSR-310 makes it easier to use seconds instead of milliseconds.


mcpherrinm:
LGTM

PR #149, opened 2015-11-04
Register Jackson JavaTimeModule in cli

mcpherrinm:
In commit 9c33fdf344ce8baf0df1b282037235de642c88b7 the deprecated JSR310Module
was removed, instead of being replaced with JavaTimeModule.  This results in
being unable to deserialize ApiDate objects (because it calls the OffsetDateTime
deserializer), and so some keywhiz.cli commands don't work without this fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: java
+sudo: required
+dist: trusty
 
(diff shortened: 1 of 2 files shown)


alokmenghrajani:
:+1:

PR #153, opened 2015-11-06
Remove foreign keys, fixes for MySQL, prepare v0.7.9

mcpherrinm:
Drop our use of foreign keys from the databases.
Postgres is a SQL migration, straightforward.

The H2 case is a bit more complicated, because you delete constraints by name and we don't know what the generated names are.  So I use a java migration to look them up then delete them.

Mysql never had working foreign keys in the first place, so get rid of them.

Adds Java code to delete where the cascading deletes would have.


server/src/main/resources/db/mysql/migration/V1__create_tables.sql
```diff
@@ -40,8 +40,8 @@ CREATE TABLE clients (
 
 CREATE TABLE accessgrants (
   id int NOT NULL AUTO_INCREMENT,
-  groupid int REFERENCES groups(id) ON DELETE CASCADE,
```
alokmenghrajani: Flyway uses checksums on the file contents and will get confused. We should avoid changing existing `.sql` files unless it's something completely broken. In this case, I would just add a new migration to remove the foreign key and call it a day.

alokmenghrajani:
I just updated this branch. I think this is everything we need for 0.7.9, right?

mcpherrinm:
I agree we're ready to roll for 0.7.9

PR #161, opened 2015-11-19
Add ability to fetch only the list of secret names

alokmenghrajani:
With >10k secrets, this significantly improves the time it takes the CLI to list all the secrets. The
results however do not contain individual entries for versions.


mcpherrinm:
:+1:

PR #162, opened 2015-11-20
Use SimpleEntry instead of javafx Pair

mcpherrinm:
JavaFX isn't shipped with the OpenJDK on my Fedora install.  It seems
ridiculous to require it just for a Pair class, when SimpleEntry will do
the same thing.


alokmenghrajani:
:+1:

PR #168, opened 2016-01-27
Rename variables which hide the field declared

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:HiddenFieldCheck - “ Local variables should not shadow class fields”. You can find more information about the issues here: https://dev.eclipse.org/sonar/rules/show/squid:HiddenFieldCheck
Please let me know if you have any questions.
Ayman Abdelghany.

### DIFF
--- a/api/src/test/java/keywhiz/api/AutomationSecretResponseTest.java
+++ b/api/src/test/java/keywhiz/api/AutomationSecretResponseTest.java
@@ -61,12 +61,12 @@ public class AutomationSecretResponseTest {
 
   @Test
   public void serializesCorrectly() throws Exception {
-    String secret = "YXNkZGFz";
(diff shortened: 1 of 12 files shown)


mcpherrinm:
While it's not the best style to shadow class fields, this PR mechanically renames them, and I don't think that adds any clarity.

I will happily accept a PR to rename variables to ones that add clarity, but please do each variable in a separate commit, and include reasoning on why the new name is chosen.

alokmenghrajani:
I don't think this kind of variable renaming in tests is worth doing, unless you are able to provide more meaningful names.

For DSLContext, shadowing is a desirable property! We want to make it harder to use the wrong context. It's easier to notice that someone wrote `this.dslContext` vs noticing that there is a `someOtherDslContext` in scope.

PR #171, opened 2016-02-02
Add a way to add local admin users to keywhiz

mcpherrinm:
If you don't use ldap, there's no exposed way to add users, except the "db-seed" command that puts a default user in.

This adds a "add-user" commadn to the server jar.

You invoke it like this, from the dev tree:

java -jar ./server/target/keywhiz-server-0.7.11-SNAPSHOT-shaded.jar add-user server/src/main/resources/keywhiz-development.yaml
New username:myuser
password for 'myuser': <type a password here>


sqshh:
LGTM.

alokmenghrajani:
:+1:

PR #175, opened 2016-02-13
Update Dockerfile

csstaub:
r: @jfrazelle @calavera @mcpherrinm @alokmenghrajani 

Updates the Dockerfile for building a Keywhiz server container.

Changes:
* Split mvn verify/install steps to cache dependency downloads
* Added entrypoint for running shaded server jar by default

Example usage for development, with persistent H2 DB in local volume:
```bash
docker volume create keywhiz-db-devel
docker run -v keywhiz-db-devel:/tmp/h2_data square/keywhiz migrate
docker run -it -v keywhiz-db-devel:/tmp/h2_data square/keywhiz server
```

### DIFF
--- a/docker.sh
+++ b/docker.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 RESET=$(tput -T xterm sgr0)
 RED=$(tput -T xterm setaf 1) 
@@ -11,6 +11,10 @@ if [ -z "$KEYWHIZ_CONFIG" ]; then
     echo "--- Use the KEYWHIZ_CONFIG environment variable to set a config file ---" 
     echo "------------------------------------------------------------------------"
     echo -n "${RESET}"
+fi
+
+if [ "$MIGRATE_ON_STARTUP" == "true" ]; then
+    java -jar server/target/keywhiz-server-*-SNAPSHOT-shaded.jar migrate $KEYWHIZ_CONFIG
 fi
 
 java -jar server/target/keywhiz-server-*-SNAPSHOT-shaded.jar $1 $KEYWHIZ_CONFIG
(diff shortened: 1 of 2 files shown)


docker.sh
```diff
@@ -0,0 +1,16 @@
...
+    KEYWHIZ_CONFIG=server/src/main/resources/keywhiz-development.yaml
+    echo -n "${RED}"
+    echo "------------------------------------------------------------------------"
+    echo "---    No configuration file specified, using development config!    ---"
+    echo "--- Use the KEYWHIZ_CONFIG environment variable to set a config file ---" 
+    echo "------------------------------------------------------------------------"
+    echo -n "${RESET}"
+fi
+
+java -jar server/target/keywhiz-server-*-SNAPSHOT-shaded.jar $1 $KEYWHIZ_CONFIG
```
mcpherrinm: I don't like our use of -*- here -- it breaks if you've built two versions.
I'm not sure what the best way to fix this is, though.

mcpherrinm:
Looks good to me.  We should investigate adding a docker test to our travis-ci config to ensure this keeps working, too.

docker.sh
csstaub: I mean, there will only ever be one version in the image/container. For a new version we'd build a new image.

mcpherrinm: I guess so - but I'm not very familiar with docker.  You can't end up in a state where you have old build artifacts hanging around, accidentally included?

csstaub: I don't think so -- the dockerfile describes how the image is built, starting from a clean slate.

mcpherrinm: Alright, sounds good then.  :ship:

calavera:
nice, looks good to me.

PR #177, opened 2016-02-18
Add docker profile and config

csstaub:
r: @mcpherrinm 
Adds a build profile for docker, and makes some improvements to the dockerfile and the development setup instructions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,16 +15,23 @@
 #
 # Development:
 #   Create a persistent data volume:
-#     docker volume create keywhiz-db-devel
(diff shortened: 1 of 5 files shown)


mcpherrinm:
LGTM, though I didn't test.  Do we need the seperate keywhiz-test.yaml.docker, or can we use the h2 one?  I didn't diff but they look the same.

csstaub:
It's the same, but for consistency I think it's nice to just have one per profile.

mcpherrinm:
okay, :+1:

PR #178, opened 2016-02-18
Add setup wizard for docker containers

csstaub:
r: @mcpherrinm @alokmenghrajani 

Add setup wizard for docker containers.

Run the wizard like this:

    docker run -it \
        -v keywhiz-data:/data \
        -v keywhiz-secrets:/secrets \
        square/keywhiz wizard

The wizard will create symmetric encryption keys and ask for TLS certificates/keys.

After it completes, run keywhiz server with the generated config like this:
    
    docker run \
        -p 4444:4444
        -v keywhiz-data:/data \
        -v keywhiz-secrets:/secrets \
        -e 'KEYWHIZ_CONFIG=/data/keywhiz-docker.yaml'
        square/keywhiz server

### DIFF
--- a/docker/wizard.sh
+++ b/docker/wizard.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+set -e
+
+function confirm() {
+    echo
+    read -p "$1 (y/n) "
+    if [[ $REPLY =~ ^[Yy]$ ]]; then
+        return 0
+    fi
+    if [[ $REPLY =~ ^[Nn]$ ]]; then
+        return 1
+    fi
+    echo "Answer must be one of y/n."
+    confirm "$1"
+}
+
+function assert_file_present() {
+    if ! [ -f $1 ]; then
+        echo "Missing file: $1. Aborting."
+        exit 1
+    fi 
+}
+
+cat - <<EOF
+*** Keywhiz Setup Wizard ***
+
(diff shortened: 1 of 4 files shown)


docker/wizard.sh
```diff
@@ -0,0 +1,123 @@
...
+export CONTENT_KEYSTORE_PATH="/secrets/content-encryption-key.jceks"
+rm -f $CONTENT_KEYSTORE_PATH
+keytool \
+    -genseckey -alias basekey -keyalg AES -keysize 128 -storepass "$CONTENT_KEYSTORE_PASSWORD" \
+    -keypass "$CONTENT_KEYSTORE_PASSWORD" -storetype jceks -keystore $CONTENT_KEYSTORE_PATH
+echo "done"
+
+echo
+echo "Keywhiz requires a SSL/TLS server certificate and private key to run."
+echo "You can import your own cert, or have the wizard generate one."
```
mcpherrinm: Can the wizard generate one? I don't see this

csstaub: Oh, whoops, no; I dropped that functionality. It seemed weird to generate it in the container and then dump the private key (which would be necessary to to make client certs). I'd rather just have people run that on their own.

mcpherrinm: :+1:

mcpherrinm:
:shipit:

PR #186, opened 2016-03-17
Add comments to Dropwizard configuration, and validateCerts: false

mcpherrinm:
Though validateCerts: false isn't used in the dev setup, there's not really any
production config examples elsewhere, and so it's helpful to include it.

cc @madtrax 

r? @csstaub @alokmenghrajani @sqshh


csstaub:
LGTM. Restarted the broken build for ya, looks like it timed out.

PR #189, opened 2016-04-15
Update travis build

csstaub:
@mcpherrinm 
Changes:
- compute coverage reports
- upload coverage to coveralls.io
- run tests once, instead of thrice

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: java
 
-# Experimental group for MySQL 5.6
-group: dev
-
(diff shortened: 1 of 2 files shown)


mcpherrinm:
We should use a Travis build matrix to run tests for all three dbs/

If we only want one db, I'd prefer we use MySQL, unless it's significantly
slower than h2.

mcpherrinm:
lgtm otherwise though, we can do matrix stuff later.

PR #191, opened 2016-04-19
Adding metrics

mcpherrinm:
(Updated.)

Adds a @Timed annotation to every endpoint in Keywhiz, so we get per-endpoint metrics.

It's not as fine-grained as I'd like, but it'll do.  Other Ad-hoc analysis (eg, per-client or by http status) will have to continue to use log files, but this at least gives us basic server health without munging logs.

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/SecretDeliveryResource.java
+++ b/server/src/main/java/keywhiz/service/resources/SecretDeliveryResource.java
@@ -90,7 +90,7 @@ public class SecretDeliveryResource {
    * @responseMessage 404 Secret with given name not found
    * @responseMessage 500 Secret response could not be generated for given Secret
    */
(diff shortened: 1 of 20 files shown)


mcpherrinm:
Change done via

perl -pi -e 's/^( +)(\@[GETPOSUDL]+)/$1\@Timed\n$1$2/' ./Automation/v2/*Resource.java ./Admin/*Resource.java ./*Resource.java

mcpherrinm:
To make sure I didn't miss any:  I ensured the log message "INFO  [2016-04-20 06:50:32,498] io.dropwizard.jersey.DropwizardResourceConfig: The following paths were found for the configured resources:" during startup had the same number of entries as there are @Timed annotations in the codebase - 64.

csstaub:
LGTM

PR #196, opened 2016-05-05
Make /_status always return JSON

mcpherrinm:
JSON looks like this:

{"db-read-write-health":{"healthy":true},"db-readonly-health":{"healthy":false,"message":"Unhealthy connection to database."},"deadlocks":{"healthy":true}}

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/StatusResource.java
+++ b/server/src/main/java/keywhiz/service/resources/StatusResource.java
@@ -1,9 +1,10 @@
 package keywhiz.service.resources;
 
-import com.codahale.metrics.annotation.Timed;
 import com.codahale.metrics.annotation.ExceptionMetered;
+import com.codahale.metrics.annotation.Timed;
 import com.codahale.metrics.health.HealthCheck;
 import com.codahale.metrics.health.HealthCheckRegistry;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import io.dropwizard.setup.Environment;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/resources/StatusResource.java
```diff
@@ -1,49 +1,42 @@
...
-    List<String> failing = results.entrySet().stream()
-        .filter(r -> !r.getValue().isHealthy())
-        .map(Map.Entry::getKey)
-        .collect(toList());
-
-    if (!failing.isEmpty()) {
-      throw new InternalServerErrorException(
-          "failing health checks: " + Arrays.toString(failing.toArray()));
+    if (results.entrySet().stream().anyMatch(r -> !r.getValue().isHealthy())) {
+      logger.warn("Health checks failed: {}", results);
```
csstaub: The failing health checks should warn by themselves, we don't want to depend on someone hitting /_status to get warn messages.

mcpherrinm: Ok. I mostly put this here because the InternalServerErrorException used to lead to a message being logged, so I wanted to ensure we still had that.

mcpherrinm:
I changed up the PR a bit to keep the message in the same format we currently log, and also providing structured healthchecks, along with a top level "status" so it's easier to see if any checks are failing.

JSON now looks like:

{"status":"ok","message":"ok","results":{"db-read-write-health":{"healthy":true},"db-readonly-health":{"healthy":true},"deadlocks":{"healthy":true}}}

{"status":"critical","message":"failing health checks: [db-readonly-health]","results":{"db-read-write-health":{"healthy":true},"db-readonly-health":{"healthy":false,"message":"Unhealthy connection to database."},"deadlocks":{"healthy":true}}}

alokmenghrajani:
👍

PR #197, opened 2016-05-05
Add ability to search/replace intermediate certificates.

alokmenghrajani:
- [ ] @csstaub 
- [ ] @mcpherrinm


mcpherrinm:
Why is this endpoint unauthenticated?

alokmenghrajani:
Easier to test (I would need to modify `keywhiz.client.KeywhizClient` otherwise) and easier to use (we'll have to write some kind of loop to hit this endpoint for every secret). The endpoint doesn't take the cert to search or replace (that's configured in the yaml file) so it seems fine.

mcpherrinm:
We could make it an automation endpoint instead of admin, and you just hit it with curl & a client cert

mcpherrinm:
This looks okay from a high level.  We should talk tomorrow about how we want to move forward with this (should we merge, or use it from a branch?)

alokmenghrajani:
Plan is to keep this in a branch.

PR #198, opened 2016-05-06
Manual status healthcheck & admin servlet to control it

mcpherrinm:
You can cause this healthcheck to go unhealthy with:
~ curl -X POST 'http://localhost:8085/status/disable'
And make it go healthy again:
~ curl -X POST 'http://localhost:8085/status/enable'

We add a servlet to the admin connector, so it's only available to localhost.

Note this doesn't change any behaviour of the server - it keeps accepting
requests.  It just causes the status check to return not-ok.

This is useful if, for example, your load balancer checks health and you want
to take a node out of the pool, perhaps because P2 is about to turn your node
off.

### DIFF
--- a/server/src/main/java/keywhiz/ManualStatusServlet.java
+++ b/server/src/main/java/keywhiz/ManualStatusServlet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015 Square, Inc.
+ * Copyright (C) 2016 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/server/src/main/java/keywhiz/ManualStatusServlet.java
+++ b/server/src/main/java/keywhiz/ManualStatusServlet.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


server/src/main/java/keywhiz/ManualStatusServlet.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
```
csstaub: nit: it's 2016

server/src/main/java/keywhiz/ManualStatusServlet.java
```diff
@@ -0,0 +1,47 @@
...
+/** A Servlet added to the Admin Connector for manually making _status report unhealthy */
+public class ManualStatusServlet extends HttpServlet implements Serializable {
+  private final ManualStatusHealthCheck mshc;
+
+  public ManualStatusServlet(ManualStatusHealthCheck mshc) {
+    this.mshc = mshc;
+  }
+
+  @Override protected void doPost(HttpServletRequest req, HttpServletResponse resp)
+      throws ServletException, IOException {
```
csstaub: Can't we make this use jax-rs like the other stuff? (i.e. have `@POST`, `@Path` annotations, etc.)

csstaub:
👍

server/src/main/java/keywhiz/ManualStatusServlet.java
mcpherrinm: Maybe, but I wasn't sure how set that up.  This servlet is pretty straightforward so I just went with this.

server/src/main/java/keywhiz/ManualStatusServlet.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015 Square, Inc.
+ * Copyright (C) 2016 Square, Inc.
```
alokmenghrajani: `Copyright (C) yyyy to present Square, Inc.`?

PR #199, opened 2016-05-09
logback 1.1.7 isn't compatible with dropwizard < 1.0.0-rc2

mcpherrinm:
Upgrade to the latest in the 0.8 train, and downgrade logback to what
we had before.  We should upgrade to a newer dropwizard soon, though


sqshh:
LGTM.

alokmenghrajani:
👍

PR #200, opened 2016-05-10
Connect to LDAP securely

### LINKED ISSUE #165: consider implementing hostname validation for ldap connections
I believe the ldap connection (https://github.com/square/keywhiz/blob/master/server/src/main/java/keywhiz/auth/ldap/LdapConnectionFactory.java) code doesn't validate the certificate. I also think we should expose a way to configure the ldap's TLS socket (supportedProtocols, supportedCipherSuites, trustStorePath, etc.) in a similar way as the socket used for https.

I could be convinced otherwise, open to comments/suggestions.

mcpherrinm:
Tell Keywhiz to validate LDAP hostnames
configure a trustStoreTrustManager with a given trust store.
Only use TLSv1.2 (hardcoded - do we want a config here too? We could have supportedProtocols, supportedCipherSuites)

Fixes #165


mcpherrinm:
R: @alokmenghrajani @csstaub @sqshh

csstaub:
LGTM

PR #201, opened 2016-05-10
Wrap socket factory to set Endpoint Identification algo

mcpherrinm:


mcpherrinm:
R?: @alokmenghrajani @csstaub @sqshh

server/src/main/java/keywhiz/auth/ldap/EndpointIdentificationSocketFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ * LDAP.  This is important for connecting to TLS through a load balancer, where we may see
+ * different certs, triggering Java8's 3shake mitigation */
+final class EndpointIdentificationSocketFactory extends SocketFactory {
+  private final SSLSocketFactory factory;
+
+  EndpointIdentificationSocketFactory(SSLSocketFactory factory) {
+    this.factory = factory;
+  }
+
+  private Socket setEndpoint(Socket socket) {
```
csstaub: `setEndpoint` is a bit misleading name

csstaub: maybe `prepareSocket`, `setParameters`, etc?

server/src/main/java/keywhiz/auth/ldap/EndpointIdentificationSocketFactory.java
```diff
@@ -0,0 +1,65 @@
...
+    SSLSocket sslSocket = (SSLSocket)socket;
+    SSLParameters parameters = sslSocket.getSSLParameters();
+    parameters.setEndpointIdentificationAlgorithm("LDAPS");
```
alokmenghrajani: There's little difference between LDAPS and HTTPS. We could make this code more re-usable by making the algorithm a parameter?

alokmenghrajani:
👍

server/src/main/java/keywhiz/auth/ldap/EndpointIdentificationSocketFactory.java
mcpherrinm: Yeah, we could -- but maybe do that when we actually need it as we'd have to make it not-private and such too.

server/src/main/java/keywhiz/auth/ldap/EndpointIdentificationSocketFactory.java
mcpherrinm: Yeah, that is a bad name.  I like prepareSocket; I'll rename to that.

PR #202, opened 2016-05-25
Fix test

alokmenghrajani:
The test assumes things are in UTC.


mcpherrinm:
seems fine by me.  This has annoyed me in IntelliJ before -- in maven, we set <argLine>${argLine} -Duser.timezone=UTC</argLine> so this shouldn't happen there.  Maybe we should do something to ensure this is always set in Keywhiz too? (app startup)

worldwise001:
👍

PR #204, opened 2016-05-26
Add expiry

alokmenghrajani:
Adds a column in the database, an expiry parameter in the admin & automation APIs
and support in keywhiz.cli.

- [ ] @csstaub 
- [ ] @mcpherrinm

### DIFF
--- a/api/src/main/java/keywhiz/api/CreateSecretRequest.java
+++ b/api/src/main/java/keywhiz/api/CreateSecretRequest.java
@@ -47,20 +47,26 @@ public class CreateSecretRequest {
   @JsonProperty
   public final ImmutableMap<String, String> metadata;
 
+  /** In seconds since epoch */
+  @JsonProperty
+  public final long expiry;
+
   public CreateSecretRequest(@JsonProperty("name") String name,
       @JsonProperty("description") @Nullable String description,
       @JsonProperty("content") String content,
       @JsonProperty("withVersion") boolean withVersion,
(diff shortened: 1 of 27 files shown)


api/src/main/java/keywhiz/api/CreateSecretRequest.java
```diff
@@ -47,20 +47,25 @@
   @JsonProperty
   public final ImmutableMap<String, String> metadata;
 
+  @JsonProperty
+  public final long expiry;
```
csstaub: What format is this? Seconds since epoch? Millis since epoch?

cli/src/main/java/keywhiz/cli/commands/AddAction.java
```diff
@@ -166,6 +167,19 @@ private void createAndAssignSecret(String secretName, byte[] content, boolean us
     return metadata;
   }
 
+  private long getExpiry() {
+    String expiry = addActionConfig.expiry;
+    if (expiry != null && !expiry.isEmpty()) {
```
csstaub: On `expiry.isEmpty()` should probably throw an error, because it means caller passed `--expiry=''` or such which is incorrect and should fail.

csstaub:
Added some comments, but otherwise ✅.

api/src/main/java/keywhiz/api/CreateSecretRequest.java
alokmenghrajani: Seconds since epoch. Adding comment.

PR #205, opened 2016-05-27
Delete SecretGenerator

alokmenghrajani:
We don't use this feature. We have also failed to document it, so probably
nobody else uses it.

- [ ] @csstaub 
- [ ] @mcpherrinm


mcpherrinm:
Solid +1 to deleting unused code!

csstaub:
✅

PR #206, opened 2016-05-31
Log if versioned api is being used.

alokmenghrajani:
We are deleting all the secret versioning logic and replacing it with
transparent versioning. It's going to result in a simpler and easier to
use API.

If you depend on the current versioned secrets logic, you might end
up having to manually migrate data. We can try to help you on our
mailing list.

- [ ] @csstaub 
- [x] @mcpherrinm


mcpherrinm:
🚢

PR #209, opened 2016-06-01
Delete versioning code.

alokmenghrajani:
- [ ] @mcpherrinm


mcpherrinm:
:ship:

We should make sure to add a release note

PR #215, opened 2016-06-07
Support to track current active version for a given secret

alokmenghrajani:
- [ ] @mcpherrinm


api/src/main/java/keywhiz/api/model/SecretSeries.java
```diff
@@ -58,4 +59,5 @@ public static SecretSeries of(
   public abstract String updatedBy();
   public abstract Optional<String> type();
   public abstract ImmutableMap<String, String> generationOptions();
+  public abstract Optional<Long> current();
```
csstaub: Nit: call this `currentVersion`? Otherwise hard to know what this actually is.

server/src/main/java/keywhiz/auth/xsrf/XsrfServletFilter.java
```diff
@@ -39,7 +39,7 @@
 @Singleton
 public class XsrfServletFilter implements Filter {
   private static final Logger logger = LoggerFactory.getLogger(XsrfServletFilter.class);
-  private static final Set<String> EXCLUDED_PATHS = ImmutableSet.of("/admin/login", "/admin/logout");
+  private static final Set<String> EXCLUDED_PATHS = ImmutableSet.of("/admin/login", "/admin/logout", "/admin/migrate/current");
```
csstaub: Why do we exclude this from XSRF protection?

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -80,6 +80,13 @@ long createSecretSeries(String name, String creator, String description, @Nullab
     return r.getId();
   }
 
+  public int setCurrent(long secretId, long secretContentId) {
```
mcpherrinm: Should we have a check here that the secretContent is for that secret?

server/src/main/java/keywhiz/auth/xsrf/XsrfServletFilter.java
alokmenghrajani: Because I want to use `curl`.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
alokmenghrajani: It will imply an extra round trip to the database but seems like a good thing to have.

csstaub:
👍

api/src/main/java/keywhiz/api/model/SecretSeries.java
alokmenghrajani: fixed

PR #217, opened 2016-06-09
Make sure fetching secrets by name honors current.

alokmenghrajani:


server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -170,6 +169,7 @@ public long createSecret(String name, String encryptedSecret,
   public ImmutableList<SimpleEntry<Long, String>> getSecretsNameOnly() {
     List<SimpleEntry<Long, String>> results = dslContext.select(SECRETS.ID, SECRETS.NAME)
         .from(SECRETS)
+        .where(SECRETS.CURRENT.isNotNull())
```
csstaub: This guaranteed for all secrets?

mcpherrinm:
lgtm

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -135,16 +135,15 @@ public long createSecret(String name, String encryptedSecret,
...
+      }
 
-    Optional<SecretContent> secretContent = secretContentDAO.getSecretContentBySecretIdOne(secretSeries.get().id());
-    if (!secretContent.isPresent()) {
-      return Optional.empty();
+      return Optional.of(SecretSeriesAndContent.of(secretSeries.get(), secretContent.get()));
     }
-
-    return Optional.of(SecretSeriesAndContent.of(secretSeries.get(), secretContent.get()));
+    return Optional.empty();
```
csstaub: Imho there are too many nested if's here so I'm not sure what the desired logic should be, but I think it looks like it does what I would expect.

csstaub:
👍

server/src/main/java/keywhiz/service/daos/SecretDAO.java
alokmenghrajani: It's currently the case for all secrets (except the few broken ones). Going forward, we will stop deleting secrets and set current to null instead.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
alokmenghrajani: if the row exists in the secrets table and the row's current is not null, then we treat the secret as present. Otherwise, the code should behave as if the secret doesn't exist.

PR #218, opened 2016-06-09
Make SecretContentDAO and SecretSeriesDAO visbility package-only

alokmenghrajani:
The code is lower level than SecretDAO and delegates some checks to SecretDAO. I think
it's cleaner to downgrade their visibiliy. Also fix up some code which was referred to these
classes instead of SecretDAO.


mcpherrinm:
Lgtm

PR #219, opened 2016-06-13
Set current to null instead of deleting secrets.

alokmenghrajani:
Gives the ability to revert accidental deletions.

- [ ] @mcpherrinm 
- [x] @csstaub


csstaub:
LGTM

PR #222, opened 2016-06-21
Fix bug when fetching secrets by name.

alokmenghrajani:
The code was returning a random (oldest in practice) secret when
multiple secrets existed.

Made sure test fails without the fix.


mcpherrinm:
LGTM.

Listing secrets, however, is still broken - it lists all the versions

PR #223, opened 2016-06-21
Cache status checks.

alokmenghrajani:
Also adds some tests for the /status/disable and /status/enable endpoints.

- [ ] @csstaub 
- [ ] @mcpherrinm

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/StatusResource.java
+++ b/server/src/main/java/keywhiz/service/resources/StatusResource.java
@@ -3,17 +3,23 @@ package keywhiz.service.resources;
 import com.codahale.metrics.annotation.ExceptionMetered;
 import com.codahale.metrics.annotation.Timed;
 import com.codahale.metrics.health.HealthCheck;
-import com.codahale.metrics.health.HealthCheckRegistry;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import io.dropwizard.setup.Environment;
+
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedMap;
+import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
+
+import keywhiz.KeywhizConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


server/src/main/java/keywhiz/service/resources/StatusResource.java
```diff
@@ -26,12 +36,50 @@
...
   private static final Logger logger = LoggerFactory.getLogger(SecretDeliveryResource.class);
+  LoadingCache<HealthCheckRegistry, SortedMap<String, HealthCheck.Result>> cachedResult;
 
-  @Inject public StatusResource(Environment environment) {
+  @Inject public StatusResource(KeywhizConfig keywhizConfig, Environment environment) {
     this.environment = environment;
+
+    Duration cacheExpiry = keywhizConfig.getStatusCacheExpiry();
+
+    cachedResult = CacheBuilder.newBuilder()
```
csstaub: Prefer `Suppliers.memoizeWithExpiration` instead of a cache: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Suppliers.html

server/src/main/resources/keywhiz-development.yaml.postgres
```diff
@@ -86,6 +86,8 @@ readonlyDatabase:
 migrationsDir:
   db/postgres/migration
 
+statusCacheExpiry: PT60S
```
csstaub: Hard-code this, or else make sure to provide a decent default value. We have too many config options as it is already.

server/src/main/java/keywhiz/service/resources/StatusResource.java
csstaub: Also: this class is not a singleton, so the cache/memoization will get lost since the class will likely get reinstantiated on every request.

alokmenghrajani: The integration test passes. Are you sure we need to mark it as singleton?

server/src/main/resources/keywhiz-development.yaml.postgres
mcpherrinm: 60 seconds is way too long.  1 second may be an appropriate value.

alokmenghrajani: Make it default to 1 second. Also set it to 1 second in the yaml files.

server/src/main/java/keywhiz/service/resources/StatusResource.java
csstaub: Nope. But are you sure the test is checking what you think it is?  :)

alokmenghrajani: Yes, I used a breakpoint. Maybe `jersey.register(injector.getInstance(Foo.class));` acts as a singleton?

csstaub: seems good to me then

csstaub:
LGTM

PR #224, opened 2016-06-24
Optimize listing secrets

alokmenghrajani:
This change should reduce the number of datbase round trips. It should
also reduce the amount of data we fetch from the database.


csstaub:
LGTM

PR #228, opened 2016-06-24
Delete old code, add some indexes.

alokmenghrajani:
Ran both versions for a bit and it seems the new code is correct and faster.


csstaub:
LGTM

PR #229, opened 2016-06-27
Optimize getSanitizedSecretFor(client, secretName)

alokmenghrajani:


mcpherrinm:
This test looks fine, however, I'm not sure it's the best change we could make.

Ignoring tests, it seems this function is only used at https://github.com/square/keywhiz/blob/master/server/src/main/java/keywhiz/service/resources/SecretDeliveryResource.java#L94-L95
and is immediately followed by getting the whole secret -- whether or not the client had access, we load the secret from the DB!

It seems to me that we could combine all that into a single database fetch, and so we'd only spend the bandwidth and HSM usage returning the row if the client has access.  Not to mention the code would simplify.

🚢  if you want, but if you're feeling up to it, we could improve this path more.

alokmenghrajani:
Yes, the reason we fetch the row twice is to distinguish between a permission error (403) and not found (404).

We can perhaps delete SecretTransformer and have Secret lazily decrypt the content when needed?

PR #232, opened 2016-06-29
Delete old code.

alokmenghrajani:
Ran getSanitizedSecretForFast for a bit and it seems to be correct.


mcpherrinm:
gif of a ship

PR #233, opened 2016-06-29
Fix various logger bugs

worldwise001:
Here's an attempt at a fix based on what I grepped through the repo.

r: @mcpherrinm @csstaub @alokmenghrajani


alokmenghrajani:
These aren't bugs but format is nicer/saner. 🚢

csstaub:
🚢 ✅

PR #234, opened 2016-06-30
Ensure cli sets expiry in seconds.

alokmenghrajani:


worldwise001:
LGTM

api/src/main/java/keywhiz/api/model/SanitizedSecret.java
```diff
@@ -93,19 +95,8 @@ public static SanitizedSecret fromSecret(Secret secret) {
         secret.getUpdatedBy(),
         secret.getMetadata(),
         secret.getType().orElse(null),
-        secret.getGenerationOptions());
-  }
-
-  /**
-   * Helper method to convert a collection of secrets to a list of sanitized secrets.
-   */
-  public static List<SanitizedSecret> fromSecrets(List<Secret> secrets) {
```
worldwise001: Is removing this fine?

alokmenghrajani: looks like dead code to me. Wouldn't compile if it wasn't fine.

Issue #180, opened 2016-03-14
Keywhiz should automatically handle CookieKeys


mcpherrinm:
Keywhiz should not require a configured CookieKey.  It should automatically generate one if not present, and store it in the database like any other secret.

csstaub:
+1, let's reduce configuration to a minimum and just use sane defaults whenever possible.

PR #235, opened 2016-07-05
Add check for missing cookieKey, generate one if not specified

### LINKED ISSUE #190: Null Pointer Exception if you start without a config file
Exception in thread "main" java.lang.IllegalArgumentException: (was java.lang.NullPointerException) (through reference chain: keywhiz.KeywhizConfig["cookieKey"])

We should print an error message, not just NPE.

I'm not sure if this is something specific to the cookieKey, if so, Issue #180 would fix it.

worldwise001:
Should fix #190 and #180 .

r: @mcpherrinm @csstaub @alokmenghrajani

### DIFF
--- a/server/src/main/java/keywhiz/ServiceModule.java
+++ b/server/src/main/java/keywhiz/ServiceModule.java
@@ -25,6 +25,7 @@ import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.db.ManagedDataSource;
 import io.dropwizard.java8.auth.Authenticator;
 import io.dropwizard.setup.Environment;
+import java.security.SecureRandom;
 import java.sql.SQLException;
 import java.time.Clock;
 import keywhiz.auth.BouncyCastle;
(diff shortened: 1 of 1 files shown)


mcpherrinm:
Could we deterministically generate one, perhaps like we do the per-secret keys?

Randomly generating one is going to lead to confusion and sadness as your cookies are only valid on the server that generated it.

worldwise001:
Yeah we could. What would we use as a stable value to deterministically generate one reliably on multiple servers?

I guess the other consideration is that if you're going to configure more than one, maybe you will need to specify one manually.

mcpherrinm:
We have a master key that we derive keys per-secret -- we could derive a key for cookies too.
However, we'd have to be careful whatever we use for derivation doesn't collide with the name of a secret.

see  https://github.com/square/keywhiz/blob/master/server/src/main/java/keywhiz/service/crypto/ContentCryptographer.java

server/src/main/java/keywhiz/ServiceModule.java
```diff
@@ -75,6 +87,14 @@ public ServiceModule(KeywhizConfig config, Environment environment) {
     bind(KeywhizConfig.class).toInstance(config);
   }
 
+  private String generateCookieKey() {
```
csstaub: This means the cookie key won't work across hosts, and it'll also get wiped on restart and everyone ends up being logged out?

worldwise001:
We can derive off the master key. I'll submit an update.

worldwise001:
Whoops I completely forgot about this. I'll see if I can get this updated today.

mcpherrinm:
Abandoned PR

PR #238, opened 2016-07-07
Prune old/unreferenced secret contents after 45 days

csstaub:
r: @alokmenghrajani @worldwise001 @mcpherrinm 
Prune old/unreferenced secret contents after 45 days.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
@@ -18,8 +18,8 @@ package keywhiz.service.daos;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -74,12 +78,39 @@ public long createSecretContent(long secretId, String encryptedContent,
     r.setExpiry(expiry);
     r.store();
 
+    pruneOldContents(secretId);
```
alokmenghrajani: I would prefer to do last N instead of last N days. Otherwise there are two risks: we might not keep the previous secret (if it was older than 45 days) or we might keep too many old secrets.

We can use the auto increment ID to figure out the order in which the content was created.

mcpherrinm: A combination of both might be cool?  So keep at least 10 secrets, and 45 days.  Then you don't risk automation gone crazy wiping out a secret by creating 11 bad versions -- as unlikely as that is.

csstaub:
@alokmenghrajani @mcpherrinm PTAL; updated this to keep last 10 + 45 days now.

alokmenghrajani:
👍

mcpherrinm:
lgtm

PR #239, opened 2016-07-14
Secret/client detail endpoints for group

csstaub:
@alokmenghrajani @mcpherrinm @worldwise001
Secret/client detail endpoints for group


alokmenghrajani:
👍

PR #242, opened 2016-07-26
Allow -Dcustom.build.directory=foo to override the normal target folder

mcpherrinm:
This is useful if you want to do multiple builds in the same source tree (but perhaps, with different profiles)

Example usage:
mvn package -am -pl server -pl cli -P h2 -Dcustom.build.directory=h2-target


csstaub:
LGTM

PR #244, opened 2016-07-28
Adds an API for listing secrets that are expiring soon

stfinney:
"Soon" is defined by argument to API call specifying latest expiry to include. Also adds option to limit listing to a specific group. Results are sorted by expiry (soonest first).

r: @mcpherrinm @worldwise001 @csstaub


server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -171,17 +168,19 @@ public long createOrUpdateSecret(String name, String encryptedSecret, String cre
     return Optional.empty();
   }
 
-  /** @return all existing secrets. */
-  public ImmutableList<SecretSeriesAndContent> getSecrets() {
+  /** @return list of secrets. can limit/sort by expiry, and for group if given */
+  public ImmutableList<SecretSeriesAndContent> getSecrets(boolean sortByExpiry, Long expireMaxTime, Group group) {
```
csstaub: Make this `@Nullable Long expireMaxTime` (or if it's not nullable, make it `long` instead of `Long`).

csstaub: Also, I'm not a huge fan of boolean flags in args. Can we just make it sort by expiry automatically if expire max time is set? If not, why not?

stfinney: I can remove the flag. Now that I think about it, it's sort of an artifact of how I was originally going to do this, not necessary.

server/src/main/java/keywhiz/service/daos/SecretController.java
```diff
@@ -57,14 +58,19 @@ public SecretController(SecretTransformer transformer, ContentCryptographer cryp
...
-  /** @return all existing sanitized secrets. */
-  public List<SanitizedSecret> getSanitizedSecrets() {
-    return secretDAO.getSecrets().stream()
+  /**
+   * @param sortByExpiry whether or not to limit to soon to expire secrets and sort by date.
+   * @param expireMaxTime timestamp for farthest expiry to include
+   * @param group limit results to secrets assigned to this group, if provided.
+   * @return all existing sanitized secrets matching criteria.
+   * */
+  public List<SanitizedSecret> getSanitizedSecrets(boolean sortByExpiry, Long expireMaxTime, Group group) {
```
mcpherrinm: Could we use an Option<Long> expireMaxTime here instead of a bool + a long

mcpherrinm:
(lgtm was given in person)

PR #246, opened 2016-08-09
Add function to retrieve latest secret via automation API

csstaub:
r: @stfinney @worldwise001
Add function to retrieve latest secret via automation API


server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -215,6 +215,28 @@ public SecretDetailResponseV2 secretInfo(@Auth AutomationClient automationClient
   }
 
   /**
+   * Retrieve latest version of a secret
```
mcpherrinm: Prefer "current" to latest.  Though in general it's latest, we'd like to support rolling back (and forward) versions at some point.

csstaub:
not needed anymore

PR #247, opened 2016-08-09
Fixes the test secretListingExpiry_success

stfinney:
Original version of this test had hardcoded timestamps, which are now in the past. Thus, all the secrets added in the test are already expired when the test runs, so none are retrieved and the test fails. The person who wrote this test is clearly an idiot.

Yes, it was me... :(

Fixed to use times relative to current time.

r: @csstaub @worldwise001


csstaub:
:+1:

PR #248, opened 2016-08-11
Adds endpoint to SecretResource which displays secret versions

jbpeirce:
SecretResource now accepts GET requests for a list of versions of a given secret, ordered with the most recently created version first.  This list can be restricted by passing indices to the GET request; it will then return only the versions between these indices in the list (inclusive).


server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -198,6 +198,33 @@ public long createOrUpdateSecret(String name, String encryptedSecret, String cre
...
+   * @param newestIdx the first index to select in a list of versions sorted by creation time
+   * @param oldestIdx the last index to select in a list of versions sorted by creation time
+   * @return Versions of a secret matching input parameters or Optional.absent().
+   */
+  public Optional<SecretSeriesAndVersions> getSecretVersionsByName(String name, int newestIdx,
+      int oldestIdx) {
+    checkArgument(!name.isEmpty());
+    checkArgument(newestIdx >= 0);
+    checkArgument(oldestIdx >= 0);
+
```
stfinney: What is the behavior if I ask for newest=10 oldest=1? You should probably have a check requiring oldest > newest, if oldest != 0. Otherwise, looking at the query in getSecretVersionsBySecretId, I believe you end up with a negative value in the limit clause, which will either result in an exception from jooq or an error from mysql.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -130,6 +133,30 @@ public long createSecretContent(long secretId, String encryptedContent,
...
+    if (oldestIdx == 0) {
+      // Retrieve all rows from newestIdx to the end
+      oldestIdx =
+          Integer.MAX_VALUE - 1; // -1 to compensate for + 1 in calculating range to retrieve
+    }
+
+    Result<SecretsContentRecord> r = dslContext.selectFrom(SECRETS_CONTENT)
+        .where(SECRETS_CONTENT.SECRETID.eq(id))
+        .orderBy(SECRETS_CONTENT.CREATEDAT.desc())
+        .limit(newestIdx, oldestIdx - newestIdx + 1)
```
stfinney: Given that mysql handles limits on query results as an offset + limit, why not implement your arguments (oldestIdx and newestIdx) in that way? Then the logic on 139-143 is unnecessary, you don't have to worry about negative values if oldest is less than newest, and (to me at least) the arguments to the API call will be more intuitive.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -198,6 +198,33 @@ public long createOrUpdateSecret(String name, String encryptedSecret, String cre
...
+    SecretContentDAO secretContentDAO = secretContentDAOFactory.using(dslContext.configuration());
+    SecretSeriesDAO secretSeriesDAO = secretSeriesDAOFactory.using(dslContext.configuration());
+
+    Optional<SecretSeries> series = secretSeriesDAO.getSecretSeriesByName(name);
+    if (series.isPresent()) {
+      long secretId = series.get().id();
+      Optional<ImmutableList<SecretContent>> contents =
+          secretContentDAO.getSecretVersionsBySecretId(secretId, newestIdx, oldestIdx);
```
stfinney: Could this be made simpler/more efficient by using a join to get the secret versions by the name in one query, rather than using the name to get an id and using the id to get the versions? It's only two tables, with a primary/foreign key link, so I don't think a join would be over-complicated.

api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
```diff
@@ -64,9 +68,21 @@ public Builder secret(Secret secret) {
           .createdAtSeconds(secret.getCreatedAt().toEpochSecond())
           .createdBy(secret.getCreatedBy())
           .type(secret.getType().orElse(null))
+          .expiry(secret.getExpiry())
           .metadata(secret.getMetadata());
     }
 
+    public Builder secretContent(SecretContent secretContent) {
```
stfinney: The name seems slightly confusing given it has no actual secret content, just metadata.

api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
```diff
@@ -117,6 +136,7 @@ public SecretDetailResponseV2 build() {
   @Override public final String toString() {
     return MoreObjects.toStringHelper(this)
         .add("name", name())
+        .add("version", version())   // TODO: Should we remove this so it's not as exposed to the user?
```
stfinney: If you don't think we want to expose it to the user, then I wouldn't include it at all. We can always add it later if there does seem to be a need for it. However, if this is part of building the response, then don't we need to include the version so it can be used for choosing prior versions to roll back to? Or at least, you need some kind of unique identifier for each version to enable that feature. If we don't want to expose the real primary key of that table, you could use sequential integers for each version (sorted), or something of that nature.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -215,6 +216,33 @@ public SecretDetailResponseV2 secretInfo(@Auth AutomationClient automationClient
...
+   * oldest update time and ranging from "newestIdx" to "oldestIdx", inclusive.  Leaving
+   * "newestIdx" at 0 will retrieve "oldestIdx" versions starting from the
+   * current version; leaving "oldestIdx" at 0 will retrieve all versions after
+   * "newestIdx".
+   *
+   * @param name Secret series name
+   * @param newestIdx The index in the list of versions of the first version to retrieve
+   * @param oldestIdx The index in the list of versions of the last version to retrieve
+   * @excludeParams automationClient
+   */
```
stfinney: Comment should include info about the response message(s).

api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
jbpeirce: We need the version as part of the SecretDetailResponseV2, since as far as I can tell it's the only easy way to retrieve a particular version.  I'm not sure whether we should put it in the toString method, though.  Having it there could be useful for debugging, but it also exposes that identifier to a user.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
jbpeirce: Good point--I'll make that change.

api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
jbpeirce: Hmm.  I named it for parallelism with the methods series(SecretSeries) and secret(Secret), but silently discarding the encrypted content could be confusing (the other two methods fill the content() field).  For now, I'll just leave the field out.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
jbpeirce: I've been experimenting with this, but it looks like we'd have to either take a generic Record from the query and process it into something useful (which could be hazardous if it isn't formatted as we expect) or define a new class to hold the information on the secret retrieved by the query, which I'm reluctant to do since we already have a lot of representations of secrets.  I'm not very familiar with the cost of queries in jOOQ--is avoiding an extra query worth these changes?

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -314,6 +397,18 @@ Response createOrUpdate(CreateOrUpdateSecretRequestV2 request, String name) thro
     return mapper.readValue(httpResponse.body().byteStream(), new TypeReference<List<String>>(){});
   }
 
+  List<SecretDetailResponseV2> listVersions(String name, int newestIdx, int oldestIdx)
```
jbpeirce: I'll rename these parameters to versionIdx and numVersions in the next commit, for clarity.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -275,6 +277,87 @@
...
 
+  @Test public void secretVersionListing_notFound() throws Exception {
+    Request put = clientRequest("/automation/v2/secrets/non-existent/versions/0-0").build();
+    Response httpResponse = mutualSslClient.newCall(put).execute();
+    assertThat(httpResponse.code()).isEqualTo(404);
+  }
+
+  @Test public void secretVersionListing_success() throws Exception {
+    int totalVersions = 6;
+    int sleepInterval = 1000; // Delay so secrets have different creation timestamps
```
worldwise001: what unit is this in?

tkordas: to me it would make more sense to set the timestamps explicitly rather than waiting like this.

worldwise001: Yeah actually that would make more sense.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -275,6 +277,87 @@
...
+    List<SecretDetailResponseV2> versions;
+    assertThat(listing()).doesNotContain("secret20");
+
+    // get current time to calculate timestamps off for expiry
+    long now = System.currentTimeMillis() / 1000L;
+
+    // Create secrets 1 second apart, so their creation order is deterministic
```
worldwise001: If this is not being deterministic, might be worth to investigate more than to put an arbitrary sleep in.

worldwise001:
The rest of it looks fine. The only thing is the arbitrary sleep thing.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
jbpeirce: Unfortunately, the secret's update and creation timestamps aren't affected by any of the requests to create or update a secret--it looks like Keywhiz automatically sets the timestamps.  I agree it'd be much better to set them manually, though; is there a workaround I could use?  (I could write and implement a new request to reset secret creation and update times, but I'm not sure whether that's worth it.)

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
jbpeirce: The comment wasn't very clear--when I retrieve versions of a secret, they're ordered by their creation time, which is stored only to the second.  To make the versions have a predictable ordering, I had to separate the creation times by a second.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -275,6 +279,145 @@
...
+  @Test public void secretVersionListing_success() throws Exception {
+    int totalVersions = 6;
+    int sleepInterval = 1000; // Delay so secrets have different creation timestamps
+    List<SecretDetailResponseV2> versions;
+    assertThat(listing()).doesNotContain("secret20");
+
+    // get current time to calculate timestamps off for expiry
+    long now = System.currentTimeMillis() / 1000L;
+
+    // Create secrets 1 second apart, so their version order is deterministic
```
stfinney: Minor thing, but I think it's worth expanding this comment to explain why it's necessary to put a 1 second delay to make them be ordered correctly. Otherwise it kind of sounds like it's a concurrency issue/race condition/etc.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -198,6 +198,52 @@ public long createOrUpdateSecret(String name, String encryptedSecret, String cre
...
+    }
+    return Optional.empty();
+  }
+
+  /**
+   * @param name of secret series for which to reset secret version
+   * @param versionId The identifier for the desired current version
+   * @return 0 for success, non-zero for failure (secret series not found)
+   */
+  public int setCurrentSecretVersionByName(String name, long versionId) {
```
csstaub: Instead of returning 0/1, just have it throw an exception in case of failure and make the return type `void`.

api/src/main/java/keywhiz/api/model/SecretSeriesAndVersions.java
```diff
@@ -0,0 +1,33 @@
...
+ * Value type containing a secret series and associated list of contents
+ */
+@AutoValue
+public abstract class SecretSeriesAndVersions {
+  public static SecretSeriesAndVersions of(SecretSeries series, ImmutableList<SecretContent> content) {
+    return new AutoValue_SecretSeriesAndVersions(series, content);
+  }
+
+  public abstract SecretSeries series();
+  public abstract ImmutableList<SecretContent> content();
```
csstaub: Since we don't expect this to return any content, we should probably introduce a new type that only holds the relevant information, e.g. `SecretVersion` or something like that.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
worldwise001: Was this addressed? I would also prefer an exception being thrown here.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
worldwise001: Normally in this situation I'd use a fake clock, but that appears to not exist in this world.

It looks like in general we don't have great things to test for timestamps correctly (and adding that is out of the scope of this PR). I think I would just manually munge the timestamps in SQL if possible to get around this for now, and open an issue to get better test coverage of timestamps for the future.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
jbpeirce: The next commit will have an exception thrown rather than the 0/1 return--it is significantly cleaner this way.

csstaub:
LGTM

worldwise001:
Looks fine to me.

stfinney:
👍

PR #249, opened 2016-08-12
Fix GitHub URL

davidjb:
The href was originally relative due to no protocol; was a 404.


worldwise001:
Great, thanks for the contribution!

PR #250, opened 2016-08-17
Allow : in names

csstaub:
Allow : in names


worldwise001:
👍 :shipit:

alokmenghrajani:
Move this code to SecretController.java and validate server side?

csstaub:
@alokmenghrajani Yes, we should probably do that; but in the meantime I need this fix to get in because... reasons.

PR #252, opened 2016-08-20
Allows for custom metadata entries on new secrets in CLI

stfinney:
Includes some restrictions on naming keys.

This is the only place I could find that put any restriction on what could be included in metadata.

r: @csstaub @worldwise001 @mcpherrinm @jbpeirce


csstaub:
LGTM

PR #253, opened 2016-08-24
Fix bug in setCurrentVersion's handling of invalid version numbers

jbpeirce:
If `setCurrentVersion` had been called with a version number that did not correspond to an ID in the `secrets_content` table, it would have thrown a `NullPointerException` when `fetchOne()` returned `null`.  This is now checked for.


csstaub:
Looks good ✅

PR #254, opened 2016-08-25
Make SetSecretVersion take a JSON request as input

jbpeirce:
This makes SetSecretVersion symmetric with CreateSecret, UpdateSecret, and
the SA implementation, and reduces redundant information in the URL.


jbpeirce:
@csstaub @stfinney

csstaub:
Looks good to me

PR #257, opened 2016-08-31
Downgrade some warn logs to info

csstaub:
@stfinney @jbpeirce @worldwise001 @mcpherrinm 
Downgrade some warn logs to info


jbpeirce:
LGTM

PR #258, opened 2016-09-06
Instruct people to add-user instead of db-seed

mcpherrinm:
db-seed puts a bunch of crap data in your database.  It's useful for unit tests, but not for users.

However, some people rely on it for adding an admin user.  We should explicitly say to do that instead.


worldwise001:
👍

csstaub:
add to readme too?

mcpherrinm:
Updated readme and dockerfile instructions too

PR #261, opened 2016-09-07
Add index on secrets_content.secretid

csstaub:
@mcpherrinm @jbpeirce 
Add index on secrets_content.secretid


mcpherrinm:
LGTM.

For posterity:  We added "pruning" of old secret versions, which had a slow query, fixed by this index.

PR #264, opened 2016-09-09
added dev startup script for docker

gyulaweber:
Hey,

It took me a while to figure out, how to start the development version to try out keywhiz, so I'd like to ease it a bit; 

Please let me know what do you think, or how to make it better.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,11 @@
 #     docker volume create --name keywhiz-db-devel
 #
 #   Initialize the database, apply migrations, and add administrative user:
-#     scripts/start-dev migrate
-#     scripts/start-dev add-user
+#     scripts/start-dev.sh migrate
+#     scripts/start-dev.sh add-user
 #
 #   Finally, run the server with the default development config:
-#     scripts/start-dev server 
+#     scripts/start-dev.sh server 
 #
 # *** Production setup wizard ***
 #   For production deployments, we have setup wizard that will initialize
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,11 @@
 #     docker volume create --name keywhiz-db-devel
 #
 #   Initialize the database, apply migrations, and add administrative user:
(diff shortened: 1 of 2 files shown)


csstaub:
There's already an entry script (see `docker/entry.sh`) that should pick up the development config automatically if you launch it without an explicit one. I'd prefer to make that script better instead of adding a second one.

gyulaweber:
Thanks, you're right. I've tried modifying, but it doesn't works, I don't really know why (but it should, because the functionality is implemented in the docker/entry.sh, it sets the default config file, migration works, but add-user don't). 

Anyway, if you think that the making the entrypoint script able to do that is a better way than just writing a wrapper, then I'll try to make it work with that, and feel free to close this PR.

csstaub:
Do you have error messages from failures you got? If you can post an error I can try to reproduce it and fix/update the entry script.

gyulaweber:
Finally I've found the real problem :) It was caused by the missing -ti parameter in the docker run, when calling add-use, which is expecting terminal input (so the example in the Dockerfile should use the ```-ti``` flag when calling add-user.

The error message when there is no ``` -ti ``` flag in docker run: 

```
------------------------------------------------------------------------
---    No configuration file specified, using development config!    ---
--- Use the KEYWHIZ_CONFIG environment variable to set a config file ---
------------------------------------------------------------------------
New username:Exception in thread "main" java.lang.NullPointerException
        at keywhiz.commands.AddUserCommand.run(AddUserCommand.java:41)
        at keywhiz.commands.AddUserCommand.run(AddUserCommand.java:29)
        at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:76)
        at io.dropwizard.cli.Cli.run(Cli.java:70)
        at io.dropwizard.Application.run(Application.java:73)
        at keywhiz.KeywhizService.main(KeywhizService.java:71)
```

PR #265, opened 2016-09-09
Support PKCS12/JCEKS/JKS backfilling also

csstaub:
Support PKCS12/JCEKS/JKS backfilling also


server/src/main/java/keywhiz/service/resources/automation/v2/ExpirationExtractor.java
```diff
@@ -17,11 +21,57 @@
...
+            }
+          }
+        }
+      }
+    } catch (KeyStoreException e) {
+      // Should never occur (ks was initialized)
+      throw Throwables.propagate(e);
+    }
+
+    return earliest;
```
worldwise001: Is there an instant where earliest is null? If so do we want to annotate the definition with @Nullable?

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -201,28 +202,41 @@ public Response createOrUpdateSecret(@Auth AutomationClient automationClient,
...
+    if (password == null) {
+      password = "";
+    }
 
     Instant expiry = null;
-    if (secret.getName().endsWith(".crt") || secret.getName().endsWith(".pem")) {
-      expiry = ExpirationExtractor.expirationFromEncodedCertificateChain(content);
-    } else if (secret.getName().endsWith(".gpg") || secret.getName().endsWith(".pgp")) {
-      expiry = ExpirationExtractor.expirationFromOpenPGP(content);
+    if (secretName.endsWith(".crt") || secretName.endsWith(".pem") || secretName.endsWith(".key")) {
```
worldwise001: Do we want to add support for ".cert"?

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -235,6 +236,7 @@
   public void backfillExpirationTest() throws Exception {
     byte[] certs = Resources.toByteArray(Resources.getResource("fixtures/expiring-certificates.crt"));
     byte[] pubring = Resources.toByteArray(Resources.getResource("fixtures/expiring-pubring.gpg"));
+    byte[] p12 = Resources.toByteArray(Resources.getResource("fixtures/expiring-keystore.p12"));
```
worldwise001: We should probably have tests for both JCEKS and JKS stores.

server/src/main/java/keywhiz/service/resources/automation/v2/ExpirationExtractor.java
```diff
@@ -17,11 +21,57 @@
...
+      ks = KeyStore.getInstance(type);
+    } catch (KeyStoreException e) {
+      // Should never occur (assuming JCE is installed)
+      throw Throwables.propagate(e);
+    }
+
+    try {
+      ks.load(new ByteArrayInputStream(content), password.toCharArray());
+    } catch (IOException | NoSuchAlgorithmException | CertificateException e) {
+      // Failed to parse
```
worldwise001: Do you want to add a warning that it failed to parse instead of silently failing?

csstaub: It returns true/false to the caller. I can log the exception if you want, but I'd rather not warn.

worldwise001: Let's log the exception as info then.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
csstaub: I added JCEKS, but I don't have a JKS example handy right now... but since it hits the same code path anyway, do we really need an example of every file?

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
csstaub: Never seen that used, but can add it I suppose.

server/src/main/java/keywhiz/service/resources/automation/v2/ExpirationExtractor.java
csstaub: added

server/src/main/java/keywhiz/service/resources/automation/v2/ExpirationExtractor.java
csstaub: done

csstaub:
@worldwise001 lgty?

mcpherrinm:
LGTM

PR #267, opened 2016-09-21
Alter SanitizedSecret to contain group names for secret

jbpeirce:
SanitizedSecret responses now include a list of the names of the groups with which a secret is associated.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -198,12 +198,16 @@ public class AclDAO {
   }
 
   public Set<Group> getGroupsFor(Secret secret) {
+    return getGroupsFor(secret.getName());
+  }
+
+  public Set<Group> getGroupsFor(String secretName) {
     List<Group> r = dslContext
         .select(GROUPS.fields())
         .from(GROUPS)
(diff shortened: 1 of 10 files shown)


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -263,7 +266,17 @@ public void findAndEvictClient(long clientId, long groupId) {
               secretContentMapper.tryToReadMapFromMetadata(row.getValue(SECRETS_CONTENT.METADATA)),
               series.type().orElse(null),
               series.generationOptions(),
-              row.getValue(SECRETS_CONTENT.EXPIRY));
+              row.getValue(SECRETS_CONTENT.EXPIRY),
+              Collections.emptyList()
+          );
+          LazyString lazyString = () -> "";
+          Set<Group> groups = getGroupsFor(
+              new Secret(secret.id(), secret.name(), secret.description(), lazyString, secret.createdAt(),
```
csstaub: Instead of doing this here (and below), we should just have an API somewhere that fetches all the groups based on a secret name. If you look at `getGroupsFor`, it just reads the name of the secret and doesn't care about anything else. It was probably typed to take a `Secret` just for convenience.

jbpeirce: Good idea.

PR #268, opened 2016-09-22
Improve secret data printing from keywhiz-cli

jbpeirce:


mcpherrinm:
Can you paste some sample output here?

jbpeirce:
Sample output (note this secret doesn't have CreatedBy and UpdatedBy set; none of the test secrets do)
```
./cli/target/keywhiz-cli*shaded.jar --url https://localhost:4444 --devTrustStore --user keywhizAdmin describe secret --name NonexistentOwner_Pass
NonexistentOwner_Pass
	Groups:
		Security
		Web
	Clients:
		CN=User1
		CN=User2
		CN=User3
		CN=User4
		client
	Metadata:
		{owner=NonExistant, mode=0400}
	Created at:
		Sep 29, 2011 8:46:00 AM
	Updated at:
		Sep 29, 2011 8:46:00 AM
```

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
```diff
@@ -89,7 +89,7 @@ public DescribeAction(DescribeActionConfig describeActionConfig, KeywhizClient c
           sanitizedSecret = keywhizClient.getSanitizedSecretByName(name);
 
           printing.printSanitizedSecretWithDetails(sanitizedSecret,
-              Arrays.asList("groups", "clients", "metadata"));
+              Arrays.asList("groups", "clients", "secret_data"));
```
mcpherrinm: A quick inspection looks like we only ever set the same options ... Maybe we can just get rid of the Options thing?

cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -112,16 +112,40 @@ public void printSanitizedSecretWithDetails(SanitizedSecret secret, List<String>
           .forEach(c -> System.out.println(INDENT + c.getName()));
     }
 
-    if (options.contains("metadata")) {
+    if (options.contains("secret_data")) {
       System.out.println("\tMetadata:");
       if(!secret.metadata().isEmpty()) {
         System.out.println(INDENT + secret.metadata().toString());
```
mcpherrinm: While you're in here, can we change secret.metadata() to be JSON instead of toString?

I think something like this will work:

      if(!secret.metadata().isEmpty()) {
        String metadata;
        try {
          metadata = new ObjectMapper().writeValueAsString(secret.metadata());
        } catch (JsonProcessingException e) {
          throw Throwables.propagate(e);
        }
        System.out.println(INDENT + metadata);
      }

Doesn't have to be in this PR

csstaub:
Not to be a downer, but this will almost certainly break a bunch of tools that screen-scrape the output. I'm not sure the improvement is worth it enough to break those folks.

mcpherrinm:
This is a human tool, and shouldn't be scraped.  I know there's an instance or two of it, but we can fix them.

I find myself running SQL queries or using `curl` to get this data instead, and it's a huge pain.

cli/src/main/java/keywhiz/cli/Printing.java
jbpeirce: New output: 
```
NonexistentOwner_Pass
	Groups:
		Security
		Web
	Clients:
		CN=User1
		CN=User2
		CN=User3
		CN=User4
		client
	Metadata:
		{"owner":"NonExistant","mode":"0400"}
	Created at:
		Sep 29, 2011 8:46:00 AM
	Updated at:
		Sep 29, 2011 8:46:00 AM
```

PR #269, opened 2016-09-22
Add metadata to groups.

jbpeirce:


api/src/main/java/keywhiz/api/CreateGroupRequest.java
```diff
@@ -30,22 +30,29 @@
   @JsonProperty
   public final String description;
 
+  @Nullable
+  @JsonProperty
+  public final String app;
+
   public CreateGroupRequest(@JsonProperty("name") String name,
```
jbpeirce: This is the change that most concerns me--it's a change to the constructor, and I wasn't able to figure out how to have two constructors one of which had only name and description (Json threw errors).  Is it worth forcing CreateGroupRequests to have an empty app name/adding a "setAppName" method separate from the constructor to avoid this?

mcpherrinm:
I think we should allow arbitrary JSON metadata, like we do with secrets

api/src/main/java/keywhiz/api/CreateGroupRequest.java
csstaub: Yes, just have one constructor and other use cases will have to pass `null` as the app name to indicate absence.

mcpherrinm:
We should print this metadata in `keywhiz.cli describe group` (like we print metadata for secrets)

mcpherrinm:
This change looks good but I'd like to go over it carefully, since it's a decently big change.

jbpeirce:
New group output from keywhiz-cli:
```
./cli/target/keywhiz-cli*shaded.jar --url https://localhost:4444 --devTrustStore --user keywhizAdmin describe group --name Web
Web
	Clients:
		CN=User1
		CN=User3
		CN=User4
	Secrets:
		Database_Password
		General_Password
		Nobody_PgPass
		NonexistentOwner_Pass
	Metadata:
	Created at:
		Jun 21, 2012 7:38:09 AM
	Updated at:
		Jun 21, 2012 7:38:09 AM
```

PR #270, opened 2016-09-26
Adds a lastSeen field to clients, and updates it when client is seen

stfinney:
Client lastSeen is updated in ClientAuthFactory, which produces the client object for every API endpoint.


csstaub:
Looks like build failure?

mcpherrinm:
Looks like this only adds migrations for Mysql.  We should support H2 and Postgres as well.

Also, we should print the value in keywhiz.cli

mcpherrinm:
Does this run once per TLS connection, or only every request?

We might need to be careful we don't swamp the database server with writes -- kwfs can be very chatty.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -73,6 +74,18 @@ public void deleteClient(Client client) {
     });
   }
 
+  public void sawClient(Client client) {
+    long now = OffsetDateTime.now().toEpochSecond();
+
+    dslContext.transaction(configuration -> {
+      DSL.using(configuration)
+          .update(CLIENTS)
+          .set(CLIENTS.LASTSEEN, now)
```
csstaub: I would do this instead:
```
.set(CLIENTS.LASTSEEN, DSL.greatest(CLIENTS.LASTSEEN, DSL.val(now, CLIENTS.LASTSEEN)))
```

Which is jOOQ DSL for:
```
lastseen = greatest(lastseen, :now)
```

That way you avoid every accidentally writing an earlier value than what is already in the database in case of clock skew.

server/src/main/java/keywhiz/service/daos/ClientMapper.java
```diff
@@ -29,6 +30,14 @@
...
 class ClientMapper implements RecordMapper<ClientsRecord, Client> {
   public Client map(ClientsRecord r) {
+    Long lastSeen = r.getLastseen();
+    ApiDate dateLastSeen;
+    if (lastSeen == null) {
+      dateLastSeen = null;
+    } else {
+      dateLastSeen = new ApiDate(lastSeen);
+    }
+
```
csstaub: Could also do:
```
ApiDate lastSeen = Optional.ofNullable(r.getLastseen()).map(ApiDate::new).orElse(null);
```

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -73,6 +74,18 @@ public void deleteClient(Client client) {
     });
   }
 
+  public void sawClient(Client client) {
+    long now = OffsetDateTime.now().toEpochSecond();
```
csstaub: Maybe add a check here: if `now` is within `threshold` seconds of `client.getLastseen()`, skip writing to the database? I think 24h-granularity for last seen is sufficient so threshold could be 24*60*60 which would avoid a lot of writes.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -76,14 +80,19 @@ public void deleteClient(Client client) {
...
-    });
+    ApiDate lastSeen = client.getLastSeen();
+
+    // only update last seen if it's been more than `lastSeenThreshold` seconds
+    // this way we can have less granularity on lastSeen and save db writes
+    if (lastSeen == null || now > lastSeen.toEpochSecond() + lastSeenThreshold) {
+      dslContext.transaction(configuration -> {
+        DSL.using(configuration)
+            .update(CLIENTS)
+            .set(CLIENTS.LASTSEEN, DSL.when(CLIENTS.LASTSEEN.isNull(), now).otherwise(DSL.greatest(CLIENTS.LASTSEEN, DSL.val(now, CLIENTS.LASTSEEN))))
```
csstaub: Neat

csstaub:
Link to build failure: https://travis-ci.org/square/keywhiz/builds/163851226#L8181

csstaub:
:( 
```
org.jooq.exception.DataAccessException: SQL [update CLIENTS set CLIENTS.LASTSEEN = case when CLIENTS.LASTSEEN is null then cast(? as bigint) else greatest(CLIENTS.LASTSEEN, cast(? as timestamp)) end where CLIENTS.ID = cast(? as int)]; Cannot parse "TIMESTAMP" constant "1475185281"; SQL statement:
update CLIENTS set CLIENTS.LASTSEEN = case when CLIENTS.LASTSEEN is null then cast(? as bigint) else greatest(CLIENTS.LASTSEEN, cast(? as timestamp)) end where CLIENTS.ID = cast(? as int) [22007-187]
	at org.h2.util.DateTimeUtils.parseDateValue(DateTimeUtils.java:275)
```

stfinney:
I guess I need to install postgres/h2 on my machine so I can test this stuff faster. Pushing and waiting on travis builds is kind of annoying.

PR #277, opened 2016-10-06
Add basic audit log

jbpeirce:
Defines an AuditLog interface and creates a simple implementation.  Calls AuditLog from all operations which affect the state of the Keywhiz database.

### DIFF
--- a/log/src/main/java/keywhiz/log/Event.java
+++ b/log/src/main/java/keywhiz/log/Event.java
@@ -1,0 +1,72 @@
+package keywhiz.log;
+
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A class to represent an event in Keywhiz
+ */
+public class Event {
+  private final Instant timestamp;
+  private final EventTag type;
+  private final String user;
+  private final String objectName; // The name of the affected object
+  private final Map<String, String> extraInfo; // Any extra information
+
+  private final DateTimeFormatter df =
+      DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss").withZone(ZoneId.of("UTC"));
+
+  public Event(Instant timestamp, EventTag type, String user, String objectName,
+      Map<String, String> extraInfo) {
+    this.timestamp = timestamp;
+    this.type = type;
+    this.user = user;
+    this.objectName = objectName;
+    this.extraInfo = extraInfo;
+  }
+
+  public Event(Instant timestamp, EventTag type, String user, String objectName) {
+    this.timestamp = timestamp;
+    this.type = type;
+    this.user = user;
+    this.objectName = objectName;
+    this.extraInfo = new HashMap<>();
+  }
+
+  public Instant getTimestamp() {
+    return timestamp;
+  }
+
+  public EventTag getType() {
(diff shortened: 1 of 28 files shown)


server/src/main/java/keywhiz/ServiceModule.java
```diff
@@ -75,6 +77,12 @@ public ServiceModule(KeywhizConfig config, Environment environment) {
     bind(KeywhizConfig.class).toInstance(config);
   }
 
+  // AuditLog
```
jbpeirce: I haven't worked much with injection before; if this should be added somewhere else, let me know.

log/src/main/java/keywhiz/log/Event.java
```diff
@@ -0,0 +1,68 @@
...
+  private final EventTag type;
+  private final String user;
+  private final String objectName; // The name of the affected object
+  private final Map<String, String> extraInfo; // Any extra information
+
+  private final DateTimeFormatter df =
+      DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss").withZone(ZoneId.of("UTC"));
+
+  public Event(Instant timestamp, EventTag type, String user, String objectName,
+      @Nullable Map<String, String> extraInfo) {
```
csstaub: You can have two constructors, one with the `extraInfo` argument and one without.

log/src/main/java/keywhiz/log/Event.java
```diff
@@ -0,0 +1,68 @@
...
+    return new StringBuilder().append(type)
+        .append(" Affected object: ")
+        .append(objectName)
+        .append(" Timestamp: ")
+        .append(df.format(timestamp))
+        .append(" User: ")
+        .append(user)
+        .append(" Additional information: ")
+        .append(extraInfo)
+        .toString();
```
csstaub: I would maybe put quotes around the fields being logged, e.g. "object name" in case the object name or such contains any weird characters or spaces or whatnot.

server/src/main/java/keywhiz/ServiceModule.java
csstaub: This looks like the right place.

pom.xml
```diff
@@ -28,6 +28,7 @@
     <module>model</module>
     <module>server</module>
     <module>testing</module>
+    <module>log</module>
```
mcpherrinm: I'm not sure this needs to be a separate module -- it could just be part of the server/ module.  It doesn't matter a ton, but it's one more POM to worry about.

server/src/main/java/keywhiz/service/resources/automation/AutomationGroupResource.java
```diff
@@ -158,21 +167,31 @@ public Group createGroup(
...
-      logger.info("Automation ({}) - Group {} already exists", automationClient.getName(), groupRequest.name);
+      logger.info("Automation ({}) - Group {} already exists", automationClient.getName(),
+          groupRequest.name);
       throw new ConflictException("Group name already exists.");
     }
 
     long id = groupDAO.createGroup(groupRequest.name, automationClient.getName(),
         nullToEmpty(groupRequest.description), groupRequest.metadata);
+    Map<String, String> extraInfo = new HashMap<>();
+    extraInfo.put("deprecated", "true");
```
mcpherrinm: nice!

PR #282, opened 2016-10-24
Add endpoint to list secrets in batches

jbpeirce:
This adds an endpoint to both the automation and admin APIs to list secrets in batches, specified by an index into a list of the secrets sorted by creation date from oldest to newest, and a number of secrets after that index to retrieve.  

The fact that an index of 0 retrieves the oldest secret is not necessarily the most convenient for a user querying secrets.  However, if the automation endpoint is used to iterate over secrets in batches, querying from newest to oldest would run into unexpected repeated secrets, and overlooked new secrets, if a secret was created between one batch and the next.  Querying from oldest to newest avoids this potential bug.

### DIFF
--- a/cli/src/main/java/keywhiz/cli/commands/ListAction.java
+++ b/cli/src/main/java/keywhiz/cli/commands/ListAction.java
@@ -24,8 +24,9 @@ import keywhiz.cli.Printing;
 import keywhiz.cli.configs.ListActionConfig;
 import keywhiz.client.KeywhizClient;
 
-public class ListAction implements Runnable {
+import static com.google.common.base.Preconditions.checkArgument;
 
+public class ListAction implements Runnable {
   private final ListActionConfig listActionConfig;
   private final KeywhizClient keywhizClient;
   private final Printing printing;
@@ -37,7 +38,7 @@ public class ListAction implements Runnable {
   }
 
   @Override public void run() {
-    List<String> listOptions = listActionConfig.listOptions;
+    List<String> listOptions = listActionConfig.listType;
     if (listOptions == null) {
       try {
         printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
@@ -61,7 +62,19 @@ public class ListAction implements Runnable {
           break;
 
         case "secrets":
-          printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          if (listActionConfig.idx == null && listActionConfig.num == null) {
+            printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          } else if (listActionConfig.idx != null && listActionConfig.num != null) {
+            checkArgument(listActionConfig.idx >= 0);
+            checkArgument(listActionConfig.num >= 0);
+            if (listActionConfig.newestFirst == null) {
+              printing.printAllSanitizedSecrets(keywhizClient.allSecretsBatched(listActionConfig.idx, listActionConfig.num, true));
+            }  else {
+              printing.printAllSanitizedSecrets(keywhizClient.allSecretsBatched(listActionConfig.idx, listActionConfig.num, listActionConfig.newestFirst));
+            }
+          } else {
+            throw new AssertionError("Both idx and num must be provided for batched secret queries");
+          }
           break;
 
         default:
--- a/cli/src/main/java/keywhiz/cli/configs/ListActionConfig.java
+++ b/cli/src/main/java/keywhiz/cli/configs/ListActionConfig.java
@@ -24,5 +24,14 @@ import java.util.List;
 public class ListActionConfig {
 
   @Parameter(description = "[<groups|clients|secrets>]")
-  public List<String> listOptions;
+  public List<String> listType;
+
+  @Parameter(names = "--idx", description = "Index to start retrieving secrets (valid only with 'secrets'; requires --num to also be specified)")
+  public Integer idx;
+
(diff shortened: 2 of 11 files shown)


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -182,6 +184,21 @@ public int setCurrentVersion(long secretId, long secretContentId) {
     return ImmutableList.copyOf(r);
   }
 
+  public ImmutableList<SecretSeries> getSecretSeriesBatched(int idx, int num) {
```
jbpeirce: Good point (I'm not certain how MySQL handles negative values, but it should be avoided)

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -524,6 +524,61 @@ private void checkSecretVersions(List<SecretDetailResponseV2> versions, String n
...
+        .groups("group16b")
+        .build());
+
+    // check limiting by batch (hard to test because the batch results heavily depend on other
+    // tests, which may be run in parallel)
+    List<String> s1 = listBatch(0, 2);
+    assertThat(s1.size()).isEqualTo(2);
+  }
+
+  @Test public void secretListingBatch_failure() throws Exception {
```
stfinney: 👍 for the often forgotten step of testing bad inputs.

stfinney:
LGTM

cli/src/main/java/keywhiz/cli/commands/ListAction.java
```diff
@@ -61,7 +60,16 @@ public ListAction(ListActionConfig listActionConfig, KeywhizClient client, Print
           break;
 
         case "secrets":
-          printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          if (listActionConfig.idx == null && listActionConfig.num == null) {
+            printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          } else if (listActionConfig.idx != null && listActionConfig.num != null) {
+            if (listActionConfig.idx < 0 || listActionConfig.num < 0) {
```
alokmenghrajani: `checkArgument(listActionConfig.idx >= 0)` is more succinct/readable.

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -141,6 +141,29 @@ protected SanitizedSecret retrieveSecret(@Auth User user, String name) {
...
+   *
+   * @excludeParams user
+   * @param idx the first index to select in a list of secrets sorted by creation time
+   * @param num the number of secrets after idx to select in the list of secrets
+   *
+   * @description Returns a list of secret names for this user, limited by the input idx and num
+   * Used by Keywhiz CLI and the web ui.
+   * @responseMessage 200 Found and retrieved Secret(s)
+   */
+  @Path("batch/{idx}-{num}")
```
alokmenghrajani: You could have kept the current API to list all secrets and added an optional idx/num (query parameter).

I feel idx and num are poor choices for path parameters.

stfinney:
Given the change to query params instead of path for idx/num, is it worth changing (in a separate PR) for the other endpoint that has a similar functionality?

`GET
  @Path("{name}/versions/{versionIdx}-{numVersions}")
  @Produces(APPLICATION_JSON)
  public Iterable<SecretDetailResponseV2> secretVersions(@Auth AutomationClient automationClient,
      @PathParam("name") String name, @PathParam("versionIdx") int versionIdx,
      @PathParam("numVersions") int numVersions) {`

This is still new so presumably the only thing making use of it is SA? So maybe not that hard to change, for the sake of a more consistent API.

cli/src/main/java/keywhiz/cli/commands/ListAction.java
```diff
@@ -61,7 +62,19 @@ public ListAction(ListActionConfig listActionConfig, KeywhizClient client, Print
           break;
 
         case "secrets":
-          printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          if (listActionConfig.idx == null && listActionConfig.num == null) {
+            printing.printAllSanitizedSecrets(keywhizClient.allSecrets());
+          } else if (listActionConfig.idx != null && listActionConfig.num != null) {
+            checkArgument(listActionConfig.idx >= 0);
+            checkArgument(listActionConfig.num >= 0);
```
alokmenghrajani: `listActionConfig.num > 0`

cli/src/main/java/keywhiz/cli/configs/ListActionConfig.java
```diff
@@ -24,5 +24,14 @@
...
-  public List<String> listOptions;
+  public List<String> listType;
+
+  @Parameter(names = "--idx", description = "Index to start retrieving secrets (valid only with 'secrets'; requires --num to also be specified)")
+  public Integer idx;
+
+  @Parameter(names = "--num", description = "Number of secrets to retrieve after index (valid only with 'secrets'; requires --idx to also be specified)")
+  public Integer num;
+
+  @Parameter(names = "--newestFirst", description = "Whether to batch the secrets from newest creation date first.  Defaults to 'true' (valid only with 'secrets'; requires --idx and --num to also be specified)")
```
alokmenghrajani: nit: you could call this parameter `order` and then add various options down the road if needed (e.g. newest, oldest, byId, bySomethingElse, etc.).

PR #283, opened 2016-10-25
Add HMAC column to secrets_content table

jbpeirce:


mcpherrinm:
I think we should probably use an appropriately sized BINARY, not `text`.

alokmenghrajani:
Storing hex-encoded strings makes other things slightly easier.

csstaub:
Why TEXT, why not CHAR (or VARCHAR) with fixed length? Consider http://mysqlserverteam.com/externally-stored-fields-in-innodb/

alokmenghrajani:
VARCHAR(64) would work too.

PR #284, opened 2016-10-25
Jessep/use hmac column in secrets

jbpeirce:


api/src/main/java/keywhiz/api/model/Secret.java
```diff
@@ -115,6 +118,8 @@ public String getSecret() {
     return secret;
   }
 
+  public String getHmac() { return hmac; }
```
csstaub: Nit: formatting

api/src/main/java/keywhiz/api/model/Secret.java
```diff
@@ -47,6 +47,7 @@
   /** Base64-encoded content of this version of the secret. */
   private String secret;
   private final LazyString encryptedSecret;
+  private final String hmac;
```
csstaub: Call this "checksum"? Fact that it's HMAC doesn't really matter to client

api/src/main/java/keywhiz/api/model/SecretContent.java
```diff
@@ -32,17 +32,18 @@
  */
 @AutoValue
 public abstract class SecretContent {
-  public static SecretContent of(long id, long secretSeriesId, String encryptedContent, ApiDate createdAt,
+  public static SecretContent of(long id, long secretSeriesId, String encryptedContent, String hmac, ApiDate createdAt,
```
csstaub: Should `hmac` be `@Nullable` here? It won't always be set, will it? (at least until a backfill is completed)

server/src/main/java/keywhiz/service/crypto/ContentCryptographer.java
```diff
@@ -149,6 +151,18 @@ public String decrypt(String ciphertextJson) {
...
 
+  public String computeHmac(byte[] data) {
+    SecretKey hmacKey = deriveKey(32, "hmackey");
+    try {
+      Mac mac = Mac.getInstance("HmacSHA256");
+      mac.init(hmacKey);
+      return BaseEncoding.base16().encode(mac.doFinal(data));
+    } catch (NoSuchAlgorithmException | InvalidKeyException e) {
+      logger.warn("Error computing HMAC: ", e);
+      return "";
```
csstaub: Shouldn't this be `null`?

api/src/main/java/keywhiz/api/model/SecretContent.java
jbpeirce: We set the migration to have a default of empty string, so hmac should always be a string (though it may be empty)

api/src/main/java/keywhiz/api/model/Secret.java
jbpeirce: I'll update the PR to do this (forgot IntelliJ's "refactor" option would make this trivial).

PR #292, opened 2016-12-01
Return current secret version's metadata and hmac in secret info

jbpeirce:
The automation secret-info endpoint now returns the metadata and content HMAC of the current version of the input secret, since the metadata often includes useful information.  (The content is omitted, since it is seldom needed and should not be exposed unnecessarily)


api/src/test/java/keywhiz/api/automation/v2/SecretDetailResponseV2Test.java
```diff
@@ -66,6 +67,26 @@
...
+    SecretDetailResponseV2 secretDetailResponse = SecretDetailResponseV2.builder()
+        .seriesAndContent(seriesAndContent)
+        .content("YXNkZGFz")
+        .build();
+
+    assertThat(asJson(secretDetailResponse))
+        .isEqualTo(jsonFixture("fixtures/v2/secretDetailResponse.json"));
+  }
+
+
```
stfinney: nit: extra blank line

PR #293, opened 2016-12-02
Add selective update endpoint

jbpeirce:
This adds an endpoint to Keywhiz which updates an existing secret, keeping all fields which are not explicitly set the same as they are in the current version of the secret.


jbpeirce:
Currently in keywhiz-cli, to update a secret without providing content you have to hit Crtl+D or otherwise trigger EOF after running the command.  Alternatively, I could set up a flag to indicate that content will be piped in (with the option, still, to not input any content).  Any opinions?

Issue #294, opened 2016-12-08
HKDF makes 3 calls to HSM unnecessarily


stfinney:
HKDF performs 3 calls to mac update, which is 3 round trips to the HSM. We can reduce this to 1 to get better performance.

PR #295, opened 2016-12-09
Remove calls to Mac.update()

alokmenghrajani:
When using HSMs, it's more efficient to perform a single call to doFinal(). If the HSM is configured with Sun's PKCS11 provider, we might save 3 calls. When using native Java Crypto Providers, we don't save as much (on the order of ~10%).

Fixes https://github.com/square/keywhiz/issues/294


hkdf/src/main/java/keywhiz/hkdf/Hkdf.java
```diff
@@ -148,10 +149,11 @@ public SecretKey extract(@Nullable SecretKey salt, byte[] ikm) {
     Mac mac = initMac(key);
     for (int roundNum = 1; roundNum <= n; roundNum++) {
       mac.reset();
-      mac.update(hashRound);
-      mac.update(info);
-      mac.update((byte) roundNum);
-      hashRound = mac.doFinal();
+      ByteBuffer t = ByteBuffer.allocate(hashRound.length + info.length + 1);
```
alokmenghrajani: note: it might look inefficient to perform allocations inside a loop, but we only ever loop once.

Issue #301, opened 2017-01-05
Maven build for H2 is failing for CLI


sureshg:
Steps followed

*  Cloned the repo `https://github.com/square/keywhiz`
*  Installed JCE `Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files`
*  Ran `mvn install -P h2`

```bash
~/code/keywhiz (master)
$ mvn -version
Apache Maven 3.2.1 (ea8b2b07643dbb1b84b6d16e1f08391b666bc1e9; 2014-02-14T09:37:52-08:00)
Maven home: /Users/xxxx/install/apache-maven-3.2.1
Java version: 1.8.0_111, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.12", arch: "x86_64", family: "mac"

$ java -version
java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)
```

Error executing maven shade plugin

```ruby
[INFO]
[INFO] --- maven-jar-plugin:2.4:jar (default-jar) @ keywhiz-cli ---
[INFO] Building jar: /Users/sgopal1/code/keywhiz/cli/target/keywhiz-cli-0.7.11-SNAPSHOT.jar
[INFO]
[INFO] --- maven-dependency-plugin:2.8:copy-dependencies (copy) @ keywhiz-cli ---
[INFO]
[INFO] --- maven-source-plugin:2.4:jar-no-fork (attach-sources) @ keywhiz-cli ---
[INFO] Building jar: /Users/sgopal1/code/keywhiz/cli/target/keywhiz-cli-0.7.11-SNAPSHOT-sources.jar
[INFO]
[INFO] --- maven-shade-plugin:1.7.1:shade (default) @ keywhiz-cli ---
[WARNING] Error injecting: org.apache.maven.plugins.shade.mojo.ShadeMojo
java.lang.NoClassDefFoundError: org/codehaus/plexus/util/xml/XmlStreamWriter
	at java.lang.Class.getDeclaredConstructors0(Native Method)
	at java.lang.Class.privateGetDeclaredConstructors(Class.java:2671)
	at java.lang.Class.getDeclaredConstructors(Class.java:2020)
[...]

PR #302, opened 2017-01-06
Use maven wrapper for fully encapsulated build setup.

sureshg:
The following changes have included in this PR.

* Use maven wrapper (3.3.9) for fully encapsulated build setup. With this changes, if a user doesn't have the necessary version of Maven installed, it will be downloaded for the user first.

* Updated to the latest version of `maven-shade-plugin` plugin (`2.4.3`). This version fixes issue #301 and [JIRA MSHADE-183](https://issues.apache.org/jira/browse/MSHADE-183) (which is mentioned in [pom.xml](https://github.com/square/keywhiz/blob/master/pom.xml#L436))


mcpherrinm:
This seems reasonable.

Is this https://github.com/takari/maven-wrapper?  I'll take a look at this project.

It seems a bit weird to me to be checking in a .jar, though.

Since I'm unfamiliar with this wrapper, as is the rest of the team, I'll want to take a little time to use it out.  Unfortunately it's already getting late on Friday, so it'll have to wait until next week.

sureshg:
@mcpherrinm Yes, it's the takari maven wrapper.

> It seems a bit weird to me to be checking in a .jar, though.

This is very common and is the way to use gradle (Eg: https://github.com/spring-projects/spring-framework/tree/master/gradle/wrapper)

For more info, check this - https://docs.gradle.org/current/userguide/gradle_wrapper.html

sureshg:
Any update on this?

mcpherrinm:
I'm concerned that my team wouldn't use this tool, because it's not part of our usual workflow.

In my experience, anything like this we accept but don't use ends up broken, and does more harm than good.

PR #305, opened 2017-01-19
Fix expiry v3 endpoint to prevent 500 errors

worldwise001:


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -358,7 +359,8 @@ public void findAndEvictClient(long clientId, long groupId, AuditLog auditLog, S
         .join(ACCESSGRANTS).on(ACCESSGRANTS.GROUPID.eq(GROUPS.ID))
         .join(SECRETS).on(ACCESSGRANTS.SECRETID.eq(SECRETS.ID))
         .where(SECRETS.ID.in(secretIdList))
-        .fetchInto(GROUPS).map(groupMapper);
+        .groupBy(GROUPS.ID)
+        .fetchInto(GROUPS).map(groupMapper).stream().collect(toList());
```
mcpherrinm: It seems silly to collect into a list, only to use it to stream() right below again.  Can we combine those two collects?

worldwise001: Yeah, I was concerned it might be unreadable, but I can do that.

worldwise001: Aha, the List was a vestige of some other weird way I was requesting this information. I've been able to combine the two. Updated commit incoming.

PR #311, opened 2017-02-16
Add keywhiz cli rollback

jbpeirce:
This PR adds the ability to list versions of a secret and to reset a secret to an older version through keywhiz-cli.  It also updates some of the secret models to contain more information (so that the version listing can display more information) and logs the current version when a secret is deleted, so that one can manually un-delete a secret by modifying the database and restore the correct version.

screen shot 2017-02-15 at 6 02 48 pm


jbpeirce:
The current command for rolling back is 
```
keywhiz.cli rollback --name Hacking_Password_3 --version 964
```
I could rename it to `reset-version` or something similar, which might be clearer since the version ID we're rolling back to may be higher than the current version ID--any opinions?

mcpherrinm:
I'm fine with rollback

PR #313, opened 2017-02-28
Do not overwrite existing expiry, but log if different

csstaub:
Changes:
* Get existing expiry if set for secret
* Run expiry extractor for the given secret
* Compare and and warn if different

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
+++ b/server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
@@ -64,6 +64,7 @@ import org.slf4j.LoggerFactory;
 
 import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.time.temporal.ChronoUnit.HOURS;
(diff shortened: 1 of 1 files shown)


server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -378,6 +384,18 @@ public boolean backfillExpiration(@Auth AutomationClient automationClient, @Path
     }
 
     if (expiry != null) {
+      if (existingExpiry.isPresent()) {
+        long offset = existingExpiry.get().until(expiry, HOURS);
+        if (offset > 24 || offset < -24) {
+          logger.warn(
+              "Extracted expiration of secret {} differs from actual by more than {} hours.",
```
jbpeirce: Maybe log the extracted expiration, so that we have the option to manually reset the secret's expiration to match the certificate's expiration?

csstaub: Yes, that's a good idea. I'll ad that

PR #314, opened 2017-03-09
Use read-only data source where possible

csstaub:
This pull request changes various resources to make sure we use the read-only data source when possible. Most of the time we don't need to talk to our primary, read-write data source. Had to upgrade jOOQ so we can disable save points on nested transactions (see https://groups.google.com/forum/#!topic/jooq-user/zG0U6CkxI5o).

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/automation/AutomationSecretResource.java
+++ b/server/src/main/java/keywhiz/service/resources/automation/AutomationSecretResource.java
@@ -76,22 +76,24 @@ public class AutomationSecretResource {
   private static final Logger logger = LoggerFactory.getLogger(AutomationSecretResource.class);
   private final SecretController secretController;
   private final SecretDAO secretDAO;
-  private final AclDAO aclDAO;
+  private final AclDAO aclDAOReadOnly;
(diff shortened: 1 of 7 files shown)


jbpeirce:
It looks like the clientDAO in SecretDeliveryResource could also be switched to readOnly (and some of the deprecated automation resources could have more read-only connections added, although since they're deprecated that's not necessarily worth it).  These changes look good!

server/src/main/java/keywhiz/service/resources/automation/AutomationSecretResource.java
```diff
@@ -128,7 +130,7 @@ public AutomationSecretResponse createSecret(
       throw new ConflictException(format("Cannot create secret %s.", request.name));
     }
     ImmutableList<Group> groups =
-        ImmutableList.copyOf(aclDAO.getGroupsFor(secret));
+        ImmutableList.copyOf(aclDAOReadOnly.getGroupsFor(secret));
```
csstaub: Yeah, it might see outdated groups. I can update this to use the R/W aclDAO instead.

PR #316, opened 2017-03-16
Remove the UI

mcpherrinm:


mcpherrinm:
It looks like this also might speed up travis-ci a bunch, since we're not installing Chrome and stuff for UI testing.  Eyeballing it, looks like saving about 2 minutes on the build (7->5 mins)

PR #319, opened 2017-03-18
Warn in keywhiz-cli when data is piped into update without the --content flag

jbpeirce:
This makes the update command print a warning if content appears to have been piped into it (stdin could be read from without blocking) but the `--content` flag was not specified.  It does not update the secret with the input content if the flag was not specified, however.

### DIFF
--- a/cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
+++ b/cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
@@ -21,6 +21,8 @@ import com.google.common.base.Throwables;
 import com.google.common.io.ByteStreams;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.UnsupportedEncodingException;
 import keywhiz.cli.configs.AddOrUpdateActionConfig;
 import keywhiz.cli.configs.UpdateActionConfig;
 import keywhiz.client.KeywhizClient;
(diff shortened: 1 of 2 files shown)


mcpherrinm:
We should make sure we never prompt for a login password on `keywhiz.cli update` if we're doing this  (Are we already? not sure)

jbpeirce:
This shouldn't be a problem with `keywhiz.cli update` on a login prompt--the new warning triggers only when output is piped into update, and if `keywhiz.cli` needs to login in that situation it throws `Exception in thread "main" java.lang.RuntimeException: Please login by running a command without piping.`  

I'm not sure I understand what the bug with the login password would have been--is it that the reader might unintentionally capture the password (but then discard it), or that the warning might be triggered unnecessarily by the password?

mcpherrinm:
I forgot we didn't allow piped-in passwords:  I was afraid of somebody doing something like `keywhiz.cli update whatever < password`, intending the password to be used to log in.

But you're right, that doesn't matter.

cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
```diff
@@ -59,6 +61,19 @@ public UpdateAction(UpdateActionConfig config, KeywhizClient client,
...
+    // If it appears that content was piped in but --content was not specified, print a warning
+    if (!config.contentProvided) {
+      try {
+        InputStreamReader reader = new InputStreamReader(stream, "UTF-8");
+        if (reader.ready()) {
+          System.out.println("\nWarning: Specify the --content flag to update a secret's content.");
+          System.out.println("The secret has not been updated with any provided content.");
+        }
+      } catch (IOException e) {
+        throw new AssertionError(String.format("Unexpected error trying to create an InputStreamReader for stdin: %s", e.getMessage()));
```
mcpherrinm: Maybe we should just log & swallow this?  We're only constructing the InputStreamReader for a warning

jbpeirce: 👍

PR #322, opened 2017-03-29
Remove SecretVersion

jbpeirce:
SecretVersion was virtually identical to SanitizedSecret (the only difference is that SanitizedSecret includes a "generationOptions" map), and used only internally.  This PR removes it and replaces its uses with SanitizedSecret.


api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
```diff
@@ -137,7 +139,7 @@ public SecretDetailResponseV2 build() {
       @JsonProperty("updatedBy") String updatedBy,
       @JsonProperty("type") @Nullable String type,
       @JsonProperty("metadata") @Nullable Map<String, String> metadata,
-      @JsonProperty("expiry") @Nullable long expiry) {
+      @JsonProperty("expiry") long expiry) {
```
jbpeirce: The database is defined with "expiry" as NOT NULL DEFAULT 0, and none of the objects from which SecretDetailResponseV2 is constructed have nullable expiry, so I removed the @Nullable annotation here (which wasn't having any effect because primitives cannot be nullable)

PR #324, opened 2017-03-29
Remove content field from SecretDetailResponseV2

jbpeirce:
The content field in SecretDetailResponseV2 was never set in Keywhiz, making its presence confusing.

SecretDetailResponseV2 is returned from automation endpoints, which return information on any existing secrets to any automation client.  Returning unencrypted secret contents in this situation leaks information unnecessarily (note that automation clients can assign secrets to themselves and view the contents using the SecretDeliveryResource, leaving an audit trail in the process).  

This PR also removes a pair of constructors for SecretDetailResponseV2 (building it from a Secret and from a SecretSeries) which were unused except in tests.  They can be restored easily if necessary.


mcpherrinm:
This looks fine, but I'd like to make sure our clients are updated first, as this is a breaking change.

PR #335, opened 2017-05-04
Perform client authentication against r/o database

csstaub:
Perform client auth against the local read-only database, instead of connecting to the read/write master.

### DIFF
--- a/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
+++ b/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
@@ -50,11 +50,11 @@ public class ClientAuthFactory {
   private final Authenticator<String, Client> authenticator;
 
   @Inject public ClientAuthFactory(ClientDAOFactory clientDAOFactory) {
-    this.authenticator = new MyAuthenticator(clientDAOFactory.readwrite());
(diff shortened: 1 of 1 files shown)


server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -89,18 +89,22 @@ public Client provide(ContainerRequest request) {
...
     }
 
     @Override public Optional<Client> authenticate(String name)
         throws AuthenticationException {
-      Optional<Client> optionalClient = clientDAO.getClient(name);
+      Optional<Client> optionalClient = clientDAOReadOnly.getClient(name);
       if (optionalClient.isPresent()) {
         Client client = optionalClient.get();
-        clientDAO.sawClient(client);
+        clientDAOReadWrite.sawClient(client);
```
alokmenghrajani: We could do this write conditionally if the last seen was >1 week or something like that.

mcpherrinm: the sawClient function does that already

mcpherrinm: I'd like to make the writeback async though, and probably batched.

PR #341, opened 2017-05-19
Print the content HMAC when describing a secret through the cli

jbpeirce:
This may be useful for debugging, such as checking whether two secrets have the same contents or verifying that an update altered secret contents.


alokmenghrajani:
❤️

PR #342, opened 2017-05-31
Update docs a bit

csstaub:
Update docs a bit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,9 +36,9 @@
 #
 #   Be ready to provide a server certificate/private key for setup.
 #
-# After keywhiz starts up, you can access the admin console by going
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -36,9 +36,9 @@
 #
 #   Be ready to provide a server certificate/private key for setup.
 #
-# After keywhiz starts up, you can access the admin console by going
-# to https://[DOCKER-MACHINE-IP]/ui in your browser. The default user
-# in development is keywhizAdmin and the default password is adminPass.
+# After keywhiz starts up, you can access it via by running keywhiz-cli and
```
jbpeirce: nit: "via by" should be either "via" or "by"

csstaub: fixed

PR #351, opened 2017-06-21
Standardize the meaning of createdAt/updatedAt for secrets.

jbpeirce:
The createdAt/By and updatedAt/By for SanitizedSecret are now
standardized to reflect the creation and update information
for the "secret series"--the secret, including all of its versions--
rather than the latest version of the secret.  If the secret's
contents are rolled back to an older version, the SanitizedSecret's
creation and update data will not be altered.

### DIFF
--- a/api/src/main/java/keywhiz/api/model/SanitizedSecretWithGroups.java
+++ b/api/src/main/java/keywhiz/api/model/SanitizedSecretWithGroups.java
@@ -64,7 +64,6 @@ public abstract class SanitizedSecretWithGroups {
 
   /** @return Name to serialize for clients. */
   public static String displayName(SanitizedSecretWithGroups sanitizedSecretWithGroups) {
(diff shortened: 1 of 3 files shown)


jbpeirce:
I'm planning to add an explicit `contentCreatedAt` and `contentCreatedBy` field to `SanitizedSecret` and `Secret` (and, possibly, `SecretDetailResponseV2`), to deal with the case of content rollback explicitly.  Secret series' `updatedAt` timestamp is reset when their version is reset.  This means that if we create a secret in March, update it in April, and roll back to the version from March in May, the secret will have `createdAt` in March and `updatedAt` in May (and, eventually, `contentCreatedAt` in March).

stfinney:
re: your comment, that makes more sense to me than having the updated date go back to March, as long as we surface the update date for the version as well.

PR #352, opened 2017-06-21
Add fields for secret content creation data

jbpeirce:
The primary changes are to Secret, SanitizedSecret, SecretDetailResponseV2, and Printing in keywhiz.cli; most of the other alterations just update tests to add parameters. 

If secrets have had their contents reset to an earlier version, or when SanitizedSecret/SecretDetailResponseV2 is used to represent an older version of a secret, the content creation information will differ from the secret's creation and update data.  This adds a field to explicitly track content creation data (which is equivalent to content update data since Keywhiz never updates content records).

### DIFF
--- a/api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
+++ b/api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
@@ -7,6 +7,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import javax.annotation.Nullable;
+import keywhiz.api.ApiDate;
 import keywhiz.api.model.SanitizedSecret;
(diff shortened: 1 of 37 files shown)


api/src/main/java/keywhiz/api/automation/v2/SecretDetailResponseV2.java
```diff
@@ -38,6 +39,8 @@ public static Builder builder() {
     public abstract Builder type(@Nullable String type);
     public abstract Builder expiry(long expiry);
     public abstract Builder version(@Nullable Long version); // Unique ID in secrets_content table
+    public abstract Builder contentCreatedAtSeconds(@Nullable Long contentCreatedAt);
+    public abstract Builder contentCreatedBy(@Nullable String person);
```
csstaub: Nit: might not be a person, could also be an automation client.

Issue #291, opened 2016-11-30
Remove non-mysql database support


mcpherrinm:
Our build systems are overly complicated by our (not great) way of supporting multiple database engines.

I think we should remove postgres/h2, and check in the jooq generated Java.

That'll simplify building a lot.  If there's somebody who wants to support another database later, then we could consider re-adding databases.

There will be additional burden on any developers making new migrations, because they'll have to check in the generated code, but we can provide a script to ensure that's done properly.

mcpherrinm:
Postgres has been removed

mbyczkowski:
H2 has been removed in #444.

PR #353, opened 2017-06-21
remove PostgreSQL reference from documentation

tfarina:
Support for PostgreSQL was dropped by commit 87c752117f8e: ("Remove postgres support").

 GH-291


mcpherrinm:
Thanks!

PR #356, opened 2017-06-27
Check in javadocs so they're on the website

mcpherrinm:
You can see this in action at https://mcpherrinm.github.io/keywhiz/javadocs/ -- they'll work on https://square.github.io/keywhiz/javadocs/ once this is merged.

The index page is pretty primitive - we could add to it later if we wanted, but it gets the job done


csstaub:
Link it in README if you haven't already.

PR #358, opened 2017-06-30
Remove unnecessary dependency from keywhiz-log

jbpeirce:
This dependency is not in use any longer (it's unclear whether it was ever required)


docs/apidocs/automation_v2-secret-management.json
```diff
@@ -694,6 +694,13 @@
         "version" : {
           "type" : "integer",
           "format" : "int64"
+        },
+        "contentCreatedAtSeconds" : {
```
mcpherrinm: Is this supposed to be included in this change?

jbpeirce: It was generated as a side effect when I built Keywhiz to make sure the dependency removal didn't break anything; I could pull it into a separate PR or commit.

mcpherrinm: separate commit is fine

jbpeirce: Done.

PR #368, opened 2018-02-14
Downgrade logback version

jbpeirce:
Dropwizard before version 1.0.0 is incompatible with logback
after 1.1.3, because logback 1.1.5 and 1.1.6 added checks to
RollingFileAppender which cause errors in Dropwizard's
FileAppenderFactory (see Dropwizard issue 1478 on Github).  The
result is that Dropwizard before 1.0.0 will log to the console but
not to files if logback is after 1.1.3.  Keywhiz currently uses
Dropwizard 0.8.0, so logback must be kept to an early version
until it can be upgraded together with Dropwizard.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
     <slf4j.version>1.7.25</slf4j.version>
     <findbugs.version>3.0.1</findbugs.version>
     <mysql.version>5.1.45</mysql.version>
-    <logback.version>1.2.3</logback.version>
+    <logback.version>1.1.3</logback.version>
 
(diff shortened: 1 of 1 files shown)


jbpeirce:
1.1.3 was the previous version of Logback Keywhiz used (before [#367](https://github.com/square/keywhiz/pull/367/files#diff-600376dffeb79835ede4a0b285078036L48)), so Keywhiz should run normally with it and that keeps Keywhiz running on the latest version of Logback that its version of Dropwizard supports.

PR #369, opened 2018-02-15
Track expiration of client certificates in database

csstaub:
Track expiration of client certificates in database, alongside the last seen value we already track. Updates at most once per 24 hour period.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/ClientDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/ClientDAO.java
@@ -18,15 +18,17 @@ package keywhiz.service.daos;
 
 import com.google.common.collect.ImmutableSet;
 
+import java.security.Principal;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.OffsetDateTime;
-import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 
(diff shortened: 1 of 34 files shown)


server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -83,18 +87,37 @@ public void deleteClient(Client client) {
...
+
         DSL.using(configuration)
             .update(CLIENTS)
-            .set(CLIENTS.LASTSEEN, DSL.when(CLIENTS.LASTSEEN.isNull(), val).otherwise(DSL.greatest(CLIENTS.LASTSEEN, val)))
+            .set(CLIENTS.LASTSEEN,
+                when(CLIENTS.LASTSEEN.isNull(), lastSeenValue)
+                    .otherwise(greatest(CLIENTS.LASTSEEN, lastSeenValue)))
+            .set(CLIENTS.EXPIRATION,
+                when(CLIENTS.EXPIRATION.isNull(), expirationValue)
+                    .otherwise(greatest(CLIENTS.EXPIRATION, expirationValue)))
```
jbpeirce: This assumes no client will ever present a certificate which expires before a certificate it used previously.  It might be safer to always overwrite the recorded expiration with the just-seen expiration.

csstaub: Good point; updated.

PR #372, opened 2018-04-16
Update ManualStatusServlet.java

bit-twidd1er:
Escape the "path" returned in the response in the case of an error in order to prevent a reflected XSS attack

### DIFF
--- a/server/src/main/java/keywhiz/ManualStatusServlet.java
+++ b/server/src/main/java/keywhiz/ManualStatusServlet.java
@@ -22,7 +22,6 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import org.apache.commons.lang3.StringEscapeUtils;
 
 /** A Servlet added to the Admin Connector for manually making _status report unhealthy */
 public class ManualStatusServlet extends HttpServlet implements Serializable {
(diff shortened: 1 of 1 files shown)


csstaub:
Alternatively, let's just drop the path reflection? It looks like escapeHtml4 in StringEscapeUtils is deprecated, so probably best not to introduce new code for it. The path doesn't need to be in the return message anyway, I think.

bit-twidd1er:
Agreed. Simply removing the path from the response is a cleaner solution.

csstaub:
Awesome, thanks for the fix @bit-twidd1er!

bit-twidd1er:
Happy to help! Thanks for working with me and being so responsive

csstaub:
Related, we should probably just set a CSP for all status endpoints and not worry about this at all. But that's a bit of a larger effort.

alokmenghrajani:
Afaik there's no XSS here. The parameters comes from the path, which the browser escapes when sending the query.

Do you have a proof of concept?

bit-twidd1er:
While most modern browsers escape the path when typed into the URL, depending on the browser to stop your users from getting popped is very poor security posture. I always recommend against relying on mitigations instead of patching issues as some percentage of users will be running IE7 on Windows XP and not be protected by their browser. 

This is a case of reflecting back user-controlled data without any sanitization when there is no real reason for it. Including that path in the message contents introduces risk and doesn't provide any benefit. 

There are some tricks that can be used in order to get around URL encoding that is done by modern browsers as well. In this case malicious Javascript or MiTM could make/modify a request (so as not to have URL encoding in the request) and while the URL that the browser will display on the response will be URL encoded by the browser, the message contents will be based upon the request (without encoding) and still execute the XSS. As a note, I had POC'd this 2 days ago, but deleted my setup yesterday once this pull request was merged.

Issue #376, opened 2018-08-27
Broken on Java 9 and 10


mbyczkowski:
Keywhiz crashes when Java 9 or 10 is used.

Java 10:
```
export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-10.jdk/Contents/Home
java -cp ./server/target/keywhiz-server-0.8.1-SNAPSHOT-shaded.jar:./haas-provider.jar keywhiz.KeywhizService server ./server/src/main/resources/keywhiz-development.yaml
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/ValidationException
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1009)
        at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
        at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:801)
        at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:699)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:622)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:190)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:499)
        at org.hibernate.validator.internal.engine.ConfigurationImpl.<init>(ConfigurationImpl.java:110)
        at org.hibernate.validator.internal.engine.ConfigurationImpl.<init>(ConfigurationImpl.java:96)
        at org.hibernate.validator.HibernateValidator.createSpecializedConfiguration(HibernateValidator.java:37)
        at org.hibernate.validator.HibernateValidator.createSpecializedConfiguration(HibernateValidator.java:34)
        at javax.validation.Validation$ProviderSpecificBootstrapImpl.configure(Validation.java:220)
        at io.dropwizard.setup.Bootstrap.<init>(Bootstrap.java:71)
        at io.dropwizard.Application.run(Application.java:68)
        at keywhiz.KeywhizService.main(KeywhizService.java:72)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.ValidationException
[...]

PR #381, opened 2018-10-18
First pass at getting keywhiz to work in JDK11

worldwise001:
@mcpherrinm and I spent some time trying to see if we could upgrade keywhiz to use JDK11. All the maven components were upgraded. We added all the relevant jaxb dependencies that used to be in J2EE (but have since been removed).

Along the way we also upgraded jooq and also removed findbugs since it was causing some weird errors. We also removed Powermock since it seems like that might not be supported in JDK11 due to stronger access/reflection requirements in JDK 10+. (It looks like it's only used to mock out the LDAP-related tests, and I have a hunch I could probably extensively mock those or use spy instead). Looks like Powermock only supports JDKs past 9 with version 2, which appears to be in beta right now.

We also fixed OkHttp issues with the sslFactory creator since that now hard fails if you do not supply a trust store (probably related to certificate pining).

Some other annotations were cleaned up/removed.

Some javadoc components may still be flaky; I got maven to run successfully in IntelliJ but the package and install commands seem to not work sometimes.

But otherwise the package compiles, tests, and runs successfully in JDK11. So we think this is a good first stab at this and will unblock upgrading different dependencies and features.

### DIFF
--- a/docker/wizard.sh
+++ b/docker/wizard.sh
@@ -65,7 +65,7 @@ export KEYSTORE_PATH="/secrets/keywhiz-server.p12"
 export KEYSTORE_PASSWORD=`head -c16 /dev/urandom | xxd -p`
 export TRUSTSTORE_PATH="/secrets/ca-bundle.p12"
 export TRUSTSTORE_PASSWORD=ponies # contains public certificates only
-export CRL_FILE_PEM="/secrets/ca-crl.pem"
+export CRL_PATH="/secrets/ca-crl.pem"
 
 CERT_CHAIN_PEM="/secrets/keywhiz.pem"
 PRIVATE_KEY_PEM="/secrets/keywhiz-key.pem"
@@ -76,7 +76,7 @@ echo "Please copy the following files into the container:"
 echo "  1. Copy your certificate chain to $CERT_CHAIN_PEM"
 echo "  2. Copy your private key to $PRIVATE_KEY_PEM"
 echo "  3. Copy your CA bundle to $CA_BUNDLE_PEM"
-echo "  4. Copy a valid CRL file to $CRL_FILE_PEM"
+echo "  4. Copy a valid CRL file to $CRL_PATH"
 echo 
 echo "Files can be copied into a running container with docker cp."
 echo
@@ -87,7 +87,7 @@ echo
 assert_file_present $CERT_CHAIN_PEM
 assert_file_present $PRIVATE_KEY_PEM
 assert_file_present $CA_BUNDLE_PEM
-assert_file_present $CRL_FILE_PEM
+assert_file_present $CRL_PATH
 
 echo -n "Bundling certificate and private key into PKCS#12 keystore... "
 rm -f $KEYSTORE_PATH


worldwise001:
This may fix issue #376

hkdf/src/main/java/keywhiz/hkdf/Hkdf.java
```diff
@@ -66,15 +62,15 @@ private Hkdf(Hash hash, Provider provider) {
    * @return Hkdf constructed with default hash and derivation provider
    */
   public static Hkdf usingDefaults() {
-    return new Hkdf(DEFAULT_HASH, DEFAULT_PROVIDER);
+    return new Hkdf(DEFAULT_HASH, null);
```
csstaub: Provider field should be @Nullable now

server/src/test/java/keywhiz/auth/ldap/LdapAuthenticatorTest.java
```diff
@@ -27,19 +27,21 @@
 import java.util.Optional;
 import keywhiz.auth.User;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
-import org.junit.runner.RunWith;
+//import org.junit.runner.RunWith;
```
csstaub: Nit, kill this line? (and the other ones)

server/src/test/java/keywhiz/auth/ldap/LdapAuthenticatorTest.java
```diff
@@ -101,24 +117,7 @@ public void ldapAuthenticatorRejectsInvalidUsername() throws Exception {
...
-  private static class IsRoleSearch extends ArgumentMatcher<SearchRequest> {
-    @Override
-    public boolean matches(Object o) {
-      if (o == null) return false;
-      SearchRequest request = (SearchRequest) o;
-      return request.getBaseDN().equals("ou=roles,dc=example,dc=com");
-    }
-  }
-
+  @Ignore
```
csstaub: Justify the `@Ignore` here

server/src/test/java/keywhiz/service/crypto/ContentCryptographerTest.java
```diff
@@ -46,7 +45,7 @@
   }
 
   @Before public void setUp() throws Exception {
-    cryptographer = new ContentCryptographer(BASE_KEY, new SunJCE(), BC, FakeRandom.create());
+    cryptographer = new ContentCryptographer(BASE_KEY, BC, BC, FakeRandom.create());
```
csstaub: You can also `Security.getProvider("SunJCE")` if you want SunJCE, instead of calling the constructor

server/src/test/java/keywhiz/auth/ldap/LdapAuthenticatorTest.java
mcpherrinm: Can we just add a comment for now, and we can fix the tests later?

worldwise001: (resolved offline, @mcpherrinm is to add the comments)

server/src/test/java/keywhiz/auth/ldap/LdapAuthenticatorTest.java
worldwise001: @mcpherrinm can you add TODOs here for the commented imports?

hkdf/src/main/java/keywhiz/hkdf/Hkdf.java
worldwise001: Update: we'll fix annotations when we find a replacement for findbugs and similar. We are thinking of using spotbugs as a replacement.

server/src/test/java/keywhiz/auth/ldap/LdapAuthenticatorTest.java
mcpherrinm: I'm doing this

server/src/test/java/keywhiz/service/crypto/ContentCryptographerTest.java
mcpherrinm: I thought that didn't work anymore either, but @csstaub tells me I'm wrong.

Either way, we've got a BC here, so no harm in using it in this test, I think?

PR #383, opened 2018-10-29
Fix docker image with h2 profile

i-shcheglakov:
Docker container won't start out of the box.

1.  _/tmp/h2_data_ directory is created under _root_ ownership (on _mvn install_ step when testing performed) and when container is switched to _keywhiz_ migration step would fail with **permission denied** error.
```
keywhiz@0c938f337f4f:/usr/src/app$ ls /tmp/ -la
drwxr-xr-x 2 root    root    4096 Oct 29 16:54 h2_data
```
2. Starting the server requires _/tmp/h2_data_ volume to be persistent between _migrate_, _add-user_ and _server_ steps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,11 @@
 #     docker volume create --name keywhiz-db-devel
 #
 #   Initialize the database, apply migrations, and add administrative user:
-#     docker run -v keywhiz-db-devel:/data square/keywhiz migrate
-#     docker run -it -v keywhiz-db-devel:/data square/keywhiz add-user
+#     docker run --rm -v keywhiz-db-devel:/data square/keywhiz migrate
+#     docker run --rm -it -v keywhiz-db-devel:/data square/keywhiz add-user
 #
 #   Finally, run the server with the default development config:
-#     docker run -it -p 4444:4444 -v keywhiz-db-devel:/data square/keywhiz server
+#     docker run --rm -it -p 4444:4444 -v keywhiz-db-devel:/data square/keywhiz server
 #
 # *** Production setup wizard ***
 #   For production deployments, we have setup wizard that will initialize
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -18,11 +18,11 @@
 #     docker volume create --name keywhiz-db-devel
 #
 #   Initialize the database, apply migrations, and add administrative user:
-#     docker run -v keywhiz-db-devel:/data square/keywhiz migrate
-#     docker run -it -v keywhiz-db-devel:/data square/keywhiz add-user
+#     docker run -v keywhiz-db-devel:/tmp/h2_data square/keywhiz migrate
```
mbyczkowski: this is probably the simplest way to address the issue, but it creates a special case for making docker work with development config (if `KEYWHIZ_CONFIG` is not provided). It might be cleaner to create a new `keywhiz-development.yaml.docker` that sets H2 to use `/data` and is being used by `entry.sh` as a default.

i-shcheglakov: Yeah, I agree, that would be much cleaner way. I'll update and commit the changes.
Btw, do you usually amend changes in pull requests or commit new ones on top?

i-shcheglakov: I see. I amended and pushed for your review.

Haven't seen your changes but I'll check them. And, yeah, there should definitely be a way to verify that docker flow works properly in CI. We should consider that.

Dockerfile
```diff
@@ -45,7 +45,7 @@
 #
 FROM maven:3.5-jdk-11
 
-ARG MAVEN_PROFILE=h2
+ARG MAVEN_PROFILE=docker
```
mbyczkowski: 👍

model/pom.xml
```diff
@@ -81,6 +81,21 @@
         <jooq.input-schema>PUBLIC</jooq.input-schema>
       </properties>
     </profile>
+
+    <profile>
+      <id>docker</id>
+      <properties>
+        <skipResetDatabase>true</skipResetDatabase>
+        <db.driver>org.h2.Driver</db.driver>
+        <db.create-url>jdbc:h2:/tmp/h2_data/keywhizdb_test</db.create-url>
```
mbyczkowski: this should probably be `jdbc:h2:/data/keywhizdb_development` but I'm not sure if it's even being used by anything, so not a big deal.

PR #385, opened 2018-10-29
Update jackson-databind to 2.9.6 to address multiple vulnerabilities

mbyczkowski:
- related vulnerabilities:
  * CVE-2017-7525
  * CVE-2018-7489
  * CVE-2017-15095
  * CVE-2017-17485

- updated dropwizard to 1.2.9
  * removed references to dropwizard-java8 as it got pulled into the
    base by default
  * tweaked how SslContexFactory is created to support dropwizard/jetty
    API change
  * updated ciphersuites, so they work with default http client
    ciphersuites
  * some jetty changes under the hood forced changes to how we create an
    instance of `SslContextFactory` and how we use `Response` object to
    serialize cookies.

- updated guice to 4.2.2
- updated hibernate-validator to 5.3.4
- updated logback to 1.1.7
- updated auto-value to 1.5.4


api/src/main/java/keywhiz/api/LoginRequest.java
```diff
@@ -26,7 +26,7 @@
 @AutoValue
 public abstract class LoginRequest {
   @JsonCreator public static LoginRequest from(
-      @JsonProperty("username") String username, @JsonProperty("password") char[] password) {
+      @JsonProperty(value="username", required=true) String username, @JsonProperty(value = "password", required=true) char[] password) {
```
mbyczkowski: I was chasing an NPE between jackson, autovalue and jersey (after an update of dropwizard), and this change makes jackson explicitly fail deserialization rather than silently passing null values to auto-value.

api/src/main/java/keywhiz/api/CreateClientRequest.java
```diff
@@ -18,10 +18,10 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
-import org.hibernate.validator.constraints.NotEmpty;
+import org.hibernate.validator.constraints.NotBlank;
 
 public class CreateClientRequest {
-  @NotEmpty
+  @NotBlank
```
mbyczkowski: protects against NPEs

PR #387, opened 2018-10-30
Ensure that Docker image builds in CI

### LINKED ISSUE #280: Build docker image in CI
We should build the docker image in CI so we don't break it by accident.

mbyczkowski:
- Added a config line to the pom.xml, so keywhiz can report its version
- Added a tweak to Docker entrypoint, so that when `-v` or `--version`
  is passed as argument, keywhiz can report its version

Fixes #280

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: java
 
+services:
+  - docker
+
 jdk:
   - openjdk11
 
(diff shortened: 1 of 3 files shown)


csstaub:
Oh wow, it actually builds now. Last time I tried this I ran up against some timeouts or resources limits.

PR #389, opened 2018-11-01
Update dropwizard-raven library so it conforms to dropwizard 1.2.x API

mbyczkowski:
It seems that github.com/tradier/dropwizard-raven has stopped receiving updates
(particularly there's a
[PR](https://github.com/tradier/dropwizard-raven/pull/20) that
specifically makes the lib compatible with dropwizard 1.2.x that has not
been merged since Nov 7, 2017). The maintenance work continues in the
[fork](github.com/dhatim/dropwizard-sentry) and there's one 1.2.x JAR
release under org.dhatim.dropwizard-raven. The project has seen been
renamed to dropwizard-sentry and new versions of the library (1.3.x+)
are published under org.dhatim.dropwizard-sentry, which I have noted in
the comments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,8 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <dropwizard.version>1.2.9</dropwizard.version>
+    <!-- 1.2.0-1 is the last version of dropwizard-raven JAR. The development continues under dropwizard-sentry -->
+    <dropwizard-raven.version>1.2.0-1</dropwizard-raven.version>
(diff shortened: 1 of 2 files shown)


mbyczkowski:
@jbpeirce this is using exactly https://github.com/dhatim/dropwizard-sentry, however the first release (`1.2.0-1`) was published as `org.dhatim.dropwizard-raven` and every other release after that was published as `org.dhatim.dropwizard-sentry`. I decided to use the `org.dhatim.dropwizard-raven` JAR because we use dropwizard `1.2.9`. It definitely makes sense to bump this lib once we upgrade to dropwizard `1.3.x`.

PR #392, opened 2018-11-09
Update .gitignore to exclude auto-generated docs files

i-shcheglakov:
Auto generated `docs/apidocs/*.json` files pollute `git status.` They should be either updated or excluded from tracking altogether. I chose the latter, but we should update them if they should be tracked.


csstaub:
I _think_ these might be used by the GitHub page. @mcpherrinm do you know?

mbyczkowski:
@i-shcheglakov #394 got merged. Can you check if you're still seeing stale docs issue?

alokmenghrajani:
We are deleting the docs (see https://github.com/square/keywhiz/pull/402)

PR #401, opened 2019-01-24
Add index by group ID on the memberships table

jbpeirce:
This should improve the behavior of the lookup for clients
associated with a secret, which requires JOINing on
the group ID in the memberships table.


jbpeirce:
Force-pushed to change the capitalization on the files (from v8__... to V8__...) so that Flyway will properly detect them.

PR #402, opened 2019-03-14
Nuke docs

alokmenghrajani:
Swagger has a XSS
(https://github.com/swagger-api/swagger-ui/issues/3847) and this is the
easiest way to fix things since we don't really care about the API docs.


mcpherrinm:
Should we also delete the javadocs from the repo?  Just removing from the pom doesn't remove them from the site, since they're checked in.

PR #406, opened 2019-04-19
Add command to remove database rows for deleted secrets

jbpeirce:
Deleted secrets in Keywhiz are still present in the database;
the only difference between a deleted secret and an existing secret
is that `secrets.current` for a deleted secret is null, while it's
the ID of a row in the `secrets_content` table for an existing
secret.

This adds a command to remove database rows for secrets which have
been deleted before an input date.  This command must be treated
extremely carefully, since its effects are permanent unless the
database is restored from a backup.

This command will be valuable in removing unused deleted secrets
since a recent commit changed the naming pattern for deleted secrets.

### DIFF
--- a/server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
+++ b/server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
@@ -1,0 +1,110 @@
+package keywhiz.commands;
+
+import io.dropwizard.cli.ConfiguredCommand;
+import io.dropwizard.setup.Bootstrap;
+import java.io.Console;
+import java.util.Scanner;
+import javax.inject.Inject;
+import keywhiz.KeywhizConfig;
+import keywhiz.service.daos.SecretDAO;
+import net.sourceforge.argparse4j.inf.Namespace;
+import net.sourceforge.argparse4j.inf.Subparser;
+import org.joda.time.DateTime;
+
+import static java.lang.String.format;
+
+/**
+ * Secrets which are deleted from Keywhiz--whether through the admin client or the automation
+ * interface--are not actually removed from Keywhiz' database.  Instead, the link between the
+ * "secret series" and "secret contents" tables is broken, but it is still possible to reconnect
+ * them manually and restore a deleted secret.
+ *
+ * This command permanently drops information about deleted secrets from the database.  This is
+ * clearly dangerous, so it should be run only when it is certain that the deleted secrets will not
+ * need to be manually recovered.
+ */
+public class DropDeletedSecretsCommand extends ConfiguredCommand<KeywhizConfig> {
+  protected static final String INPUT_DELETED_BEFORE = "deleted-before";
+  private SecretDAO secretDAO;
+
+  @Inject
+  protected DropDeletedSecretsCommand(SecretDAO secretDAO) {
+    super("drop-deleted-secrets", "PERMANENTLY REMOVES database records for deleted secrets");
+    this.secretDAO = secretDAO;
+  }
+
+  @Override public void configure(Subparser subparser) {
+    subparser.addArgument("--deleted-before")
+        .dest(INPUT_DELETED_BEFORE)
+        .type(String.class)
+        .required(true)
+        .help(
+            "secrets deleted before this date will be PERMANENTLY REMOVED.  Format: 2006-01-02T15:04:05Z");
+  }
+
+  @Override public void run(Bootstrap<KeywhizConfig> bootstrap, Namespace namespace,
+      KeywhizConfig config) {
+    // validate the input
+    String deletedBeforeStr = namespace.getString(INPUT_DELETED_BEFORE);
+    DateTime deletedBefore = getDateIfValid(deletedBeforeStr);
+    if (deletedBefore == null) {
+      return;
+    }
+
+    // determine how many secrets would be affected and get user confirmation
+    long affectedCount = secretDAO.countSecretsDeletedBeforeDate(deletedBefore);
+
+    System.out.format(
+        "WARNING: This will PERMANENTLY remove all secrets deleted before %s.  %d secrets will be removed.  Confirm? (y/n)\n",
+        deletedBefore.toString(), affectedCount);
+
+    String confirm;
+    if (System.console() != null) {
+      Console console = System.console();
+      confirm = console.readLine();
+    } else {
+      // In test environments, system.console may not be set; fall back to System.in.
+      Scanner scanner = new Scanner(System.in);
+      confirm = scanner.nextLine();
+    }
+    if (confirm == null || !(confirm.equalsIgnoreCase("y") || confirm.equalsIgnoreCase("yes"))) {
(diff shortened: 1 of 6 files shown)


jbpeirce:
I tried to be consistent in using "remove" for removing[/deleting] rows from the database and "delete" for soft-deleting secrets from Keywhiz; please point out any inconsistencies.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -385,6 +403,48 @@ public void deleteSecretsByName(String name) {
...
+  public void dangerPermanentlyRemoveSecretsDeletedBeforeDate(DateTime deleteBefore) {
+    checkArgument(deleteBefore != null);
+    SecretSeriesDAO secretSeriesDAO = secretSeriesDAOFactory.using(dslContext.configuration());
+    SecretContentDAO secretContentDAO = secretContentDAOFactory.using(dslContext.configuration());
+
+    // identify the secrets deleted before this date
+    List<Long> ids = secretSeriesDAO.getIdsForSecretSeriesDeletedBeforeDate(deleteBefore);
+
+    // batch the list of secrets to be removed, to reduce load on the database
+    List<List<Long>> partitionedIds = Lists.partition(ids, MAX_ROWS_REMOVED_PER_TRANSACTION);
```
mcpherrinm: It might be nice to have a sleep between batches, so we can tune how much load this incurs

jbpeirce: Good idea--added.

server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
```diff
@@ -0,0 +1,129 @@
...
+import static java.lang.String.format;
+
+/**
+ * Secrets which are deleted from Keywhiz--whether through the admin client or the automation
+ * interface--are not actually removed from Keywhiz' database.  Instead, the link between the
+ * "secret series" and "secret contents" tables is broken, but it is still possible to reconnect
+ * them manually and restore a deleted secret.
+ *
+ * This command permanently drops information about deleted secrets from the database.  This is
+ * clearly dangerous, so it should be run only when it is certain that the deleted secrets will not
```
stfinney: One possible option to make this less dangerous would be to put a minimum threshold into the code, so no matter what date you enter, it will never drop things that haven't been deleted for at least a week (or a month or whatever, just an arbitrary number for the example).

jbpeirce: True; it seems, though, like what threshold is safe is going to depend fairly heavily on how Keywhiz is being used (and setting too strict of a threshold will make this endpoint less useful).

stfinney: That's definitely true that it depends on how you use keywhiz. What if it was a configurable value?

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -385,6 +403,53 @@ public void deleteSecretsByName(String name) {
...
+  }
+
+  /**
+   * Permanently removes the series and all secrets-contents records ("versions") which were deleted
+   * before the given date.
+   *
+   * Unlike the "delete" endpoints above, THIS REMOVAL IS PERMANENT and cannot be undone by editing
+   * the database to restore the "current" entries.
+   *
+   * @param deleteBefore the cutoff date; secrets deleted before this date will be deleted
```
stfinney: nit: I think the second "deleted" here should be "removed" to avoid ambiguity.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -385,6 +403,53 @@ public void deleteSecretsByName(String name) {
...
+    SecretContentDAO secretContentDAO = secretContentDAOFactory.using(dslContext.configuration());
+
+    // identify the secrets deleted before this date
+    List<Long> ids = secretSeriesDAO.getIdsForSecretSeriesDeletedBeforeDate(deleteBefore);
+
+    // batch the list of secrets to be removed, to reduce load on the database
+    List<List<Long>> partitionedIds = Lists.partition(ids, MAX_ROWS_REMOVED_PER_TRANSACTION);
+    for (List<Long> idBatch : partitionedIds) {
+      // permanently remove the `secrets_contents` entries originally associated with these secrets
+      secretContentDAO.dangerPermanentlyRemoveRecordsForGivenSecretsIDs(idBatch);
```
stfinney: Should both the deletion operations occur within a transaction? If the first operation succeeds and the second fails, we're just left with secret series record(s) with no content, which I guess is not too bad? They would get deleted the next time the command is used.

server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
```diff
@@ -0,0 +1,129 @@
...
+    if (sleepMillis < 0) {
+      System.out.format("Milliseconds to sleep must be nonnegative; got %d", sleepMillis);
+      return;
+    }
+
+    // determine how many secrets would be affected and get user confirmation
+    long affectedCount = secretDAO.countSecretsDeletedBeforeDate(deletedBefore);
+
+    System.out.format(
+        "WARNING: This will PERMANENTLY remove all secrets deleted before %s.  %d secrets will be removed.  Confirm? (y/n)\n",
```
stfinney: It would make it run a little slower, but it might be nice to also display the total number of deleted secrets, so that the number that will be removed has a frame of reference. If it says it's removing 100 secrets, I might feel differently about whether that is correct or not if there are 1000 total vs 100 total.

server/src/test/java/keywhiz/commands/DropDeletedSecretsCommandTest.java
```diff
@@ -0,0 +1,289 @@
...
+        .set(SECRETS_CONTENT.METADATA,
+            objectMapper.writeValueAsString(secret3.content().metadata()))
+        .execute();
+  }
+
+  @Test
+  public void testSecretDeletion_allDeletedSecrets() {
+    runCommandWithConfirmationAndDate("yes", "2019-02-01T00:00:00Z", 0);
+
+    // check the database state; secret2 and secret3 should have been deleted
```
stfinney: nit: s/deleted/removed
Also applies to similar comment in the other tests.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
jbpeirce: The next run would clean up the secret series, so I'm not too concerned.  Generally the two tables have been handled separately by their DAOs.

server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
jbpeirce: That could work; I'm not convinced that it adds a lot of protection beyond what's already provided by the configurable `deletedBefore` argument and the confirmation prompts, though.

PR #407, opened 2019-04-24
Register command with Keywhiz server and fix parameter error

jbpeirce:
This adds a method to manually construct a SecretDAO instance
from an injected DSLContext.  This is more error-prone than using
injection; however, injection doesn't appear to be available at the
bootstrapping phase.

### DIFF
--- a/server/src/main/java/keywhiz/KeywhizService.java
+++ b/server/src/main/java/keywhiz/KeywhizService.java
@@ -32,9 +32,11 @@ import keywhiz.auth.mutualssl.ClientCertificateFilter;
 import keywhiz.auth.xsrf.XsrfServletFilter;
 import keywhiz.commands.AddUserCommand;
 import keywhiz.commands.DbSeedCommand;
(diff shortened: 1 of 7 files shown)


jbpeirce:
An alternative design would be to not use a SecretDAO/SecretSeriesDAO/SecretContentsDAO at all, and just manually edit the database from this command (similarly to how the MigrateCommand directly writes data), or create a dedicated helper class with a simpler structure that just requires a DslContext to start (like the AddUserCommand and its UserDAO).

stfinney:
> An alternative design would be to not use a SecretDAO/SecretSeriesDAO/SecretContentsDAO at all, and just manually edit the database from this command (similarly to how the MigrateCommand directly writes data), or create a dedicated helper class with a simpler structure that just requires a DslContext to start (like the AddUserCommand and its UserDAO).

That does sound like it might be simpler. But you've already done it this way, not sure it's worth going back and re-writing it now.

PR #408, opened 2019-04-26
Add unique secret name constraint

jbpeirce:
This re-adds the database constraint that secrets must have unique names which was removed in #337 (that attempt to enforce uniqueness through code proved prone to race conditions).  The original issue (users being able to access the contents of deleted secrets by making a new secret with the same name) is now addressed by renaming secrets to an invalid name once they are deleted.

This PR still does not allow retrieving deleted secret series (a constraint which was introduced in #337).  Since all uses of the secret series retrieval that I could find assumed the series were non-deleted if present (including uses which weren't changed in #337), changing that assumption seemed likely to cause subtle bugs or at least unexpected behavior (such as being able to grant groups access to deleted secrets after deletion).

In addition, this PR cleans up some test files' assertions about Optionals.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
@@ -215,8 +215,13 @@ public class SecretSeriesDAO {
   public void deleteSecretSeriesByName(String name) {
     long now = OffsetDateTime.now().toEpochSecond();
     dslContext.transaction(configuration -> {
+      // find the record and lock it until this transaction is complete
       SecretsRecord r = DSL.using(configuration)
-          .fetchOne(SECRETS, SECRETS.NAME.eq(name).and(SECRETS.CURRENT.isNotNull()));
+          .select()
+          .from(SECRETS)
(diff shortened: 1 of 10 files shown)


jbpeirce:
Force-push to rebase on master

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -238,17 +237,21 @@ public void deleteSecretSeriesByName(String name) {
...
-          .set(SECRETS.UPDATEDAT, now)
-          .where(SECRETS.ID.eq(id))
-          .execute();
-
-      DSL.using(configuration)
-          .delete(ACCESSGRANTS)
-          .where(ACCESSGRANTS.SECRETID.eq(id))
-          .execute();
+      SecretsRecord r = DSL.using(configuration)
+          .fetchOne(SECRETS, SECRETS.ID.eq(id).and(SECRETS.CURRENT.isNotNull()));
```
mcpherrinm: I think this should maybe call `.forUpdate()`, so the row is locked until the subsequent UPPDATE

jbpeirce: Updated to use forUpdate on this select.

PR #413, opened 2019-06-05
Simplify the SecretsResponse equals method

WilliamCollishaw:


csstaub:
Looks good. Thanks!

WilliamCollishaw:
@csstaub there are a lot more equals methods in the api that can be simplified but are not as elegant as this and #412. I am trying to decide if I should do the remaining simplifying for consistency. Please see sample below:
Screen Shot 2019-06-06 at 9 44 53 PM

Opinion?

csstaub:
Not sure that's worth it, to be honest. All that code was auto-generated anyway and if it's a few lines longer than it needs to be that's not really a big deal.

WilliamCollishaw:
Sounds good to me! Thanks for the speedy response :)

alokmenghrajani:
Could we switch to [@AutoValue](https://github.com/google/auto/blob/master/value/userguide/index.md)?

PR #414, opened 2019-06-05
Replace deprecated Mockito Matchers class with ArgumentMatchers

WilliamCollishaw:
Additionally remove unused imports from classes containing the org.mockito.Matchers imports. Some classes imported the deprecated class and were no longer using so removed those also.

Matchers class got replaced with ArguementMatchers, the following is from the [Javadoc](https://static.javadoc.io/org.mockito/mockito-core/2.23.0/org/mockito/Matchers.html):

> Deprecated. 
> Use ArgumentMatchers. This class is now deprecated in order to avoid a name clash with Hamcrest org.hamcrest.Matchers class. This class will likely be removed in version 3.0.


csstaub:
Thank you for your contribution!

PR #416, opened 2019-06-06
Remove redundant static declaration from nested enum

WilliamCollishaw:
Nested enum types are implicitly static, resulting in a redundant declaration


WilliamCollishaw:
😮 speedy

alokmenghrajani:
everyone's asleep... Best time to be productive :)

Issue #182, opened 2016-03-14
Distrust the Database


mcpherrinm:
Our current threat model requires the database acting properly, and not changing data beneath us.

We rely on database joins to compute ACLs.  This means a malicious database could mis-compute those joins, and return a list of secrets a client shouldn't have access to.

While the database can't directly steal secrets (they're encrypted), we shouldn't rely on its correct behaviour.

We should explore signing database objects, and verifying those signatures on every retrieval.

alokmenghrajani:
:+1:

alokmenghrajani:
🎉

PR #438, opened 2019-07-31
Generate and verify row_hmacs when creating and retrieving secrets

captiosus:
In reference to issue #182, we are adding a new column to clients, secrets, accessgrants, memberships, and secrets_content to protect against db admins tampering with the database. We generate an hmac for each of these tables to ensure that data can't be tampered without access to the private key. 

Instead of auto-incrementing id's we generate random 64 bit id's which are concatenated with the other identifying columns in each respective table (ie. for client we use the name column and for secrets_content we use the encrypted content). This value is then hmac'd, stored, and verified whenever a secret is retrieved.

The hmac verification is only guaranteed on Secret(s)DeliveryResource. Other locations are either admin endpoints or only show secret details and not actual secret content. 

This change will not affect users beyond having to set a new flag in the config. Setting **rowHmacCheck** to disabled will turn off row_hmac verification. The row_hmac column will still be set even if this flag is disabled. A future PR will add backfills to generate row_hmacs for prior database entries. The migration will default to an empty string and with **rowHmacCheck** disabled, this should not affect keywhiz operations.


server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -86,6 +86,13 @@
   @JsonProperty
   private String statusCacheExpiry;
 
+  @JsonProperty
+  private String rowHmacCheck;
+
+  public enum RowHmacCheck {
+    DISABLED, LOGGING, ENABLED
```
alokmenghrajani: `DISABLED`, `DISABLED_BUT_LOG`, `ENFORCE` might be better word choices. It's not immediately clear that LOGGING implies it only logs.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -176,6 +183,15 @@ public String getDerivationProviderClass() {
     return backupExportKeys.get(name);
   }
 
+  public RowHmacCheck getRowHmacCheck() {
+    if (rowHmacCheck.equals("enabled")) {
```
alokmenghrajani: a switch statement with an exception for the default case is cleaner.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -176,6 +183,15 @@ public String getDerivationProviderClass() {
...
+      return RowHmacCheck.ENABLED;
+    } else if (rowHmacCheck.equals("logging")) {
+      return RowHmacCheck.LOGGING;
+    }
```
alokmenghrajani: Having a typo, such as "enbled", result in the check being disabled is bad.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -48,17 +51,30 @@
...
+    this.cryptographer = cryptographer;
+    this.random = random;
   }
 
   public long createClient(String name, String user, String description) {
     ClientsRecord r = dslContext.newRecord(CLIENTS);
 
     long now = OffsetDateTime.now().toEpochSecond();
 
+    byte[] generateIdBytes = new byte[8];
```
alokmenghrajani: make the "give me a random long" a helper function.

server/src/main/java/keywhiz/KeywhizConfig.java
captiosus: Whats the alternative? Defaulting to enforced could also be problematic

alokmenghrajani: Throw an exception. This code runs at app startup time and the app won't come up if this field is not properly configured.

alokmenghrajani:
This PR needs a better title and a description explaining what we are doing, what the migration is going to look like, etc.

PR #444, opened 2019-08-02
remove h2 and convert travis-ci to use mysql

captiosus:
h2 was beginning to diverge from mysql and started to affect tests. Maintaining both h2 and mysql caused too much friction. Also, having tests on mysql better mimics the production environment


.gitignore
```diff
@@ -14,5 +14,3 @@ website/apidocs/
 *.ipr
 *.iws
 .DS_Store
-server/src/main/resources/keywhiz-development.yaml
```
alokmenghrajani: We might want to keep this if people want to configure their dev keywhiz and avoid accidentally committing it.

mcpherrinm: We have a keywhiz-development.yaml in the repo now, though.

Previously we copied this from a mysql/h2 specific version in maven, which is why it was ignored

PR #461, opened 2019-08-05
generate and verify row_hmac

captiosus:
In reference to issue #182, we are adding a new column to clients, secrets, accessgrants, memberships, and secrets_content to protect against db admins tampering with the database. We generate an hmac for each of these tables to ensure that data can't be tampered without access to the private key.

Instead of auto-incrementing id's we generate random 64 bit id's which are concatenated with the other identifying columns in each respective table (ie. for client we use the name column and for secrets_content we use the encrypted content). This value is then hmac'd, stored, and verified whenever a secret is retrieved.

The hmac verification is only guaranteed on Secret(s)DeliveryResource. Other locations are either admin endpoints or only show secret details and not actual secret content.

This change will not affect users beyond having to set a new flag in the config. Setting rowHmacCheck to disabled will turn off row_hmac verification. The row_hmac column will still be set even if this flag is disabled. A future PR will add backfills to generate row_hmacs for prior database entries. The migration will default to an empty string and with rowHmacCheck disabled, this should not affect keywhiz operations.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -27,6 +27,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import keywhiz.KeywhizConfig;
+import keywhiz.KeywhizConfig.RowHmacCheck;
 import keywhiz.api.ApiDate;
 import keywhiz.api.model.Client;
 import keywhiz.api.model.Group;
@@ -40,6 +42,7 @@ import keywhiz.log.AuditLog;
 import keywhiz.log.Event;
 import keywhiz.log.EventTag;
 import keywhiz.service.config.Readonly;
+import keywhiz.service.crypto.RowHmacGenerator;
 import keywhiz.service.daos.ClientDAO.ClientDAOFactory;
 import keywhiz.service.daos.GroupDAO.GroupDAOFactory;
 import keywhiz.service.daos.SecretContentDAO.SecretContentDAOFactory;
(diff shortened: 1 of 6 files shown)


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -329,26 +324,93 @@ public void findAndEvictClient(long clientId, long groupId, AuditLog auditLog, S
...
+        .map(row -> processSanitizedSecretRow(row, client));
+  }
+
+  private SanitizedSecret processSanitizedSecretRow(Record row, Client client) {
+    boolean rowHmacLog = config.getRowHmacCheck() == RowHmacCheck.DISABLED_BUT_LOG;
+    boolean rowHmacFail = config.getRowHmacCheck() == RowHmacCheck.ENFORCED;
+
+    SecretSeries series = secretSeriesMapper.map(row.into(SECRETS));
+
+    String secretHmac = rowHmacGenerator.computeRowHmac(
```
alokmenghrajani: There's probably a way to make this code less repetitive. It's fine for now.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -128,7 +145,27 @@ public long createSecretContent(long secretId, String encryptedContent, String h
...
+    }
+
+    String rowHmac = rowHmacGenerator.computeRowHmac(
+        SECRETS_CONTENT.getName(), r.getEncryptedContent(), r.getId());
+
+    if (!rowHmac.equals(r.getRowHmac())) {
+      String errorMessage = String.format(
+          "Secret Content HMAC verification failed for secretContent: %d", r.getId());
+      if (config.getRowHmacCheck() == RowHmacCheck.DISABLED_BUT_LOG) {
+        logger.info(errorMessage);
```
mbyczkowski: @captiosus this should probably be a warning?

PR #467, opened 2019-08-07
add metadata field to row_hmac for secrets_content

captiosus:
keysync uses the metadata field in secrets_content for file permissions when writing secrets to disk. we need to include this value in the hmac to prevent secrets from being made world-readable by attackers

### DIFF
--- a/server/src/main/java/keywhiz/commands/DbSeedCommand.java
+++ b/server/src/main/java/keywhiz/commands/DbSeedCommand.java
@@ -104,6 +104,11 @@ public class DbSeedCommand extends ConfiguredCommand<KeywhizConfig> {
     return computeHmac(hmacContent.getBytes(UTF_8));
   }
 
+  private static String computeRowHmac(String table, String content, String metadata, long id) {
+    String hmacContent = table + "|" + content + "|" + metadata + "|" + id;
+    return computeHmac(hmacContent.getBytes(UTF_8));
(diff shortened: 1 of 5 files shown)


server/src/main/java/keywhiz/commands/DbSeedCommand.java
```diff
@@ -104,6 +104,11 @@ private static String computeRowHmac(String table, long id1, long id2) {
     return computeHmac(hmacContent.getBytes(UTF_8));
   }
 
+  private static String computeRowHmac(String table, String content, String metadata, long id) {
```
alokmenghrajani: I still think `computeRowHmac(String table, Vector<String> fields)` would be cleaner.

server/src/main/java/keywhiz/service/crypto/RowHmacGenerator.java
```diff
@@ -35,6 +35,11 @@ public String computeRowHmac(String table, long id1, long id2) {
     return cryptographer.computeHmacWithSecretKey(hmacContent.getBytes(UTF_8), hmacKey);
   }
 
+  public String computeRowHmac(String table, String content, String metadata, long id) {
```
alokmenghrajani: copy pasta with DbSeedCommand.java. DbSeedCommand.java should just call this version.

PR #488, opened 2019-08-13
refactor backfills to fill in all missing row_hmacs

captiosus:


server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,170 +62,210 @@
...
-      @PathParam("secretIdFrom") Long secretIdFrom, @PathParam("secretIdTo") Long secretIdTo) {
-    int modifiedRows = 0;
-    logger.info("backfill-secrets: processing secrets from {} to {}", secretIdFrom, secretIdTo);
-    for (long secretId = secretIdFrom; secretId < secretIdTo; secretId++) {
-      Optional<SecretSeriesAndContent> secret = secretDAO.getSecretById(secretId);
-
-      if (secret.isEmpty()) {
-        logger.info("backfill-secrets {}: does not exist", secretId);
-        continue;
+  public void backfillSecretsRowHmac() {
```
mbyczkowski: It would be good to be able to skip over a number of records, e.g. if the job crashes at some point you can skip over the records that were already processed.

mbyczkowski: Also, it might be a good idea to have a max ID, so you can ensure that the job ends after processing a number of records, say 1M.

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,170 +62,210 @@
...
-    logger.info("backfill-secrets-content: processing secrets content from {} to {}",
-        secretContentIdFrom, secretContentIdTo);
-    for (long secretContentId = secretContentIdFrom; secretContentId < secretContentIdTo; secretContentId++) {
-      Optional<SecretContent> secretContent =
-          secretContentDAO.getSecretContentById(secretContentId);
-
-      if (secretContent.isEmpty()) {
-        logger.info("backfill-secrets-content {}: does not exist", secretContentId);
-        continue;
+  public void backfillSecretsContentRowHmac() {
```
mbyczkowski: same comment as above

mbyczkowski:
There seem to be some ongoing issues with coveralls.io that broke the CI, but the tests are still passing.

PR #489, opened 2019-08-13
change logs to warn

captiosus:


mcpherrinm:
Why are we truncating the hmac?

captiosus:
> Why are we truncating the hmac?

16 bytes should be enough randomness to avoid collisions. mostly to save space

mcpherrinm:
Doesn't seem like a worthwhile tradeoff to me.  It's a relatively small amount of data, all things considered -- we waste plenty more space elsewhere if we're concerned about it.

PR #490, opened 2019-08-13
Do not collect warnings by default

jbpeirce:
This saves an extra query on each database transaction.  Currently,
SHOW WARNINGS is a common and essentially unused query in Keywhiz'
regular operations.


mcpherrinm:
We'd miss any mysql warnings this way, which I'd definitely want to see in dev/test/staging environments.  Can we make this configurable?

alokmenghrajani:
jOOQ supports reading settings from a XML file. We should probably do that. Then have different xml files for different envs.

jbpeirce:
True.  I'll close this for now, and return to it later if we're seeing performance impacts.

PR #492, opened 2019-08-14
remove backfill from master, but keep in backfill_row_hmac branch

captiosus:


alokmenghrajani:
lgtm. Let's merge this after we flip the switch and we are sure we no longer need this backfill.

PR #509, opened 2019-09-06
Remove unused parameters from expiration-v3 endpoint

jbpeirce:
The pagination implementation in the v4 expiration retrieval
endpoint for automation clients is significantly easier to work
with than the implementation in v3.  Since the v3 implementation
has not been released, this removes its extra parameters (which
allows some simplifications to the internal secrets-management
code).


mbyczkowski:
for context, this reverts changes in #503 (which itself was superseded by #506)

PR #519, opened 2019-09-25
Update jackson-databind to 2.9.10

mbyczkowski:


pom.xml
```diff
@@ -41,7 +41,7 @@
     <!-- 1.2.0-1 is the last version of dropwizard-raven JAR. The development continues under dropwizard-sentry -->
     <dropwizard-raven.version>1.2.0-1</dropwizard-raven.version>
     <guice.version>4.2.2</guice.version>
-    <jackson.version>2.9.9.1</jackson.version>
+    <jackson.version>2.9.10</jackson.version>
     <!-- Some jackson libraries have not received a patch update, so peg them to the latest 2.9.x -->
```
jbpeirce: Are these libraries still not updated?

mbyczkowski: Good point! We should definitely fold it now.

Issue #464, opened 2019-08-06
Flyway 6.0 blockers


alokmenghrajani:
We currently can't upgrade to Flyway 6.0 for the following reasons:
- We are calling deprecated methods which go away in 6.0 (`.setDataSource`, `.setLocations`).
- The schema table changed names, from "schema_version" to "flyway_schema_history".

PR #526, opened 2019-10-07
Update Flyway to 6.0.4

123Haynes:
This PR updates Flyway to 6.0.4 and changes the programatical
initialization to the fluent api provided by flyway.

I also set the table name for the schema versioning explicitly because the default name changed.

Closes https://github.com/square/keywhiz/issues/464


123Haynes:
@mbyczkowski sure, I'll work on that in the next few days when i have some free minutes :)

123Haynes:
@mbyczkowski
I **think** this is ready for a second review. I hope I got all places where the Flyway schema version table needs to be set-

PR #549, opened 2019-11-14
Fix admin client to handle negative secret IDs correctly

jbpeirce:
The version-listing and rollback admin client endpoints initially
assumed that negative secret IDs were impossible.  However, since
secret IDs are now randomly generated, the endpoints must handle
negative values properly.

This also fixes a bug in the display of secret creators, adds a display of metadata, and improves the test for listing secret versions.

A manual test which included rolling back to a previous (negative) version ID:
Screen Shot 2019-11-14 at 11 22 24 AM

### DIFF
--- a/cli/src/main/java/keywhiz/cli/Printing.java
+++ b/cli/src/main/java/keywhiz/cli/Printing.java
@@ -25,6 +25,8 @@ import java.text.SimpleDateFormat;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import keywhiz.api.ClientDetailResponse;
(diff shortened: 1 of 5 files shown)


cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -56,7 +58,8 @@ public void printClientWithDetails(Client client) {
     if (clientDetails.lastSeen == null) {
       System.out.println("\tLast Seen: never");
     } else {
-      String lastSeen = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss").format(new Date (clientDetails.lastSeen.toEpochSecond()*1000));
+      String lastSeen = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss").format(
```
alokmenghrajani: nit: we should render the timezone used for time conversion.

PR #667, opened 2020-05-29
Add spiffe uri to clients

jbpeirce:
This adds a SPIFFE URI field to clients. Currently, each client can only be paired with a single SPIFFE URI, which simplifies the migration from the existing client certificates to identifying certificates based on SPIFFE URIs.

Later PRs will add a new endpoint and update the existing secret retrieval endpoints to consider SPIFFE URIs on clients, if present.

This PR is currently based on https://github.com/square/keywhiz/pull/666, since updating the admin API in that PR reduced the number of changes needed to support creating clients with SPIFFE URIs through the automation/v2 and admin APIs.


cli/src/test/java/keywhiz/cli/commands/AddActionTest.java
```diff
@@ -87,22 +90,27 @@ public void addCallsAddForSecret() throws Exception {
...
+    verify(keywhizClient, times(1)).createSecret(secret.getName(), "", content,
+        secret.getMetadata(), 1136214245);
   }
 
   @Test
   public void addCallsAddForClient() throws Exception {
     addActionConfig.addType = Arrays.asList("client");
     addActionConfig.name = client.getName();
+    addActionConfig.description = "description";
+    addActionConfig.spiffeUri = "spiffe//test.com/client";
```
mcpherrinm, line 107: minor nit, but `example.org` is the best choice for test data, as it is explicitly set aside for the purposes of testing -- not that it matters too much in cases like this

cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -71,6 +63,42 @@ public void printClientWithDetails(Client client) {
     clientDetails.secrets.stream()
         .sorted(Comparator.comparing(SanitizedSecret::name))
         .forEach(s -> System.out.println(INDENT + SanitizedSecret.displayName(s)));
+
+    if (clientDetails.lastSeen == null) {
+      System.out.println("\tLast Seen: never");
+    } else {
+      String lastSeen = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss").format(
```
mcpherrinm, line 70: I know you're not changing this here, but we inconsistently use the yyyy-mm-dd format and MM/dd/yyyy formats in keywhiz.  It would be good to standardize (on yyyy-mm-dd I think)

cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -71,6 +63,42 @@ public void printClientWithDetails(Client client) {
...
+          new Date(clientDetails.lastSeen.toEpochSecond() * 1000));
+      System.out.printf("\tLast Seen: %s%n", lastSeen);
+    }
+
+    if (!clientDetails.description.isEmpty()) {
+      System.out.println("\tDescription:");
+      System.out.println(INDENT + clientDetails.description);
```
mcpherrinm, line 77: Kinda weird we have a constant INDENT for "\t\t" but then a literal "\t" in the string above it -- again, I know you're not adding this pattern, but it is weird.

cli/src/test/java/keywhiz/cli/commands/AddActionTest.java
```diff
@@ -52,10 +52,13 @@
   AddActionConfig addActionConfig;
   AddAction addAction;
 
-  Client client = new Client(4, "newClient", null, null, null, null, null, null, null, true, false, null);
+  Client client =
+      new Client(4, "newClient", null, null, null, null, null, null, null, true, false, null);
```
isemaya-square, line 56: I'm sure the code here is functional (and looked like this beforehand, so no issues were introduced in this PR), but aren't there a lot of arguments for this constructor?

server/src/main/resources/db/mysql/migration/V11__add_spiffe_uri_to_clients.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE clients ADD spiffe_uri varchar(255);
```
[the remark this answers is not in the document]
mweissbacher: +1 re length - this is a good catch

cli/src/test/java/keywhiz/cli/commands/AddActionTest.java
jbpeirce: It definitely is. In a later PR, I'd like to move all of these to the builder pattern.

cli/src/main/java/keywhiz/cli/Printing.java
jbpeirce: For now, I'll standardize this to match the other date formats used by this class--DateFormat.getDateTimeInstance().format(<date>).

server/src/main/resources/db/mysql/migration/V11__add_spiffe_uri_to_clients.sql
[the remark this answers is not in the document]
jbpeirce: The rename and rearranging ended up being a significant refactor, so I combined it and the initial small commits to incrementally add the ID to clients into a single commit (plus a couple of documentation/cleanup commits).

server/src/main/resources/db/mysql/migration/V11__add_spiffe_id_to_clients.sql
```diff
@@ -0,0 +1,4 @@
+# https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#23-maximum-spiffe-id-length
+# maximum SPIFFE ID length is 2048 bytes using (per https://tools.ietf.org/html/rfc3986) US-ASCII
+# chars, which in utf-8 encoding occupy one byte each
+ALTER TABLE clients ADD spiffe_id varchar(2048);
```
mbyczkowski, line 4: Something I should have caught in the 1st pass, but didn't. `clients.name` has unique index. Do we expect `clients.spiffe_id` to also be unique? If so, then there might be a [Mysql limitation there](https://dev.mysql.com/doc/refman/5.7/en/innodb-limits.html) -- although given appropriate config is in place, this should be fine.

jbpeirce: We actually don't expect `spiffe_id` to be unique. If this is used alongside Keysync, for instance, we may give a single SPIFFE ID to each Keysync instance, link that ID to all of the clients that Keysync needs to sync secrets for, and use that to retrieve secrets for multiple clients in one call.

PR #685, opened 2020-06-26
Retrieve clients based on their SPIFFE IDs

jbpeirce:
This adds the constraint that SPIFFE IDs are unique per-client, since supporting non-unique SPIFFE IDs requires a significant redesign of Keywhiz' trust model. It also refactors client authorization to separate extracting information about the client from a request (either from the security context or the XFCC header) and retrieving a client based on that information into separate classes.

Note that none of these changes affect automation clients. A later PR will update the automation client factory to share code with the client creation factory, since the logic is almost entirely identical.

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/ClientDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/ClientDAO.java
@@ -136,8 +136,13 @@ public class ClientDAO {
     }
   }
 
-  public Optional<Client> getClient(String name) {
+  public Optional<Client> getClientByName(String name) {
     ClientsRecord r = dslContext.fetchOne(CLIENTS, CLIENTS.NAME.eq(name));
+    return Optional.ofNullable(r).map(clientMapper::map);
+  }
+
+  public Optional<Client> getClientBySpiffeId(String spiffeId) {
+    ClientsRecord r = dslContext.fetchOne(CLIENTS, CLIENTS.SPIFFE_ID.eq(spiffeId));
     return Optional.ofNullable(r).map(clientMapper::map);
   }
 
--- a/server/src/main/java/keywhiz/service/daos/ClientDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/ClientDAO.java
@@ -34,8 +34,6 @@ import org.jooq.Configuration;
 import org.jooq.DSLContext;
 import org.jooq.Param;
 import org.jooq.impl.DSL;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.time.Instant.EPOCH;
@@ -45,7 +43,6 @@ import static org.jooq.impl.DSL.greatest;
 import static org.jooq.impl.DSL.when;
 
 public class ClientDAO {
-  private static final Logger logger = LoggerFactory.getLogger(ClientDAO.class);
   private final static Duration LAST_SEEN_THRESHOLD = Duration.ofSeconds(24 * 60 * 60);
 
   private final DSLContext dslContext;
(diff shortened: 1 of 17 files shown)


server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
```diff
@@ -0,0 +1,172 @@
...
+      String spiffeId) {
+    if (createMissingClient && !name.isEmpty()) {
+      /*
+       * If a client is seen for the first time, authenticated by certificate, and has no DB entry,
+       * then a DB entry is created here. The client can be disabled in the future by flipping the
+       * 'enabled' field.
+       */
+      long clientId = clientDAOReadWrite.createClient(name, "automatic",
+          "Client created automatically from valid certificate authentication", spiffeId);
+      return clientDAOReadWrite.getClientById(clientId);
```
jbpeirce, line 111: There's a small behavior change here: the original authenticator returned
`return Optional.of(clientDAOReadWrite.getClientById(clientId).get());`
which would throw a `NoSuchElementException` if the client couldn't be created. However, the only situations I can think of where the client creation fails would be if the client already existed (not the case here, barring race conditions) or when the database is down; in either case, cleanly denying authentication rather than throwing an exception is probably better.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -135,11 +136,16 @@ public void sawClient(Client client, @Nullable Principal principal) {
...
   }
 
-  public Optional<Client> getClient(String name) {
+  public Optional<Client> getClientByName(String name) {
     ClientsRecord r = dslContext.fetchOne(CLIENTS, CLIENTS.NAME.eq(name));
     return Optional.ofNullable(r).map(clientMapper::map);
   }
 
+  public Optional<Client> getClientBySpiffeId(String spiffeId) {
+    ClientsRecord r = dslContext.fetchOne(CLIENTS, CLIENTS.SPIFFE_ID.eq(spiffeId));
```
mbyczkowski, line 145: are SPIFFE IDs going to be unique?

mweissbacher, line 145: the trust domain part should be lower-cased here for comparison

server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
mbyczkowski: makes sense to me

server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
```diff
@@ -0,0 +1,172 @@
...
+
+    // The sub-lists returned by getSubjectAlternativeNames have an integer for the first
+    // entry, representing a field, and the value as a string as the second entry.
+    List<String> spiffeUriNames = sans.stream()
+        .filter(sanPair -> sanPair.get(0).equals(URINAME_SAN))
+        .map(sanPair -> (String) sanPair.get(1))
+        .filter(uri -> uri.startsWith(SPIFFE_SCHEME))
+        .collect(Collectors.toUnmodifiableList());
+
+    if (spiffeUriNames.size() > 1) {
```
mbyczkowski, line 165: good call on limiting to only one SPIFFE ID

server/src/main/resources/db/mysql/migration/V12__make_spiffe_id_unique.sql
```diff
@@ -0,0 +1,8 @@
+# Keywhiz does not currently support an entity acting on behalf of other clients without being
```
mbyczkowski, line 1: I think this answers my earlier question about SPIFFE ID uniqueness :)

server/src/main/java/keywhiz/service/daos/ClientDAO.java
jbpeirce: Good call--I'll make a commit to handle all SPIFFE IDs as URIs rather than strings (but I'll put it in a new PR so this one doesn't need to be re-reviewed).

mweissbacher: I don't think this is critical, you could argue it's the client's responsibility to acquire certs that are well formatted.

server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
mweissbacher: I think exiting if more than one is present is a good call, it breaks with the standard so probably best to not make assumptions.

PR #686, opened 2020-06-30
Share lookup code for regular and automation clients

jbpeirce:
"Automation" clients, which have significant power to manage
secrets in Keywhiz (due to access to additional APIs), are stored
in the same database table and authenticated in the same way as
normal Keywhiz clients. This means that their authentication code
can share logic with the standard client lookup process, except
for the final stage of attempting to use the retrieved client
as an automation client.

Note that although at this point the ClientAuthFactory and
AutomationClientAuthFactory have significant code differences
(because the ClientAuthFactory supports extracting client
information from the x-forwarded-client-cert header and looking
up clients based on their SPIFFE IDs), these differences are
because of recent changes to the ClientAuthFactory. Before the
ClientAuthFactory was updated, the only differences in behavior
between the two factories were the ForbiddenException thrown by
the AutomationClient factory and the AutomationClient factory not
creating a missing client, both of which are preserved in this commit.

### DIFF
--- a/server/src/test/java/keywhiz/service/providers/AutomationClientAuthFactoryTest.java
+++ b/server/src/test/java/keywhiz/service/providers/AutomationClientAuthFactoryTest.java
@@ -16,6 +16,7 @@
 
 package keywhiz.service.providers;
 
+import java.net.URI;
 import java.security.Principal;
 import java.util.Optional;
 import javax.ws.rs.ForbiddenException;
@@ -24,6 +25,8 @@ import javax.ws.rs.core.SecurityContext;
 import keywhiz.api.model.AutomationClient;
 import keywhiz.api.model.Client;
 import keywhiz.auth.mutualssl.SimplePrincipal;
+import keywhiz.service.config.ClientAuthConfig;
+import keywhiz.service.config.ClientAuthTypeConfig;
 import keywhiz.service.daos.ClientDAO;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.junit.Before;
@@ -33,6 +36,7 @@ import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
 
+import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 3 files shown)


jbpeirce:
Compare the current AutomationClientAuthFactory against the ClientAuthFactory from [252383bc5b3b0fbc522d9e6af11f6919cb4c83a9](https://github.com/square/keywhiz/blob/252383bc5b3b0fbc522d9e6af11f6919cb4c83a9/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java), before the XFCC-handling changes, to see the initial similarity between the two classes.

server/src/test/java/keywhiz/service/providers/AutomationClientAuthFactoryTest.java
```diff
@@ -78,19 +93,12 @@ public void automationClientRejectsClientsWithoutAutomation() {
     factory.provide(request);
   }
 
-  @Test(expected = ForbiddenException.class)
+  @Test(expected = NotAuthorizedException.class)
```
mbyczkowski, line 96: shouldn't we still test for ForbiddenException, assuming that the behavior is preserved?

jbpeirce: This exception did change from `ForbiddenException` to `NotAuthorizedException`. Originally retrieving a `Client` that didn't exist would have caused a `NotAuthorizedException` ([here](https://github.com/square/keywhiz/blob/252383bc5b3b0fbc522d9e6af11f6919cb4c83a9/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java#L66-L68)) while retrieving an `AutomationClient` that didn't exist threw a `ForbiddenException` ([here](https://github.com/square/keywhiz/blob/252383bc5b3b0fbc522d9e6af11f6919cb4c83a9/server/src/main/java/keywhiz/service/providers/AutomationClientAuthFactory.java#L60-L62)). Both systems threw a `NotAuthorizedException` when client information wasn't provided in the request.

I decided to standardize on `NotAuthorizedException` when there's an error retrieving the client, and `ForbiddenException` when the client is present but not an automation client.

PR #688, opened 2020-06-30
Adding batch secrets endpoint for SPIFFE clients

mweissbacher:
Tests will be added in another commit on this PR

Signed-off-by: Michael Weissbacher <mweissbacher@squareup.com>

### DIFF
--- a/api/src/main/java/keywhiz/api/BatchSecretRequest.java
+++ b/api/src/main/java/keywhiz/api/BatchSecretRequest.java
@@ -1,35 +1,23 @@
 package keywhiz.api;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.base.Objects;
+import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import org.hibernate.validator.constraints.NotEmpty;
+
+import javax.validation.constraints.NotNull;
 
 /**
  * Batch Secret Request.
  */
-public class BatchSecretRequest {
-  @NotEmpty
-  @JsonProperty
-  public final ImmutableList<String> secrets;
-
-  public BatchSecretRequest(@JsonProperty("secrets") ImmutableList<String> secrets) {
-    this.secrets = secrets;
+@AutoValue
+public abstract class BatchSecretRequest {
+  public static BatchSecretRequest create(ImmutableList<String> secrets) {
+    return new AutoValue_BatchSecretRequest(secrets);
   }
 
-  @Override public int hashCode() {
-    return Objects.hashCode(secrets);
-  }
+  @JsonProperty("secrets")
+  @NotNull
+  public abstract ImmutableList<String> secrets();
 
-  @Override public boolean equals(Object o) {
-    if (this == o) return true;
-    if (o instanceof BatchSecretRequest) {
-      BatchSecretRequest that = (BatchSecretRequest) o;
-      if (Objects.equal(this.secrets, that.secrets)) {
-        return true;
-      }
-    }
-    return false;
-  }
+
 }
--- a/api/src/main/java/keywhiz/api/BatchSecretRequest.java
+++ b/api/src/main/java/keywhiz/api/BatchSecretRequest.java
@@ -1,0 +1,35 @@
+package keywhiz.api;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.hibernate.validator.constraints.NotEmpty;
+
+/**
+ * Batch Secret Request.
+ */
+public class BatchSecretRequest {
+  @NotEmpty
+  @JsonProperty
+  public final ImmutableList<String> secrets;
+
+  public BatchSecretRequest(@JsonProperty("secrets") ImmutableList<String> secrets) {
+    this.secrets = secrets;
+  }
+
+  @Override public int hashCode() {
+    return Objects.hashCode(secrets);
+  }
+
+  @Override public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o instanceof BatchSecretRequest) {
+      BatchSecretRequest that = (BatchSecretRequest) o;
+      if (Objects.equal(this.secrets, that.secrets)) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
--- a/api/src/main/java/keywhiz/api/BatchSecretRequest.java
+++ b/api/src/main/java/keywhiz/api/BatchSecretRequest.java
@@ -1,10 +1,16 @@
 package keywhiz.api;
 
+import autovalue.shaded.org.checkerframework$.checker.nullness.qual.$MonotonicNonNull;
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.hibernate.validator.constraints.NotEmpty;
 
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
+import java.util.List;
 
 /**
  * Batch Secret Request.
@@ -15,9 +21,10 @@ public abstract class BatchSecretRequest {
     return new AutoValue_BatchSecretRequest(secrets);
   }
 
-  @JsonProperty("secrets")
-  @NotNull
-  public abstract ImmutableList<String> secrets();
+  @JsonProperty("secrets") public abstract ImmutableList<String> secrets();
 
+  @JsonCreator public static BatchSecretRequest forName(@JsonProperty("secrets") ImmutableList<String> secrets) {
+    return new AutoValue_BatchSecretRequest(secrets);
+  }
 
 }
--- a/api/src/main/java/keywhiz/api/BatchSecretRequest.java
+++ b/api/src/main/java/keywhiz/api/BatchSecretRequest.java
@@ -23,7 +23,7 @@ public abstract class BatchSecretRequest {
 
   @JsonProperty("secrets") public abstract ImmutableList<String> secrets();
 
-  @JsonCreator public static BatchSecretRequest forName(@JsonProperty("secrets") ImmutableList<String> secrets) {
+  @JsonCreator public static BatchSecretRequest forSecrets(@JsonProperty("secrets") ImmutableList<String> secrets) {
     return new AutoValue_BatchSecretRequest(secrets);
   }
 
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -333,6 +333,35 @@ public class AclDAO {
         .map(row -> processSanitizedSecretRow(row, client));
   }
 
+  public List<SanitizedSecret> getBatchSanitizedSecretsFor(Client client, List<String> secretNames) {
+    checkNotNull(client);
+    checkArgument(!secretNames.isEmpty());
+
+    SelectQuery<Record> query = dslContext.select(SECRETS.fields())
+            .from(SECRETS)
+            .join(ACCESSGRANTS).on(SECRETS.ID.eq(ACCESSGRANTS.SECRETID))
+            .join(MEMBERSHIPS).on(ACCESSGRANTS.GROUPID.eq(MEMBERSHIPS.GROUPID))
+            .join(CLIENTS).on(CLIENTS.ID.eq(MEMBERSHIPS.CLIENTID))
+            .join(SECRETS_CONTENT).on(SECRETS_CONTENT.ID.eq(SECRETS.CURRENT))
+            .where(CLIENTS.NAME.eq(client.getName())
+                    .and(SECRETS.CURRENT.isNotNull())
+                    .and(SECRETS.NAME.in(secretNames)))
+            .limit(1)
+            .getQuery();
+    query.addSelect(SECRETS_CONTENT.CONTENT_HMAC);
+    query.addSelect(SECRETS_CONTENT.CREATEDAT);
+    query.addSelect(SECRETS_CONTENT.CREATEDBY);
+    query.addSelect(SECRETS_CONTENT.METADATA);
+    query.addSelect(SECRETS_CONTENT.EXPIRY);
+    query.addSelect(ACCESSGRANTS.ROW_HMAC);
+    query.addSelect(MEMBERSHIPS.ROW_HMAC);
+    query.addSelect(MEMBERSHIPS.GROUPID);
+    query.addSelect(CLIENTS.ROW_HMAC);
+    query.addSelect(SECRETS.ROW_HMAC);
+
+    return query.fetch().map(row -> processSanitizedSecretRow(row, client));
+  }
+
   private SanitizedSecret processSanitizedSecretRow(Record row, Client client) {
     boolean rowHmacLog = config.getRowHmacCheck() == RowHmacCheck.DISABLED_BUT_LOG;
     boolean rowHmacFail = config.getRowHmacCheck() == RowHmacCheck.ENFORCED;
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -346,7 +346,6 @@ public class AclDAO {
             .where(CLIENTS.NAME.eq(client.getName())
                     .and(SECRETS.CURRENT.isNotNull())
                     .and(SECRETS.NAME.in(secretNames)))
-            .limit(1)
             .getQuery();
     query.addSelect(SECRETS_CONTENT.CONTENT_HMAC);
     query.addSelect(SECRETS_CONTENT.CREATEDAT);
--- a/server/src/main/java/keywhiz/service/daos/SecretDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretDAO.java
@@ -22,10 +22,8 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import java.time.Instant;
 import java.time.OffsetDateTime;
+import java.util.*;
 import java.util.AbstractMap.SimpleEntry;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.ws.rs.BadRequestException;
@@ -281,6 +279,43 @@ public class SecretDAO {
       return Optional.of(SecretSeriesAndContent.of(series.get(), secretContent.get()));
     }
     return Optional.empty();
+  }
+
+  /**
+   * @param names of secrets series to look up secrets by.
+   * @return Secrets matching input parameters.
+   */
+  public List<SecretSeriesAndContent> getSecretsByName(List<String> names) {
+    checkArgument(!names.isEmpty());
+
+    SecretContentDAO secretContentDAO = secretContentDAOFactory.using(dslContext.configuration());
+    SecretSeriesDAO secretSeriesDAO = secretSeriesDAOFactory.using(dslContext.configuration());
+
+    List<SecretSeries> multipleseries = secretSeriesDAO.getSecretsSeriesByName(names);
+
+
+    Iterator siter = multipleseries.iterator();
+
+    List<SecretSeriesAndContent> ret = new ArrayList<SecretSeriesAndContent>();
+
+    while (siter.hasNext()) {
+
+      SecretSeries series = (SecretSeries) siter.next();
+
+      if (series.currentVersion().isPresent()) {
+        long secretContentId = series.currentVersion().get();
+        Optional<SecretContent> secretContent =
+                secretContentDAO.getSecretContentById(secretContentId);
+        if (secretContent.isPresent()) {
+          ret.add(SecretSeriesAndContent.of(series, secretContent.get()));
+        } else {
+          // TODO: how to handle this?
+        }
+
+      }
+    }
+    return ret;
+
   }
 
   /**
--- a/server/src/main/java/keywhiz/service/daos/SecretDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretDAO.java
@@ -20,14 +20,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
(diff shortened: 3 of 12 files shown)


server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -283,6 +281,43 @@ public boolean setExpiration(String name, Instant expiration) {
...
+      SecretSeries series = (SecretSeries) siter.next();
+
+      if (series.currentVersion().isPresent()) {
+        long secretContentId = series.currentVersion().get();
+        Optional<SecretContent> secretContent =
+                secretContentDAO.getSecretContentById(secretContentId);
+        if (secretContent.isPresent()) {
+          ret.add(SecretSeriesAndContent.of(series, secretContent.get()));
+        } else {
+          // TODO: how to handle this?
```
jbpeirce, line 312: Since `getSecretsSeriesByName` requires `SECRETS.CURRENT.IsNotNull()`, this should be impossible, so it's probably fine to throw an exception here.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -185,6 +185,10 @@ public int setCurrentVersion(long secretId, long secretContentId, String updater
     return Optional.ofNullable(r).map(secretSeriesMapper::map);
   }
 
+  public List<SecretSeries> getSecretsSeriesByName(List<String> names) {
```
jbpeirce, line 188: nit: maybe name this `getMultipleSecretSeriesByName` to make the API difference clearer--I think it'll be easy to miss the extra `s` in `SecretsSeries` versus `SecretSeries`

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,156 @@
...
+   * responseMessage 500 Secret response could not be generated for given Secret
+   */
+  @Timed
+  @ExceptionMetered
+  @POST
+  @Consumes(APPLICATION_JSON)
+  public List<SecretDeliveryResponse> getBatchSecret(@Auth Client client, @Valid BatchSecretRequest request) {
+
+
+    List<SanitizedSecret> sanitizedSecrets = aclDAO.getBatchSanitizedSecretsFor(client, request.secrets);
```
jbpeirce, line 96: I'd rename these (and the booleans below) to something like `clientAccessibleSecrets` and `existingSecrets` to make the significance of a secret missing from one or the other list clearer.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,156 @@
...
+    Iterator reqsecretsiter = request.secrets.iterator();
+
+
+    // The request fails whenever a single secret is requested that is not accessible
+    // The client is responsible for only requesting secrets they have permission for
+    while (reqsecretsiter.hasNext()) {
+      String secretname = (String) reqsecretsiter.next();
+
+      // We iterate over both arrays since the order of the two arrays might not match
+      SanitizedSecret sanitizedsecret = null;
```
jbpeirce, line 111: Since we're just checking whether these secrets exist, this section can be simplified to 
```
boolean secretExists = secrets.stream().anyMatch(s -> s.getName().equals(secretname));
boolean sanitizedSecretExists = sanitizedSecrets.stream().anyMatch(s -> s.getName.equals(secretname));
```
(although as noted above, I'd suggest renaming the variables)

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,156 @@
...
+          if (clientExists) {
+            logger.info("Client {} requested unknown secret {}", client.getName(), secretname);
+          }
+          throw new NotFoundException();
+
+        }
+      }
+    }
+
+    logger.info("Client {} granted access to {}.", client.getName(), sanitizedSecrets);
```
jbpeirce, line 145: nit: maybe just log the secret names here, rather than the full SanitizedSecret data which includes all fields except the secret contents, for more readable log messages.

`sanitizedSecrets.stream().map(s -> s.getName()).collect(toList())` should do it

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,156 @@
...
+  private final SecretController secretController;
+  private final AclDAO aclDAO;
+  private final ClientDAO clientDAO;
+
+  @Inject
+  public BatchSecretDeliveryResource(@Readonly SecretController secretController,
+                                     AclDAOFactory aclDAOFactory, ClientDAOFactory clientDAOFactory) {
+    this.secretController = secretController;
+    this.aclDAO = aclDAOFactory.readonly();
+    this.clientDAO = clientDAOFactory.readwrite();
```
jbpeirce, line 66: nit: can this be `readonly()`?

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -22,10 +22,8 @@
 import com.google.common.collect.Lists;
 import java.time.Instant;
 import java.time.OffsetDateTime;
+import java.util.*;
```
jbpeirce, line 25: I generally prefer importing individual classes rather than the full package.

api/src/main/java/keywhiz/api/BatchSecretRequest.java
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.hibernate.validator.constraints.NotEmpty;
+
+/**
+ * Batch Secret Request.
+ */
+public class BatchSecretRequest {
```
jbpeirce, line 12: I'd recommend using [AutoValue](https://github.com/google/auto/blob/master/value/userguide/index.md) here; it handles `equals()` and `hashCode()` automatically and makes creating builders easy (though this is a simple enough class we probably don't need a builder yet). [AutomationSecretResponse](https://github.com/square/keywhiz/blob/master/api/src/main/java/keywhiz/api/AutomationSecretResponse.java) and [ClientDetailResponseV2](https://github.com/square/keywhiz/blob/master/api/src/main/java/keywhiz/api/automation/v2/ClientDetailResponseV2.java) are a couple of the Keywhiz classes that use it.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
mweissbacher: Good call!

server/src/main/java/keywhiz/service/daos/SecretDAO.java
mweissbacher: Agree, missed this. This must have been the IntelliJ auto-reorganizer

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -283,6 +281,43 @@ public boolean setExpiration(String name, Instant expiration) {
...
+    SecretSeriesDAO secretSeriesDAO = secretSeriesDAOFactory.using(dslContext.configuration());
+
+    List<SecretSeries> multipleseries = secretSeriesDAO.getSecretsSeriesByName(names);
+
+
+    Iterator siter = multipleseries.iterator();
+
+    List<SecretSeriesAndContent> ret = new ArrayList<SecretSeriesAndContent>();
+
+    while (siter.hasNext()) {
```
jbpeirce, line 301: Rather than using iterators (here and in `BatchSecretDeliveryResource`), I generally prefer a foreach loop:
```
for (SecretSeries series : multipleSeries) {
  ...
}
```
It's a little easier to read and avoids possible mistakes from forgetting to check `hasNext()`/not calling `next()`.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,138 @@
...
+import keywhiz.service.daos.AclDAO.AclDAOFactory;
+import keywhiz.service.daos.ClientDAO;
+import keywhiz.service.daos.ClientDAO.ClientDAOFactory;
+import keywhiz.service.daos.SecretController;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.validation.Valid;
+import javax.ws.rs.*;
```
jbpeirce, line 39: nit: remove import `*`

server/src/test/java/keywhiz/service/resources/BatchSecretDeliveryResourceIntegrationTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+
+import static keywhiz.testing.HttpClients.testUrl;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BatchSecretDeliveryResourceIntegrationTest {
```
jbpeirce, line 43: Good tests! A couple of notes:
- Please document why "General_Password" is accessible to this client, but "Hacking_Password" isn't (or just create new secrets for this test).
- Please add a test of retrieving multiple secrets which succeed, and multiple secrets where one succeeds and another fails (to make sure that the whole request fails in that case)

jbpeirce: I hadn't yet read the mocked test, which already tests multiple secrets which succeed :) Alternatively, to keep the integration test lightweight, we could add the test with multiple secrets where one succeeds and the other fails to the mocked test. With a new test for AclDAO to make sure that retrieving batched secrets works as expected, that should be sufficient. 

(Documentation of why these secrets are accessible/inaccessible would still be good, though.)

server/src/test/java/keywhiz/service/resources/BatchSecretDeliveryResourceTest.java
```diff
@@ -0,0 +1,163 @@
...
+                .thenReturn(List.of(sanitizedSecret, sanitizedSecret2));
+        when(secretController.getSecretsByName(secretnames))
+                .thenReturn(List.of(secret, secret2));
+
+        List<SecretDeliveryResponse> response = batchSecretDeliveryResource.getBatchSecret(client, req);
+
+        // The order doesn't matter
+        boolean matches = false;
+
+        if (response.equals(ImmutableList.of(SecretDeliveryResponse.fromSecret(secret), SecretDeliveryResponse.fromSecret(secret2)))) {
```
jbpeirce, line 108: `assertThat(list).containsExactlyInAnyOrder(item1, item2, item3...)` can simplify this

server/src/test/java/keywhiz/service/resources/BatchSecretDeliveryResourceTest.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        assertThat(matches).isTrue();
+    }
+
+    @Test(expected = NotFoundException.class)
+    public void returnsNotFoundWhenClientDoesNotExist() throws Exception {
+        when(aclDAO.getBatchSanitizedSecretsFor(client, ImmutableList.of(secret.getName()))).thenReturn(ImmutableList.<SanitizedSecret>of());
+        when(clientDAO.getClientByName(client.getName())).thenReturn(Optional.empty());
+        when(secretController.getSecretsByName(ImmutableList.of(secret.getName())))
```
jbpeirce, line 121: This test should be set up with an existing secret, but missing client--currently both the client and the secret are missing, so it's hard to tell which failure case is actually being triggered.

server/src/test/java/keywhiz/service/resources/BatchSecretDeliveryResourceTest.java
```diff
@@ -0,0 +1,163 @@
...
+        String name = secretBase64.getName();
+
+        when(aclDAO.getBatchSanitizedSecretsFor(client, ImmutableList.of(name)))
+                .thenReturn(ImmutableList.of(SanitizedSecret.fromSecret(secretBase64)));
+        when(secretController.getSecretsByName(ImmutableList.of(name)))
+                .thenReturn(ImmutableList.of(secretBase64));
+
+        List<SecretDeliveryResponse> response = batchSecretDeliveryResource.getBatchSecret(client, BatchSecretRequest.create(ImmutableList.of(secretBase64.getName())));
+        assertThat(response.get(0).getSecret()).isEqualTo(secretBase64.getSecret());
+        assertThat(response.size()).isEqualTo(1);
```
jbpeirce, line 160: nit: I like to assert the list size first and retrieve items from it second, so that if it's empty the assertion fails rather than the test throwing an `IndexOutOfBoundsException`.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -44,14 +35,22 @@
...
+
 import static com.google.common.base.Preconditions.checkNotNull;
 import static keywhiz.jooq.tables.Accessgrants.ACCESSGRANTS;
 import static keywhiz.jooq.tables.Groups.GROUPS;
 import static keywhiz.jooq.tables.Secrets.SECRETS;
 import static keywhiz.jooq.tables.SecretsContent.SECRETS_CONTENT;
-import static org.jooq.impl.DSL.decode;
-import static org.jooq.impl.DSL.least;
-import static org.jooq.impl.DSL.val;
+import static org.jooq.impl.DSL.*;
```
jbpeirce, line 53: nit: remove import `*`

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -185,6 +184,10 @@ public int setCurrentVersion(long secretId, long secretContentId, String updater
     return Optional.ofNullable(r).map(secretSeriesMapper::map);
   }
 
+  public List<SecretSeries> getMultipleSecretSeriesByName(List<String> names) {
```
jbpeirce, line 187: Probably worth adding a small test for this (the query is simple enough I don't expect there to be a bug, but adding a test is easy).

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,150 @@
...
+
+    List<String> forbiddenSecrets = new ArrayList<>();
+
+    // The request fails whenever a single secret is requested that is not accessible
+    // The client is responsible for only requesting secrets they have permission for
+    for (String secretname : request.secrets()) {
+      boolean secretExists = existingSecrets.stream().anyMatch(s -> s.getName().equals(secretname));
+      boolean secretAccessible = clientAccessibleSecrets.stream().anyMatch(s -> s.name().equals(secretname));
+
+      if (!secretAccessible) {
```
jbpeirce, line 115: nit: We can simplify this section to make it easier to read:
```
if (!(clientExists && secretExists)) {
  log("Client {} or secret {} does not exist (client exists={}, secret exists={})", client.getName(), secretname, clientExists, secretExists);
  throw new NotFoundException();
}

if (!secretAccessible) {
  // at this point we know the client and secret both exist
  throw new ForbiddenException(format("Access denied: %s to secret '%s'", client.getName(), secretName));
}
```
If multiple secrets are forbidden, only one of them will get logged, but I think the clearer logic is worth it.

jbpeirce: This slightly obscures the case of "a secret shows up in the ACLs for a client without actually existing in the `secrets` table" (since it gets rolled into the "missing-client-or-secret" case), but that error should not be possible with a real database so I'm okay with not having a separate error case for it.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    // If *any* of the secrets is forbidden
+    if ( ! forbiddenSecrets.isEmpty() ) {
+      throw new ForbiddenException(format("Access denied: %s to secret(s) '%s'", client.getName(), forbiddenSecrets));
+    }
+
+    logger.info("Client {} granted access to {}.", client.getName(), clientAccessibleSecrets.stream().map(s -> s.name()).collect(toList()));
+    try {
+      return existingSecrets.stream()
```
jbpeirce, line 141: nit: It could be worth either explicitly returning `clientAccessibleSecrets` or adding a comment here to clarify that we can't reach this code unless `clientAccessibleSecrets` and `existingSecrets` are identical--I initially thought this could return existing secrets that the client didn't have access to.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -263,4 +263,54 @@ private void assertListDoesNotContainSecretsWithIds(List<SecretSeries> secrets,
   private int tableSize() {
     return jooqContext.fetchCount(SECRETS);
   }
+
+  @Test public void getMultipleSecretSeriesByNameReturnsOne() {
```
jbpeirce, line 267: nit: could we add a test that requesting two different secrets returns both secrets? (or update this test)

server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -333,6 +333,34 @@ public void findAndEvictClient(long clientId, long groupId, AuditLog auditLog, S
         .map(row -> processSanitizedSecretRow(row, client));
   }
 
+  public List<SanitizedSecret> getBatchSanitizedSecretsFor(Client client, List<String> secretNames) {
```
jbpeirce, line 336: nit: It might be worth adding a test specifically for this endpoint, although the BatchSecretsIntegrationTest implicitly tests it.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
mweissbacher: We can do that, but it would remove determinism. The advantage of the more complex code is that regardless of the input order it always behaves the same. Either works for me though.

jbpeirce: You're right, I'd overlooked determinism with multiple secrets. In that case, I'd suggest
```
if (!(clientExists && secretExists)) {
  log("Client {} or secret {} does not exist (client exists={}, secret exists={})", client.getName(), secretname, clientExists, secretExists);
  throw new NotFoundException();
}

if (!secretAccessible) {
  // at this point we know the client and secret both exist
  forbiddenSecrets.add(secretname);
}
```
It still simplifies the `if/else` structure considerably.

mweissbacher: That is indeed nicer!

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
mweissbacher: I added a comment. Summary offline discussion: `clientAccessibleSecrets`are SanitizedSecrets, that change wouldn't be possible.

jbpeirce:
This looks great--thank you!

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -283,6 +286,35 @@ public boolean setExpiration(String name, Instant expiration) {
...
+
+    List<SecretSeries> multipleSeries = secretSeriesDAO.getMultipleSecretSeriesByName(names);
+
+    List<SecretSeriesAndContent> ret = new ArrayList<SecretSeriesAndContent>();
+
+    for (SecretSeries series : multipleSeries) {
+      if (series.currentVersion().isPresent()) {
+        long secretContentId = series.currentVersion().get();
+        Optional<SecretContent> secretContent =
+                secretContentDAO.getSecretContentById(secretContentId);
```
mbyczkowski, line 307: wouldn't it make sense to also make this a batch query? (so `getSecretContentsByIds`?)

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -283,6 +286,35 @@ public boolean setExpiration(String name, Instant expiration) {
     return Optional.empty();
   }
 
+  /**
+   * @param names of secrets series to look up secrets by.
+   * @return Secrets matching input parameters.
+   */
+  public List<SecretSeriesAndContent> getSecretsByName(List<String> names) {
```
mbyczkowski, line 293: I've seen in other places we had API such that it would return `Optional<ImmutableList>`. I'm not sure if that's better, but there's a question about overall API consistency here.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -283,6 +286,35 @@ public boolean setExpiration(String name, Instant expiration) {
...
+
+    for (SecretSeries series : multipleSeries) {
+      if (series.currentVersion().isPresent()) {
+        long secretContentId = series.currentVersion().get();
+        Optional<SecretContent> secretContent =
+                secretContentDAO.getSecretContentById(secretContentId);
+        if (secretContent.isPresent()) {
+          ret.add(SecretSeriesAndContent.of(series, secretContent.get()));
+        } else {
+          throw new NotFoundException("Secret not found.");
```
mbyczkowski, line 311: If we have batch endpoint, do we want to fail if even one secret is not found or do we want to just return what's been found? How would an endpoint sending a non-existent secret name know how to drop an offending name and carry on?

mweissbacher: We've had lengthy discussions around whether we should keep this best-effort and ended on no. A.) It is possible for the client to describe secrets and see which they can access, this only becomes an issue if a secret changes right after a describe B.) the error message informs the client which secret was causing the error. A reason for this is keeping things simple - best-effort would have required being able to respond with different error messages per secret - that would have made the parsing on the client side more complicated.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
mweissbacher: There's a couple of things I'd like to make more consistent, especially around testing. We should earmark brushing up this code for another hackweek.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
mweissbacher: Yes, I'll make myself a ticket to implement this later. Good call!

PR #689, opened 2020-06-30
Use HttpServletRequest to check XFCC port

jbpeirce:
The ContainerRequest constructed by Jersey does not include information
about where the request was received by Keywhiz; the port in its URI
is derived from the port (if any) specified by the original requestor.
The HttpServletRequest, however, identifies which Keywhiz port traffic
came to, which allows us to enforce controls--notably whether certificates
are allowed to be set by the x-forwarded-client-cert header--based on
the port associated with incoming traffic.

There's a merge conflict between this change and https://github.com/square/keywhiz/pull/686, since the AutomationClientAuthFactory creation and use will need to be updated to include the ServletRequest as well. However, the fix will be straightforward (I'll rebase and update whichever PR is merged second to address it).

### DIFF
--- a/server/src/main/java/keywhiz/service/providers/AuthResolver.java
+++ b/server/src/main/java/keywhiz/service/providers/AuthResolver.java
@@ -19,6 +19,8 @@ package keywhiz.service.providers;
 import io.dropwizard.auth.Auth;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.core.Context;
 import keywhiz.api.model.AutomationClient;
 import keywhiz.api.model.Client;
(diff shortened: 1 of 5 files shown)


jbpeirce:
Rebased and updated after merging https://github.com/square/keywhiz/pull/686.

server/src/main/java/keywhiz/service/providers/AuthResolver.java
```diff
@@ -19,6 +19,8 @@
 import io.dropwizard.auth.Auth;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.core.Context;
```
klyubin, line 23: Looks like an unused import?

PR #697, opened 2020-07-07
Fix handling of SPIFFE SVIDs with multiple URIs, other adjustments

jbpeirce:
Removes an unnecessary comment, renames some variables, and (most
importantly) updates the handling of certificates that look like
SPIFFE SVIDs to reject certificates with a SPIFFE-format URI but
other URIs present as well.

### DIFF
--- a/server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
+++ b/server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
@@ -160,28 +160,41 @@ public class ClientAuthenticator {
 
     // The sub-lists returned by getSubjectAlternativeNames have an integer for the first
     // entry, representing a field, and the value as a string as the second entry.
-    List<String> spiffeUriNames = sans.stream()
+    List<String> providedUris = sans.stream()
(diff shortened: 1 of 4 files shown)


server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
```diff
@@ -160,28 +160,41 @@ public ClientAuthenticator(
...
+    // https://github.com/spiffe/spiffe/blob/master/standards/X509-SVID.md#2-spiffe-id
+    // > An X.509 SVID MUST contain exactly one URI SAN.
+    List<String> spiffeUriNames = providedUris.stream()
         .filter(uri -> uri.startsWith(SPIFFE_SCHEME))
         .collect(Collectors.toUnmodifiableList());
 
     if (spiffeUriNames.size() > 1) {
       logger.warn("Got multiple SPIFFE URIs from certificate: {}", spiffeUriNames);
       return Optional.empty();
+    } else if (spiffeUriNames.size() == 1 && providedUris.size() > 1) {
```
mbyczkowski, line 180: It would be good to have a test for a cert that has one SPIFFE ID and multiple URI SANs

PR #699, opened 2020-07-08
WIP upgrading dropwizard

jbpeirce:
Unfortunately, it appears that Jersey's stricter injection
rules (which prevent using @Context for a class registered as an
instance of a class rather than by the class name) are
incompatible with the injections needed by an AbstractValueParamProvider.

The refactor of AuthResolver is based on https://github.com/dropwizard/dropwizard/pull/2395/files and http://hoardingdata.blogspot.com/2019/02/jersey-226-and-injection-changes.html.

The problem with injecting @Context is documented in https://www.dropwizard.io/en/latest/manual/upgrade-notes/upgrade-notes-2_0_x.html#upgrade-notes-dropwizard-2-0-x.

I'm not sure how much longer I'll be able to spend on this in the near future.

### DIFF
--- a/server/src/main/java/keywhiz/auth/cookie/AuthenticatedEncryptedCookieFactory.java
+++ b/server/src/main/java/keywhiz/auth/cookie/AuthenticatedEncryptedCookieFactory.java
@@ -26,19 +26,18 @@ import javax.crypto.AEADBadTagException;
 import javax.inject.Inject;
 import javax.ws.rs.core.NewCookie;
 import keywhiz.auth.User;
-import org.eclipse.jetty.http.HttpCookie;
-import org.eclipse.jetty.http.HttpHeader;
-import org.eclipse.jetty.server.HttpChannel;
-import org.eclipse.jetty.server.HttpConfiguration;
-import org.eclipse.jetty.server.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


server/src/main/java/keywhiz/auth/cookie/AuthenticatedEncryptedCookieFactory.java
```diff
@@ -118,15 +118,7 @@ public NewCookie getSessionCookie(User user, ZonedDateTime expiration) {
...
-    HttpCookie cookie = new HttpCookie(config.getName(), "expired", config.getDomain(), config.getPath(),
-        0, config.isHttpOnly(), config.isSecure());
-
-    Response response = newResponse();
-    response.addCookie(cookie);
-    return NewCookie.valueOf(response.getHttpFields().get(HttpHeader.SET_COOKIE));
-  }
-
-  private Response newResponse() {
-    return new Response(new HttpChannel(null, new HttpConfiguration(), null, null), null);
```
jbpeirce, line 130: Creating an HttpChannel with a null Connector is not supported after the Jetty upgrade in Dropwizard, but it appears that the only purpose of this Response was to convert HttpCookie to NewCookie.

jbpeirce:
Closing this as I'm no longer working closely with Keywhiz, and the upgrade process will likely require different steps at this point.

PR #705, opened 2020-07-13
Customize MySQL port with KEYWHIZ_CUSTOM_MYSQL_PORT environment variable

jbpeirce:
This updates Keywhiz' model and tests to allow customizing what
port is used for its connection to MySQL by setting the
KEYWHIZ_CUSTOM_MYSQL_PORT environment variable. This is useful
when the default MySQL instance is unsuitable for building and
testing Keywhiz (for instance, if the MySQL server installed
on the default port 3306 runs MySQL 5.6, since Keywhiz now uses
MySQL 5.7).

If the environment variable is not set, Keywhiz will use the
default MySQL port as specified in the "mysql-connector"
documentation (currently 3306, or 33060 for a connection using
the X protocol).

See https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-jdbc-url-format.html#connector-j-url-single-host-without-props for the new format for JDBC connections and https://maven.apache.org/guides/introduction/introduction-to-profiles.html for the custom profiles used in keywhiz-model.


jbpeirce:
Dropwizard does support environment variable interpolation in config files, but appears to only allow this if the environment variable is the only entry in that field: https://www.dropwizard.io/en/latest/manual/core.html#environment-variables

PR #706, opened 2020-07-16
Generate Jooq code without requiring a database

jbpeirce:
Jooq's default code generation plugin works by connecting to a database, examining the schema for the database, and reverse engineering the required classes based on that schema. This requires having a database present and running for the code generation step.

This can be avoided using Jooq's [DDLDatabase](https://www.jooq.org/doc/3.13/manual/code-generation/codegen-ddl/). This creates an H2 database in memory, applies the specified migrations to that database, and then examines the database's schema and reverse engineers the required classes. Although this consumes more memory during the build process, it decouples code generation from the presence of a running database. 

Unfortunately, making the migrations compatible with Jooq's mysql-to-H2 conversion required adjusting some details of past migrations, which will alter their checksums as applied by Flyway.

After this PR, specifying `-Dmaven.test.skip=true` removes Keywhiz' requirement for a running database.


jbpeirce:
Closing this as I'm no longer working closely with Keywhiz, and updates to our internal infrastructure made this unnecessary.

PR #720, opened 2020-08-03
Configure whether to create clients when missing

jbpeirce:
Keywhiz supports the ability to automatically create clients
(without assigning them to any groups or secrets) when a new
client certificate is presented. This has previously been
automatic, but is now configurable. WARNING: If the configuration
is not specified, it defaults to "false."

### DIFF
--- a/server/src/main/java/keywhiz/service/config/ClientAuthConfig.java
+++ b/server/src/main/java/keywhiz/service/config/ClientAuthConfig.java
@@ -28,13 +28,23 @@ import java.util.List;
 public abstract class ClientAuthConfig {
   @JsonCreator public static ClientAuthConfig of(
       @JsonProperty("xfcc") List<XfccSourceConfig> sourceConfigs,
-      @JsonProperty("type") ClientAuthTypeConfig typeConfig) {
-    return new AutoValue_ClientAuthConfig(sourceConfigs, typeConfig);
+      @JsonProperty("type") ClientAuthTypeConfig typeConfig,
+      @JsonProperty("createMissingClients") boolean createMissingClients) {
+    return new AutoValue_ClientAuthConfig(sourceConfigs, typeConfig, createMissingClients);
   }
 
-  /** connection sources that can set x-forwarded-client-cert headers */
+  /**
+   * connection sources that can set x-forwarded-client-cert headers
+   */
   public abstract List<XfccSourceConfig> xfccConfigs();
 
-  /** what identifier(s) to use for clients */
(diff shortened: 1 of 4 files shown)


server/src/main/java/keywhiz/service/config/ClientAuthConfig.java
```diff
@@ -28,13 +28,23 @@
 public abstract class ClientAuthConfig {
   @JsonCreator public static ClientAuthConfig of(
       @JsonProperty("xfcc") List<XfccSourceConfig> sourceConfigs,
-      @JsonProperty("type") ClientAuthTypeConfig typeConfig) {
-    return new AutoValue_ClientAuthConfig(sourceConfigs, typeConfig);
+      @JsonProperty("type") ClientAuthTypeConfig typeConfig,
+      @JsonProperty("createMissingClients") boolean createMissingClients) {
```
jbpeirce, line 32: Unfortunately, Jackson does not support providing a default value for missing properties--although the [JsonProperty](https://fasterxml.github.io/jackson-annotations/javadoc/2.5/com/fasterxml/jackson/annotation/JsonProperty.html) annotation has a "defaultValue" field, this is **only for documentation** and does not affect how an empty field is interpreted.

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -293,7 +293,7 @@ private boolean containsUri(List<String> uriList, URI targetUri) {
    * @return whether to create a client that can't be found, if the client doesn't exist
    */
   protected boolean createMissingClient() {
-    return true;
+    return clientAuthConfig.createMissingClients();
```
isemaya-square, line 296: Is it necessary to have both `createMissingClients()` and `createMissingClient()`? Since createMissingClient() ends up calling `createMissingClients()`anyway.

jbpeirce: Yes; the `createMissingClient()` method gets overridden by the `AutomationClientAuthFactory`, because that factory never creates clients ([here](https://github.com/square/keywhiz/blob/2268480482fdfbf7389aded803605b01ebc3d583/server/src/main/java/keywhiz/service/providers/AutomationClientAuthFactory.java#L64-L68))

isemaya-square: Thanks for clarifying!

PR #731, opened 2020-08-25
Improve handling of empty secret contents on update

jbpeirce:
When empty secret contents are passed to the keywhiz.cli
command-line tool, it silently does not update the secret's
contents. This updates the command-line tool to error when
empty contents are passed but the --content flag is specified.

This also updates Keywhiz' internal update endpoint to check
for empty contents, since that endpoint is not accessed
through the SecretController's Builder class (unlike the create
and complete-update endpoints, which use SecretController's
validation to ensure that secret contents are present).

### DIFF
--- a/cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
+++ b/cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
@@ -58,6 +58,9 @@ public class UpdateAction implements Runnable {
     byte[] content = {};
     if (config.contentProvided) {
       content = readSecretContent();
+      if (content.length == 0) {
+        throw new IllegalArgumentException("Secret content must be provided if --content flag is present; check inputs");
+      }
     }
     partialUpdateSecret(secretName, content, config);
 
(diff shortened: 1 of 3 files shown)


cli/src/main/java/keywhiz/cli/commands/UpdateAction.java
```diff
@@ -58,6 +58,9 @@ public UpdateAction(UpdateActionConfig config, KeywhizClient client,
     byte[] content = {};
     if (config.contentProvided) {
       content = readSecretContent();
+      if (content.length == 0) {
+        throw new IllegalArgumentException("Secret content must be provided if --content flag is present; check inputs");
+      }
```
jbpeirce, line 63: Before this change, `partialUpdateSecret` would tell the server that content had _not_ been provided if the content was empty, so the contents would silently not be reset (but they would not be cleared, even if the `--content` flag had been set).
```
keywhizClient.updateSecret(secretName, config.description != null,
          config.getDescription(), content.length > 0, content, config.json != null,
          config.getMetadata(mapper), config.expiry != null, config.getExpiry());
```

wtmcc (approved):

mbyczkowski (approved):

PR #757, opened 2020-10-16
Added endpoints to undelete a secret - server-side updates only

isemaya-square:
In order to undelete a secret, there are 4 endpoints that will be called - 
these are the methods added for each endpoint:
1) **findDeletedSecretsByName** - this returns all deleted secrets with the
inputted name (before they were deleted) - the user will have to choose
secret to undelete amongst these
2) **getSecretsContentsBySecretId** - this lists all the secrets contents
that exist assocaited with the given secret id - the user will have to
choose which one to set to the deleted secret's current
3) **renameSecret** - if there is already a secret with the same name as the
secret that is to be deleted, the deleted secret will need to be renamed
4) **updateSecretsCurrent** - this changes the secret's current from null to
the given current id

The reason why multiple endpoints are needed to implement undeleting a
secret (rather than an "undelete" endpoint) is because the user might need 
to choose one secret and one secrets content from multiple secrets and 
multiple secrets contents, respectively.

Note that this PR doesn't have any updates for the cli


server/pom.xml
```diff
@@ -201,6 +201,21 @@
...
+    <dependency>
+      <groupId>org.jooq</groupId>
+      <artifactId>jooq</artifactId>
+      <version>3.13.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jooq</groupId>
+      <artifactId>jooq</artifactId>
+      <version>3.13.5</version>
+    </dependency>
```
jbpeirce, line 218: There should be at most one dependency for a given `groupId` and `artifactId` in the same file, and rather than specifying a version it should inherit the version specified in the top-level POM--for instance, dropwizard in this file has
```
    <dependency>
      <groupId>io.dropwizard</groupId>
      <artifactId>dropwizard-auth</artifactId>
    </dependency>
```
so that it uses the dropwizard version from the top-level file. (The httpcore import should probably also be updated to do the same thing.)

I'm surprised that this dependency would be newly required, since the server code already used Jooq. Is it possible that this isn't necessary, and there was an issue with the devenv?

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -449,6 +489,29 @@ public void deleteSecretsByName(String name) {
         .deleteSecretSeriesByName(name);
   }
 
+  /**
+   * Renames the secret, specified by the secret id, to the name provided
+   * Assumes that there are no secrets with the name provided - i.e. that getSecretsByName
```
jbpeirce, line 494: nit: I'm not sure how to interpret this comment--it actually wouldn't be safe to assume there are no secrets with the name provided, since there could be a race between something else creating the new secret and our call to rename even if we'd called `getSecretsByName` just before calling `renameSecretById`. However, in practice the `UNIQUE` index on secret names means that if there is such a race, one of the two will fail. 

It is good to document that this will throw an exception if the name exists, though, and to suggest calling `getSecretsByName` or something similar to reduce the likelihood of the exception.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -449,6 +489,29 @@ public void deleteSecretsByName(String name) {
...
+   */
+  public void renameSecretById(long secretId, String name) {
+    checkArgument(!name.isEmpty());
+
+    secretSeriesDAOFactory.using(dslContext.configuration())
+        .renameSecretSeriesById(secretId, name);
+  }
+
+  /**
+   * Deletes the series and all associated version of the given secret series name.
```
jbpeirce, line 507: nit: incorrect comment

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -306,6 +317,20 @@ public void deleteSecretSeriesById(long id) {
     });
   }
 
+  public void renameSecretSeriesById(long secretId, String name) {
```
jbpeirce, line 320: This should set the `updatedAt` and `updatedBy` fields as well.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -306,6 +317,20 @@ public void deleteSecretSeriesById(long id) {
...
+    dslContext.update(SECRETS)
+        .set(SECRETS.NAME, name)
+        .where(SECRETS.ID.eq(secretId))
+        .execute();
+  }
+
+  public void updateSecretSeriesContentById(long secretId, long secretContentId) {
```
jbpeirce, line 327: The existing `setCurrentVersion` endpoint will actually handle this already, with some additional checks (for instance, making sure that this `secretContentId` was linked to this `secretId`).

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -52,7 +52,7 @@
 
 /**
  * parentEndpointName secret
- * <p>
+ * <p>unless
```
jbpeirce, line 55: nit: typo?

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -487,6 +490,135 @@ public Response deleteSecret(@Auth User user, @PathParam("secretId") LongParam s
...
+  }
+
+  /**
+   * Rename Secret by ID to the given name
+   *
+   * @param user     the admin user performing this operation
+   * @param secretId the ID of the Secret to be deleted
+   * @return 200 if secret deleted, 404 if not found
+   * <p>
+   * description Deletes a single Secret if found. Used by Keywhiz CLI and the web ui.
```
jbpeirce, line 523: nit: update comment

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -487,6 +490,135 @@ public Response deleteSecret(@Auth User user, @PathParam("secretId") LongParam s
...
+  @Path("rename/{secretId}/{secretName}")
+  @Timed @ExceptionMetered
+  @POST
+  public Response renameSecret(@Auth User user, @PathParam("secretId") LongParam secretId,
+      @PathParam("secretName") String secretName) {
+    Optional<Secret> secret = secretController.getSecretByName(secretName);
+    if (secret.isPresent()) {
+      logger.info("User '{}' tried renaming a secret, but another secret with that name "
+              + "already exists (name={})", user, secretId.get());
+      throw new NotAllowedException("That name is already taken by another secret"); //use different error
```
jbpeirce, line 538: a `ConflictException`, which this UI throws when trying to create a secret whose name already exists, could make sense :)

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -487,6 +490,135 @@ public Response deleteSecret(@Auth User user, @PathParam("secretId") LongParam s
...
+    }
+
+    logger.info("User '{}' renamed secret id={} to name='{}'", user, secretId,
+        secretName);
+
+    secretDAOReadWrite.renameSecretById(secretId.get(), secretName);
```
jbpeirce, line 544: This will throw a `DataAccessException` if the secret name already exists; we should catch it and throw an exception here as well.

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -487,6 +490,135 @@ public Response deleteSecret(@Auth User user, @PathParam("secretId") LongParam s
...
+   * responseMessage 200 Found and deleted Secret with given ID
+   * <p>
+   * responseMessage 404 Secret with given ID not Found
+   */
+  @Path("secretcontents/update/{secretsContentId}/{secretId}")
+  @Timed @ExceptionMetered
+  @POST
+  public Response updateSecretsCurrent(@Auth User user, @PathParam("secretId") LongParam secretId,
+      @PathParam("secretsContentId") LongParam secretsContentId) {
+    Optional<Secret> secret = secretController.getSecretById(secretId.get());
```
jbpeirce, line 603: Is the plan for this endpoint to be used to restore deleted secrets? If so, we should actually remove this check because `getSecretById` works only for non-deleted secrets.

server/src/test/java/keywhiz/service/daos/SecretDAOTest.java
```diff
@@ -526,6 +526,148 @@ public void getSecretByIdOneThrowsExceptionIfCurrentVersionIsInvalid() {
     assertThat(versions.get().size()).isEqualTo(1);
   }
 
+  @Test public void undeleteSecret() {
```
jbpeirce, line 529: Nice tests!

server/src/test/java/keywhiz/service/daos/SecretDAOTest.java
```diff
@@ -526,6 +526,148 @@ public void getSecretByIdOneThrowsExceptionIfCurrentVersionIsInvalid() {
...
+    Optional<SecretSeriesAndContent> secret = secretDAO.getSecretById(secretId);
+    assertThat(secret.get().series().name()).isEqualTo("toBeRenamed_renameSecretByIdName");
+
+    secretDAO.renameSecretById(secretId, "newName");
+
+    secret = secretDAO.getSecretById(secretId);
+    assertThat(secret).isPresent();
+    assertThat(secret.get().series().name()).isEqualTo("newName");
+  }
+
```
jbpeirce, line 670: Let's also double-check that renaming fails if a secret with the same name already exists.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -168,6 +168,73 @@ public void setCurrentVersion_failsWithIncorrectSecretContent() {
...
+        secretSeriesDAO.getSecretSeriesByDeletedName("toBeFound_getSecretSeriesByDeletedName");
+    assertThat(deletedSecretSeries.size()).isEqualTo(1);
+    assertThat(deletedSecretSeries.get(0).name()).contains("toBeFound_getSecretSeriesByDeletedName");
+    assertThat(deletedSecretSeries.get(0).name()).isNotEqualTo("toBeFound_getSecretSeriesByDeletedName");
+    assertThat(deletedSecretSeries.get(0).id()).isEqualTo(id);
+  }
+
+  @Test public void getDeletedSecretSeriesById() {
+    long now = OffsetDateTime.now().toEpochSecond();
+    long id = secretSeriesDAO.createSecretSeries("toBeFound_getSecretSeriesByDeletedName",
```
jbpeirce, line 223: nit: the secret should probably be called `toBeFound_getSecretSeriesByDeletedId` so that it has a unique name compared to the other new test. Unfortunately, the table is not reset between tests, so duplicate secret names can cause unexpected behavior.

jbpeirce (approved):

jbpeirce:
Looking good--thanks for the updates!

PR #792, opened 2021-01-22
Add a custom header to parse the caller Spiffe Id from

KenjiChao:
Add an optional custom header identifies the "caller" who sent the
original request. If present, we attempt to parse the client Spiffe
Id from this header instead of the XFCC header.

It is to support a scenario where a real client is behind a proxy.
The Spiffe Id extracted from the XFCC header may or may not match
the one set in this custom header. For example, considering a
request flow like this: client -> proxy -> Envoy -> Keywhiz. In
this case, the XFCC header set by the envoy instance contains the
proxy cert, while the custom Spiffe Id header contains the client
information behind the proxy.

### DIFF
--- a/server/src/main/java/keywhiz/auth/mutualssl/SpiffePrincipal.java
+++ b/server/src/main/java/keywhiz/auth/mutualssl/SpiffePrincipal.java
@@ -1,0 +1,29 @@
+package keywhiz.auth.mutualssl;
+
+import java.net.URI;
+import java.security.Principal;
+
+public class SpiffePrincipal implements Principal {
+  private final URI spiffeId;
+
+  public SpiffePrincipal(URI spiffeId) {
+    this.spiffeId = spiffeId;
+  }
+
+  @Override public String getName() {
+    return spiffeId.toString();
+  }
+
+  public URI getSpiffeId() {
+    return spiffeId;
+  }
+
+  /**
+   * Use the workload id of a Spiffe Id as the client name.
+   */
+  public String getClientName() {
+    String path = spiffeId.getPath();
+    // Drop the leading '/' character.
+    return path.isEmpty() ? path : path.substring(1);
+  }
+}
--- a/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
+++ b/server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
@@ -35,6 +35,7 @@ import javax.ws.rs.NotAuthorizedException;
 import keywhiz.KeywhizConfig;
 import keywhiz.api.model.Client;
 import keywhiz.auth.mutualssl.CertificatePrincipal;
+import keywhiz.auth.mutualssl.SpiffePrincipal;
 import keywhiz.service.config.ClientAuthConfig;
 import keywhiz.service.config.XfccSourceConfig;
 import keywhiz.service.daos.ClientDAO;
@@ -107,10 +108,10 @@ public class ClientAuthFactory {
     // on the security context of this request
     if (possibleXfccConfig.isEmpty()) {
       // The XFCC header is not used; use the security context of this request to identify the client
-      return authenticateClientFromCertificate(requestPrincipal);
+      return authenticateClientFromPrincipal(requestPrincipal);
     } else {
       return authorizeClientFromXfccHeader(possibleXfccConfig.get(), xfccHeaderValues,
-          requestPrincipal);
+          requestPrincipal, containerRequest);
     }
   }
 
@@ -134,11 +135,33 @@ public class ClientAuthFactory {
   }
 
   private Client authorizeClientFromXfccHeader(XfccSourceConfig xfccConfig,
-      List<String> xfccHeaderValues, Principal requestPrincipal) {
+      List<String> xfccHeaderValues, Principal requestPrincipal,
+      ContainerRequest containerRequest) {
     // Do not allow the XFCC header to be set by all incoming traffic. This throws a
     // NotAuthorizedException when the traffic is not coming from a source allowed to set the
     // header.
     validateXfccHeaderAllowed(xfccConfig, requestPrincipal);
+
+    Optional<String> callerSpiffeIdHeader = Optional.ofNullable(xfccConfig.callerSpiffeIdHeader());
+    List<String> callerSpiffeIdList = callerSpiffeIdHeader.map(
+        header -> Optional.ofNullable(containerRequest.getRequestHeader(header))
+            .orElse(List.of()))
+        .orElse(List.of());
+    int size = callerSpiffeIdList.size();
+
+    Optional<URI> callerSpiffeId = callerSpiffeIdHeader.flatMap(
+        header -> ClientAuthenticator.getSpiffeIdFromHeader(containerRequest, header));
+
+    if (size > 1 || size == 1 && callerSpiffeId.isEmpty()) {
+      throw new NotAuthorizedException(format(
+          "Invalid caller Spiffe Id header. It should contain only one URI and follow Spiffe Id format. size: %d, header: %s",
+          size, callerSpiffeIdList));
+    }
+
+    if (callerSpiffeId.isPresent()) {
+      SpiffePrincipal spiffePrincipal = new SpiffePrincipal(callerSpiffeId.get());
+      return authenticateClientFromPrincipal(spiffePrincipal);
+    }
 
     // Extract client information from the XFCC header
     X509Certificate clientCert =
@@ -149,9 +172,9 @@ public class ClientAuthFactory {
 
     CertificatePrincipal certificatePrincipal =
         new CertificatePrincipal(clientCert.getSubjectDN().toString(),
-            new X509Certificate[] {clientCert});
+            new X509Certificate[] { clientCert });
 
-    return authenticateClientFromCertificate(certificatePrincipal);
+    return authenticateClientFromPrincipal(certificatePrincipal);
   }
 
   private void validateXfccHeaderAllowed(XfccSourceConfig xfccConfig, Principal requestPrincipal) {
@@ -296,7 +319,7 @@ public class ClientAuthFactory {
     return clientAuthConfig.createMissingClients();
   }
 
-  private Client authenticateClientFromCertificate(Principal clientPrincipal) {
+  private Client authenticateClientFromPrincipal(Principal clientPrincipal) {
     Optional<Client> possibleClient =
         authenticator.authenticate(clientPrincipal, createMissingClient());
     return possibleClient.orElseThrow(() -> new NotAuthorizedException(
--- a/server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
+++ b/server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
@@ -12,12 +12,14 @@ import java.util.stream.Collectors;
 import javax.ws.rs.NotAuthorizedException;
 import keywhiz.api.model.Client;
 import keywhiz.auth.mutualssl.CertificatePrincipal;
+import keywhiz.auth.mutualssl.SpiffePrincipal;
 import keywhiz.service.config.ClientAuthConfig;
 import keywhiz.service.daos.ClientDAO;
 import org.bouncycastle.asn1.x500.RDN;
 import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x500.style.BCStyle;
 import org.bouncycastle.asn1.x500.style.IETFUtils;
+import org.glassfish.jersey.server.ContainerRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -120,6 +122,10 @@ public class ClientAuthenticator {
   }
 
   static Optional<String> getClientName(Principal principal) {
+    if (principal instanceof SpiffePrincipal) {
+      return Optional.of(((SpiffePrincipal) principal).getClientName());
+    }
+
     X500Name name = new X500Name(principal.getName());
     RDN[] rdns = name.getRDNs(BCStyle.CN);
     if (rdns.length == 0) {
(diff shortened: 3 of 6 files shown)


jbpeirce (changes requested):
The new header to identify the peer that's forwarding a request significantly reduces Keywhiz' security. With the original code, impersonating Envoy/another proxy required getting access to that proxy's identification certificates (to control the SecurityContext's "Principal" identity on the incoming request). As far as I can tell from the code and description, in order to impersonate a proxy (like Envoy) with this change, all that's required is the ability to set the `peerSpiffeIdHeader` to a value listed in `allowedPeerWorkloadIds`. If that's the case, anyone with control of the headers for the request they're forging and some internal knowledge of Keywhiz (or time to experiment) can impersonate any Keywhiz client.

I'd be comfortable approving a PR that allowed the proxy to set a header to identify the real client rather than forwarding the real client's entire certificate--Keywhiz is trusting the proxy no matter what. However, allowing the proxy to also identify itself using a header alone significantly weakens Keywhiz' security here.

FWIW, my original plan for handling the pain of maintaining the list was to identify proxies by their SPIFFE IDs; then, the list will need changes only when a new proxy or new environment is added.

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -149,43 +150,48 @@ private Client authorizeClientFromXfccHeader(XfccSourceConfig xfccConfig,
...
+    Optional<URI> callerSpiffeId = spiffeIdHeader.isPresent()
+        ? ClientAuthenticator.getSpiffeIdFromHeader(containerRequest, spiffeIdHeader.get())
+        : Optional.empty();
+
     // All XFCC traffic must be checked against allowlists; all connections that set the XFCC
     // header must send identifiers for the requester, as well as the XFCC certificate
     Optional<String> requestName = ClientAuthenticator.getClientName(requestPrincipal);
-    Optional<URI> requestSpiffeId = ClientAuthenticator.getSpiffeId(requestPrincipal);
+    Optional<URI> requestSpiffeId = callerSpiffeId.isPresent()
+        ? callerSpiffeId : ClientAuthenticator.getSpiffeId(requestPrincipal);
```
jbpeirce, line 176: This should always be `ClientAuthenticator.getSpiffeId(requestPrincipal)`.

There are two entities involved in the process of handling a proxied connection to Keywhiz: the proxy and the actual client whose request was proxied. This code uses the "request" prefix to indicate data related to the proxy, because that data comes from the mTLS request that Keywhiz received directly. The requester/proxy should _not_ be identified based on a header.

This function only checks that the entity which connected to Keywhiz is allowed to do so, so the forwarded "caller spiffe ID" shouldn't be used here.

The code that needs to be updated is the "extract client information from XFCC header" section starting in line 144--that's where the client whose request is being proxied is identified. Currently there's only one branch there because the only way that a client can be identified is from the XFCC header, but we should add a new case for a client identified by the caller-spiffe-ID header.

(Please feel free to add comments describing the logic--it's more confusing than I initially realized when writing it.)

jbpeirce (changes requested):

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
KenjiChao: Thanks for the feedback and the pointer. So for the caller authentication can I extract the Spiffe Id from the header and check if it's in the `allowedCallerSpiffeId`? Or do you think I should create a principal extracted from the caller Spiffe Id and reuse `authenticateClientFromCertificate`?

jbpeirce: Handcrafting a principal is pretty tricky iirc, so I'd weakly recommend not trying to reuse `authenticateClientFromCertificate`--but you can probably refactor ClientAuthenticator to share code between `authenticateClientFromCertificate` and `authenticateClientFromSpiffeId` (or whatever the new method's called). But it's definitely worth checking whether you can create a principal with the caller Spiffe ID and use it in a (renamed) `authenticateClientFromCertificate` method.

KenjiChao: @jbpeirce I made some changes so that now we authorize the requests from a custom header starting in line 144. Thank you again for the pointer. I didn't fully understand the auth process in my previous iterations. Let me know if this looks good to you.

server/src/main/java/keywhiz/auth/mutualssl/SpiffePrincipal.java
```diff
@@ -0,0 +1,28 @@
...
+  public URI getSpiffeId() {
+    return spiffeId;
+  }
+
+  /**
+   * Use the workload id of a Spiffe Id as the client name.
+   */
+  public String getClientName() {
+    String path = spiffeId.getPath();
+    return path.isEmpty() ? path : path.substring(1);
```
jbpeirce, line 26: Why do we drop the first character from the `spiffeId` path here, with `path.substring(1)`?

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -134,12 +135,23 @@ public Client provide(ContainerRequest containerRequest,
   }
 
   private Client authorizeClientFromXfccHeader(XfccSourceConfig xfccConfig,
```
jbpeirce, line 137: nit: Since the XFCC header isn't necessarily what's used, maybe this should be renamed to `authoritizeClientFromForwardedData` or something.

jbpeirce: (though this would be a bigger change, and should probably be handled in a separate commit, since the term "XFCC" is used a lot in this code)

server/src/main/java/keywhiz/auth/mutualssl/SpiffePrincipal.java
KenjiChao: To drop the '/' character

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -134,12 +135,23 @@ public Client provide(ContainerRequest containerRequest,
...
-      List<String> xfccHeaderValues, Principal requestPrincipal) {
+      List<String> xfccHeaderValues, Principal requestPrincipal,
+      ContainerRequest containerRequest) {
     // Do not allow the XFCC header to be set by all incoming traffic. This throws a
     // NotAuthorizedException when the traffic is not coming from a source allowed to set the
     // header.
     validateXfccHeaderAllowed(xfccConfig, requestPrincipal);
 
+    Optional<String> spiffeIdHeader = Optional.ofNullable(xfccConfig.callerSpiffeIdHeader());
+    Optional<URI> callerSpiffeId = spiffeIdHeader.isPresent()
```
jbpeirce, line 146: nit: You can use [Optional.flatMap()](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#flatMap-java.util.function.Function-) for this rather than a ternary:
```
spiffeIdHeader.flatMap(header -> ClientAuthenticator.getSpiffeIdFromHeader(containerRequest, header))
```
will return an Optional which is empty if `spiffeIdHeader` was empty, and otherwise is the value of `getSpiffeIdFromHeader`.

server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
```diff
@@ -120,6 +122,10 @@ public ClientAuthenticator(
   }
 
   static Optional<String> getClientName(Principal principal) {
+    if (principal instanceof SpiffePrincipal) {
```
jbpeirce, line 125: nit: Since we control both `SpiffePrincipal` and `CertificatePrincipal`, we could get rid of a lot of `instanceof` checks by having both principals declare a `getClientName` method and `getSpiffeId` method (maybe have a `KeywhizPrincipal` interface that they both extend which adds those two methods). Again, though, that should be a separate commit and possibly another PR.

server/src/main/java/keywhiz/service/providers/ClientAuthenticator.java
```diff
@@ -129,20 +135,24 @@ public ClientAuthenticator(
...
-    // This chain is either from the XFCC header's "Cert" field, which includes only the
-    // client certificate rather than the chain, or from the CertificateSecurityContext
-    // configured by Keywhiz' ClientCertificateFilter, which sets it based on
-    // X509Certificate[] chain =
-    //        (X509Certificate[]) context.getProperty("javax.servlet.request.X509Certificate");
-    // which appears to place the leaf as the zero-index entry in the chain.
-    X509Certificate cert = ((CertificatePrincipal) principal).getCertificateChain()
-        .get(0);
-    return getSpiffeIdFromCertificate(cert);
+    // A SPIFFE ID can only be parsed from a principal with a certificate
```
jbpeirce, line 154: nit: this comment is no longer accurate

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
```diff
@@ -193,37 +206,71 @@ public void clientWhenPrincipalAbsentThrows() {
     assertThat(factory.provide(request, httpServletRequest)).isEqualTo(client);
   }
 
-  @Test(expected = NotAuthorizedException.class)
-  public void rejectsXfcc_requesterAuthMissing() {
+  @Test public void returnsClientWhenClientPresent_fromXfccHeader_customSpiffeIdHeader() {
```
jbpeirce, line 209: Do we ever expect both the XFCC header and the custom SPIFFE ID header to be set? If so, Keywhiz should check that the same client is extracted from both headers. If not, Keywhiz should check that only one header is set and deny access if they're both set, since the request is malformed.

jbpeirce, line 209: Do we ever expect both the XFCC header and the custom SPIFFE ID header to be set? If so, Keywhiz should check that the same client is extracted from both headers (and deny access if they don't match). If not, Keywhiz should check that at most one header is set and deny access if both are set.

jbpeirce: Please make sure to have some tests with only the custom SPIFFE ID header set, so that we can be sure Keywhiz isn't falling back to the XFCC header path somehow (for instance, if the SPIFFE ID name mismatched) and still retrieving the client that way.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
```diff
@@ -193,37 +206,71 @@ public void clientWhenPrincipalAbsentThrows() {
...
     when(request.getRequestHeader(ClientAuthFactory.XFCC_HEADER_NAME)).thenReturn(
         List.of(xfccHeader));
     when(securityContext.getUserPrincipal()).thenReturn(xfccPrincipal);
 
-    when(xfccSourceConfig.allowedClientNames()).thenReturn(List.of());
-    when(xfccSourceConfig.allowedSpiffeIds()).thenReturn(List.of(xfccSpiffe));
+    when(xfccSourceConfig.callerSpiffeIdHeader()).thenReturn(callerSpiffeHeaderName);
+    when(request.getRequestHeader(callerSpiffeHeaderName)).thenReturn(List.of("not-spiffe-id"));
+
+    assertThat(factory.provide(request, httpServletRequest)).isEqualTo(client);
```
jbpeirce, line 231: Why is this test passing? `not-spiffe-id` in the caller-identification header should be unable to identify a client.

(In this case, I suspect the check is somehow falling back to the XFCC header, but we should definitely have a test that an invalid SPIFFE ID in the caller-identification header can't retrieve a client if the caller-identification header is the only one provided.)

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
```diff
@@ -193,37 +206,71 @@ public void clientWhenPrincipalAbsentThrows() {
...
+    when(request.getRequestHeader(ClientAuthFactory.XFCC_HEADER_NAME)).thenReturn(
+        List.of(xfccHeader));
+    when(securityContext.getUserPrincipal()).thenReturn(null);
 
     factory.provide(request, httpServletRequest);
   }
 
   @Test(expected = NotAuthorizedException.class)
-  public void rejectsXfcc_requesterSpiffeNotAllowed() {
+  public void rejectsXfcc_requesterNameNotAllowed_requesterSpiffeNotAllowed() {
```
jbpeirce, line 267: Does this test still need to be changed? The goal here iirc was to ensure that if the proxy provides both a name and SPIFFE ID, and _either_ of those is disallowed, then the request will be rejected even if the other was allowed. Unless we've changed that behavior, it should still be tested.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
```diff
@@ -319,4 +366,18 @@ public void rejectsXfcc_missingClientCert() {
...
+  @Test(expected = NotAuthorizedException.class)
+  public void rejectsXfcc_customSpiffeIdHeader_clientNotPresent() {
+    when(httpServletRequest.getLocalPort()).thenReturn(xfccAllowedPort);
+    when(request.getRequestHeader(ClientAuthFactory.XFCC_HEADER_NAME)).thenReturn(
+        List.of(xfccHeader));
+    when(securityContext.getUserPrincipal()).thenReturn(xfccPrincipal);
+
+    when(xfccSourceConfig.callerSpiffeIdHeader()).thenReturn(callerSpiffeHeaderName);
+    when(request.getRequestHeader(callerSpiffeHeaderName)).thenReturn(
+        List.of(clientSpiffeNotAllowed));
```
jbpeirce, line 379: nit: since the issue here is that the client doesn't exist and we didn't configure Keywhiz to create missing clients, I'd recommend naming this variable `nonexistentClientSpiffe` or something similar. ("not allowed" makes me think the client might exist, but be forbidden to do whatever I tried to do.)

jbpeirce (changes requested):
This is looking a lot better! I still have some questions on the test about the interaction between the XFCC header and the new header, but this is very close. Feel free to disregard any of the "nit" comments (particularly since some of them would require a refactor and a separate commit).

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
KenjiChao: I think it's fine to keep XFCC name because this custom header is an optional config inside XFCC, just to support a scenario where a real client is behind a proxy.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
KenjiChao: I've updated the commit message to reflect the fact that the custom header should come with the XFCC header, and they may or may not match depending whether a client is accessing Keywhiz through a proxy.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
KenjiChao: I changed the behavior so now the request would be rejected if the URI doesn't follow SPIFFE schema.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
KenjiChao: My bad. I've changed it back

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -134,12 +135,34 @@ public Client provide(ContainerRequest containerRequest,
...
+    List<String> callerSpiffeIdList = callerSpiffeIdHeader.map(
+        header -> Optional.ofNullable(containerRequest.getRequestHeader(header))
+            .orElse(List.of()))
+        .orElse(List.of());
+    int size = callerSpiffeIdList.size();
+
+    Optional<URI> callerSpiffeId = callerSpiffeIdHeader.flatMap(
+        header -> ClientAuthenticator.getSpiffeIdFromHeader(containerRequest, header));
+
+    if (size > 1 || size == 1 && callerSpiffeId.isEmpty()) {
```
jbpeirce, line 155: nit: although this works by the operator rules (&& is evaluated before ||), parentheses would make it slightly more readable.

server/src/test/java/keywhiz/service/providers/ClientAuthFactoryTest.java
jbpeirce: Ah, I see how this works now :) (I hadn't realized this was falling back to the client identity in the xfccHeader.)

jbpeirce (approved):
This is looking good to me--thanks for all the updates!

jbpeirce:
Suspect that the Travis CI failure is caused by merging from a fork; since the other build passes, I'm merging this to the main branch (cc @wtmcc who also reviewed).

PR #866, opened 2021-06-17
CLI Rename: Adding in resource type parameter (currently, secret-only)

mmontgomery-square:


cli/src/main/java/keywhiz/cli/commands/RenameAction.java
```diff
@@ -16,26 +16,24 @@ public RenameAction(RenameActionConfig config, KeywhizClient keywhiz) {
   }
 
   @Override public void run() {
-    validate();
+    switch (config.resourceType) {
```
mmontgomery-square, line 19: I removed support for ID-based rename, it looks like almost everything else in the CLI uses names (good for humans), which means we no longer have the exclusive-or validation case, which in turn means we can now rely on JCommander's built-in validation now that all parameters are required.

cli/src/main/java/keywhiz/cli/commands/RenameAction.java
```diff
@@ -16,26 +16,24 @@ public RenameAction(RenameActionConfig config, KeywhizClient keywhiz) {
...
+      case "secret":
+        renameSecret();
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format(
+                "Unsupported resource type %s. Only the secret resource type can be renamed.",
+                config.resourceType));
+    }
```
mmontgomery-square, line 28: If we ever add other types this should be changed to list out some enum values, but I thought this was a bit more user-friendly for now.

cli/src/main/java/keywhiz/cli/configs/RenameActionConfig.java
```diff
@@ -6,12 +6,12 @@
...
-  @Parameter(names = "--name", description = "Name of the secret to rename")
-  public String secretName;
+  @Parameter(names = "--oldName", description = "Resource to rename", required = true)
+  public String oldName;
 
-  @Parameter(names = "--id", description = "ID of the secret to rename")
-  public Long secretId;
+  @Parameter(names = "--newName", description = "New name", required = true)
+  public String newName;
 }
```
mmontgomery-square, line 17: Hopefully this is more clear.

cli/src/test/java/keywhiz/cli/CliMainTest.java
```diff
@@ -20,30 +25,44 @@ public void renameCommandMapsToRenameActionConfig() {
...
+  }
+
+  @Test
+  public void failsToParseRenameCommandWithoutNewName() {
+    CliMain.CommandLineParsingContext context = new CliMain.CommandLineParsingContext();
+    JCommander commander = context.getCommander();
+
+    thrown.expect(ParameterException.class);
+    commander.parse("rename", "secret", "--oldName", "foo");
   }
```
mmontgomery-square, line 52: Now that our validation is at the JCommander level our missing-argument tests need to be as well.

cli/src/main/java/keywhiz/cli/commands/RenameAction.java
mbyczkowski: nice!

cli/src/main/java/keywhiz/cli/commands/RenameAction.java
mbyczkowski: I think that's a good call

mbyczkowski (approved):

PR #875, opened 2021-06-18
Tweaking rename action description so it makes sense in the help output

mmontgomery-square:


cli/src/main/java/keywhiz/cli/configs/RenameActionConfig.java
```diff
@@ -6,7 +6,7 @@
 @Parameters(commandDescription = "Rename a secret")
 public class RenameActionConfig {
 
-  @Parameter(description = "Resource type: <secret>", required = true)
+  @Parameter(description = "<secret>", required = true)
   public String resourceType;
 
   @Parameter(names = "--oldName", description = "Resource to rename", required = true)
```
mmontgomery-square, line 12: I should provide more context:

In the help output, JCommander joins the description of "inline" (non-named) parameters directly with the command name itself, whereas named parameters are described separately in a bullet list. So, if your inline parameter description is "X Y Z" you'll get the following help output:

`myCommand X Y Z`

In our case, the existing style (based on other commands) is that the message reflects the actual command format, with any choices enclosed in angle brackets. Since our description includes "Resource type: " that is obviously not the case. Removing that prefix and leaving just `<secret>` makes the output match the desired format.

mbyczkowski (approved):

PR #885, opened 2021-07-06
Adding support for maximum secret size limit

mmontgomery-square:


server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -222,6 +225,12 @@ public ClientAuthConfig getClientAuthConfig() {
     return clientAuthConfig;
   }
 
+  public Long getMaximumSecretSizeInBytes() { return maximumSecretSizeInBytes; }
+
+  public void setMaximumSecretSizeInBytes(Long maximumSecretSizeInBytes) {
+    this.maximumSecretSizeInBytes = maximumSecretSizeInBytes;
+  }
```
mmontgomery-square, line 232: This property also has a setter for ease of testing (also, there's precedent for a handful of other properties). Without this you have to rely on Jackson or some other framework to transform JSON/YAML/etc. into a POJO, which feels like overkill.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -61,7 +61,7 @@
   private final RowHmacGenerator rowHmacGenerator;
   private final KeywhizConfig config;
 
-  private SecretContentDAO(DSLContext dslContext, ObjectMapper mapper,
+  SecretContentDAO(DSLContext dslContext, ObjectMapper mapper,
       SecretContentMapper secretContentMapper, RowHmacGenerator rowHmacGenerator,
       KeywhizConfig config) {
     this.dslContext = dslContext;
```
mmontgomery-square, line 67: For ease of testing, this constructor is now package private. This avoids use of the associated DAO factory and its built-in behavior.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -75,6 +75,16 @@ public long createSecretContent(long secretId, String encryptedContent, String h
...
 
+    if (config.getMaximumSecretSizeInBytes() != null) {
+      if (encryptedContent.length() > config.getMaximumSecretSizeInBytes()) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Secret is too large. Secret size %d bytes exceeds maximum size of %d bytes",
+                encryptedContent.length(),
+                config.getMaximumSecretSizeInBytes()));
+      }
+    }
```
mmontgomery-square, line 86: This is the actual behavior change. In hindsight, it probably makes sense to push it down into its own method so that this one doesn't get too cluttered.

In keeping with existing validation for secret series names, this logic is specified at the DAO level since it's a chokepoint for various APIs.

server/src/test/java/keywhiz/KeywhizConfigTest.java
```diff
@@ -0,0 +1,60 @@
...
+  public void parsesMaximumSecretLength() {
+    KeywhizConfig config = loadConfig("with-secret-size-limit-123456.yaml");
+    assertEquals(Long.valueOf(123456), config.getMaximumSecretSizeInBytes());
+  }
+
+  @Test
+  public void missingMaximumSecretLengthIsNull() {
+    KeywhizConfig config = loadConfig("without-secret-size-limit.yaml");
+    assertNull(config.getMaximumSecretSizeInBytes());
+  }
```
mmontgomery-square, line 31: This class verifies that the property is parsed when present and ignored when not (i.e. backwards compatibility). It uses some fancy logic that's copied from KeywhizTestRunner.

server/src/test/java/keywhiz/KeywhizConfigTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private static KeywhizConfig loadConfig(String resource) {
+    Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+    ObjectMapper objectMapper = createObjectMapper();
+    File yamlFile = new File(Resources.getResource(TEST_CONFIGS_PREFIX + resource).getFile());
```
mmontgomery-square, line 36: I created a new folder for test configs, similar to how there are already folders for other test resource types. The idea was to be able to easily load configs with specific properties and run tests against them.

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
```diff
@@ -0,0 +1,119 @@
...
+  public void rejectsSecretWithSizeGreaterThanMaximumSecretSize() {
+    Long maximumSecretSizeInBytes = 100L;
+    String secret = StringUtils.repeat("x", (int) (maximumSecretSizeInBytes + 1));
+
+    SecretContentDAO dao = createDaoWithMaximumSecretSize(maximumSecretSizeInBytes);
+
+    thrown.expect(IllegalArgumentException.class);
+
+    createSecret(dao, secret);
+  }
```
mmontgomery-square, line 45: There are tests for greater than, equal to, less than, and large secrets (65KB) with no limit set. In order to keep the tests readable there's a lot of magic/boilerplate hidden behind helper methods.

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private void createSecret(SecretContentDAO dao, String secret) {
+    dao.createSecretContent(
+        NO_SECRET_ID,
+        secret,
+        NO_HMAC,
+        NO_CREATOR,
+        NO_METADATA,
+        NO_EXPIRY,
+        NO_NOW);
+  }
```
mmontgomery-square, line 86: This just exists because the method invocation is ugly (tons of default-value arguments).

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
```diff
@@ -0,0 +1,119 @@
...
+      for (int i = 0; i < mockExecuteContext.autoGeneratedKeys(); i++) {
+        Record1<Long> record = context.newRecord(SecretsContent.SECRETS_CONTENT.ID);
+        Record1<Long> values = record.values(1L);
+        MockResult result = new MockResult(values);
+        results.add(result);
+      }
+
+      return results.toArray(new MockResult[0]);
+    }
+  }
```
mmontgomery-square, line 118: This is taken from the Jooq unit testing docs. Unfortunately there is no actual implementation of MockDataProvider so you always have to roll your own. The behavior we see here is to satisfy the case where an insert statement is run and an auto-generated ID is expected as the result. In other words, the behavior of SecretsContentRecord.store() from our DAO.

The alternative is many, many layers of mocking the Jooq APIs which I very much wanted to avoid.

server/src/test/resources/configs/with-secret-size-limit-123456.yaml
```diff
@@ -0,0 +1,159 @@
+maximumSecretSizeInBytes: 123456
+
+server:
```
mmontgomery-square, line 3: Front and center.

Unfortunately the heavy validation means that a lot of unrelated values need to be specified in order to have a valid config. One alternative would be to disable validation in the tests but that wouldn't match Keywhiz's actual behavior.

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.jooq.tools.jdbc.MockConnection;
+import org.jooq.tools.jdbc.MockDataProvider;
+import org.jooq.tools.jdbc.MockExecuteContext;
+import org.jooq.tools.jdbc.MockResult;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.mockito.Mockito;
+
+public class SecretContentDAOMockTest {
```
mmontgomery-square, line 24: There's already a SecretContentDAOTest but it relies on dependency injection via a JUnit Rule to instantiate the object graph. The problem with this is that the configuration file is hardcoded so we have no clean way of using different configurations (we could grab the injected config and update the value, but the whole injection thing is a lot of overhead for something which should be easy to test).

server/src/main/java/keywhiz/KeywhizConfig.java
isemaya-square: Maybe leave a comment for this setter method saying that it its main purpose is to make testing easier and not intended for use outside of testing?

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
```diff
@@ -0,0 +1,119 @@
...
+    Long maximumSecretSizeInBytes = 100L;
+    String secret = StringUtils.repeat("x", (int) (long) (maximumSecretSizeInBytes));
+
+    SecretContentDAO dao = createDaoWithMaximumSecretSize(maximumSecretSizeInBytes);
+
+    createSecret(dao, secret);
+  }
+
+  @Test
+  public void acceptsSecretWithSizeLessThanMaximumSize() {
```
isemaya-square, line 58: If I'm not mistaken, isn't the contents of this test the same as `acceptsSecretWithSizeEqualToMaximumSize` above?

server/src/test/resources/configs/with-secret-size-limit-123456.yaml
isemaya-square: Makes sense, not pretty to look at but at least we're testing keywhiz' actual behavior.

isemaya-square: Though curious: since we have two separate yamls here (one with the secret size, one without, could we have a "common" yaml (that isn't actually used in any of the tests) and `with-secret-size-limit-123456.yaml` could then have an `"%include "test-common.yaml"`?

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
isemaya-square: Dang, making us roll our own method! Out of curiosity, did you know to roll this out (rather than mock Jooq APIs) immediately, or did you try to mock Jooq APIs and realize it was going to be tedious?

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
mmontgomery-square: I'll trace through the call stack to see how the secret comes in. IIRC we're already expecting it to be encoded in some fashion but I want to be sure.

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
mmontgomery-square: I started with mocking first, but Jooq's API is builder-ish in that it likes to return lots of objects, and those objects are built up by lots of calls. So, multiple methods on multiple classes would have been necessary for mocking. I've been down that road before and wanted to avoid it if at all possible. OTOH it's a one-time investment, but hopefully so is a working example of how to test "by the rules".

server/src/test/resources/configs/with-secret-size-limit-123456.yaml
mmontgomery-square: Huh, I've seen the include syntax but it hadn't occurred to me to try it. I don't know how it works under the hood but my guess is that it'll fail for this use case because the test files are resources packaged up in JARs and not actual files on the filesystem like they are in deployed apps.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
```diff
@@ -75,6 +75,16 @@ public long createSecretContent(long secretId, String encryptedContent, String h
       String creator, Map<String, String> metadata, long expiry, long now) {
     SecretsContentRecord r = dslContext.newRecord(SECRETS_CONTENT);
 
+    if (config.getMaximumSecretSizeInBytes() != null) {
```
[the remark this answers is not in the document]
mmontgomery-square: Generally I like to keep if statements as simple as possible so they're easier to understand. In this case I wanted to signal that the following block of code only applies if there is actually a size limit set in the configuration - a guard clause. However, in hindsight I think this is better moved to a helper function where the guard clause can return early, reducing nesting (and also clutter in the main method).

server/src/test/java/keywhiz/service/daos/SecretContentDAOMockTest.java
mmontgomery-square: Good catch! I got the +1 case correct, looks like I just derped on this one.

server/src/main/java/keywhiz/KeywhizConfig.java
mmontgomery-square: Normally I would slap on a @VisibleForTesting but I wasn't sure what the purpose of the other setters was (they aren't similarly marked). I'll go ahead and add this.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
mmontgomery-square: I'll add @VisibleForTesting here as well.

server/src/test/resources/configs/with-secret-size-limit-123456.yaml
mmontgomery-square: I tested this out locally and, as expected, include directives do not work for classpath resources.

server/src/main/java/keywhiz/service/daos/SecretContentDAO.java
mmontgomery-square: Our server-side APIs declare that the secret content is @ValidBase64, and in the Keywhiz CLI client there's a line which is responsible for transforming the raw secret into base64 before sending it across the wire. I'll put together an error message which takes this into account and also provides a more useful hint as to the acceptable raw secret size.

mmontgomery-square: Upon further reflection, I think moving the validation before encryption happens will allow us to more accurately reflect the maximum secret size back to the caller. Fortunately there is another chokepoint, SecretController, that I'd missed previously. Moving the logic there makes this a better user experience and also simplifies testing somewhat.

Stay tuned.

server/src/main/java/keywhiz/service/daos/SecretController.java
```diff
@@ -219,6 +224,23 @@ public SecretBuilder builder(String name, String secret, String creator, long ex
...
+
+    int rawSecretLength = Base64.getDecoder().decode(base64EncodedSecret).length;
+
+    if (rawSecretLength > config.getMaximumSecretSizeInBytesInclusive()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Secret is too large. Secret size %,d bytes exceeds maximum size of %,d bytes.",
+              rawSecretLength,
+              config.getMaximumSecretSizeInBytesInclusive()));
+    }
```
mmontgomery-square, line 240: New and improved:

* Configuration option name is now even more self-describing via the addition of "Inclusive"
* Exact raw secret size is now used for calculation (previously we were using the post-encryption, post-encoding version). This should be more user-friendly.
* Large byte values are now formatted with a thousands separator (this is what `%,d` means)

server/src/test/java/keywhiz/service/daos/SecretControllerMockTest.java
```diff
@@ -0,0 +1,84 @@
...
+public class SecretControllerMockTest {
+  private static final Long MAXIMUM_SECRET_SIZE_IN_BYTES_INCLUSIVE = 100L;
+  private static final long NO_EXPIRY = 0;
+  private static final Long NO_MAXIMUM_SECRET_SIZE = null;
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void rejectsSecretWithSizeGreaterThanMaximumSecretSize() {
```
mmontgomery-square, line 25: A nice side effect of moving the validation logic from the DAO layer to the controller layer is that we don't have to deal with Jooq, making the test setup much simpler.

mmontgomery-square:
In this new revision, we move the validation logic from the DAO layer to the controller, giving us access to the secret before it's encrypted. This allows us to make exact calculations against the original secret size, providing a better user experience.

server/src/test/resources/configs/without-secret-size-limit.yaml
```diff
@@ -0,0 +1,157 @@
...
+      keyStoreType: PKCS12
+      trustStorePath: dev_and_test_truststore.p12
+      trustStorePassword: ponies
+      trustStoreType: PKCS12
+      wantClientAuth: true
+      validateCerts: true
+      validatePeers: true
+      crlPath: dev_and_test.crl
+      supportedProtocols: [TLSv1.2]
+      supportedCipherSuites:
```
mbyczkowski, line 16: It would be nice if there was a way to combine keywhiz base config with a small portion of YAML that we'd like to have modified, but this is probably fine for now.

mbyczkowski (approved):

server/src/test/resources/configs/without-secret-size-limit.yaml
mmontgomery-square: Actually, one really cheesy way to do this would be to have the code load two files and then just merge them together, but I'll leave it as-is for now.

mmontgomery-square: Yeah, the problem is there's modest encryption and packaging (JSON, I think) overhead for our secrets, so the size of the raw secret that the user uploads doesn't exactly match what the storage layer sees. But, I thought it was better to speak in user-facing terms and put the burden of calculating a reasonable buffer on our end.

PR #886, opened 2021-07-07
First pass at OpenTracing integration

mmontgomery-square:


server/src/main/java/keywhiz/Tracing.java
```diff
@@ -0,0 +1,91 @@
...
+      throw e;
+    }
+  }
+
+  private static void tagErrors(Span span, Exception e) {
+    span.setTag(Tags.ERROR, true);
+
+    tagErrorsOpenTracing(span, e);
+    tagErrorsDataDog(span, e);
+  }
```
mmontgomery-square, line 49: I'm wary of introducing any mention of DataDog (let alone a dependency, which would grant us access to @Trace and make this helper code obsolete), but, it's popular and that's what we use, and using their specific tags results in a much better experience on our end.

server/src/main/java/keywhiz/Tracing.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private static void tagErrorsOpenTracing(Span span, Exception e) {
+    https://github.com/opentracing/specification/blob/11dd7f8b16739f02761a9541f04e39f27738f525/semantic_conventions.md#span-and-log-errors
+
+    span.log(
```
mmontgomery-square, line 54: The fact that URLs are legal Java syntax always cracks me up

server/src/main/java/keywhiz/auth/ldap/LdapAuthenticator.java
```diff
@@ -53,6 +55,10 @@ public LdapAuthenticator(LdapConnectionFactory connectionFactory, LdapLookupConf
 
   @Override
   public Optional<User> authenticate(BasicCredentials credentials) {
+    return trace("LdapAuthenticator.authenticate", () -> doAuthenticate(credentials));
+  }
+
+  private Optional<User> doAuthenticate(BasicCredentials credentials) {
```
mmontgomery-square, line 61: Rather than wrapping larger methods in a lambda and adding another level of indentation, I find it more readable to just move the logic into a "doX" helper method so that the metrics wrapping is hidden from the actual business logic as much as possible.

server/src/main/java/keywhiz/service/providers/ClientAuthFactory.java
```diff
@@ -114,6 +116,15 @@ public ClientAuthFactory(ClientDAOFactory clientDAOFactory, KeywhizConfig keywhi
    */
   public Client provide(ContainerRequest containerRequest,
       HttpServletRequest httpServletRequest) {
+    return trace("ClientAuthFactory.provide", () -> {
+      Client client = doProvide(containerRequest, httpServletRequest);
+      setTag("authenticatedClient", client.getName());
+      return client;
+    });
```
mmontgomery-square, line 123: This will provide us a nice record of who we're talking to in our traces.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -130,9 +131,11 @@ public BatchSecretDeliveryResource(@Readonly SecretController secretController,
     logger.info("Client {} granted access to {}.", client.getName(), clientAccessibleSecrets.stream().map(s -> s.name()).collect(toList()));
     try {
       // This is only possible if all secrets are both existing AND accessible to the client
-      return existingSecrets.stream()
+      List<SecretDeliveryResponse> secrets = existingSecrets.stream()
               .map(SecretDeliveryResponse::fromSecret)
               .collect(toList());
+      setTag("nSecrets", secrets.size());
+      return secrets;
```
mmontgomery-square, line 138: For methods which occasionally take in the hundreds of milliseconds, presumably due to larger result set sizes, I added a record of just how many things we found.

It's worth noting that this batching method has poor behavior in that if X secrets are found then X sequential calls are made to the HSM. We should look into a batching API at the HSM level.

server/src/main/java/keywhiz/service/resources/admin/SessionLoginResource.java
```diff
@@ -80,6 +81,8 @@ public Response login(@Valid LoginRequest request) {
     String username = request.username();
     String password = String.copyValueOf(request.password());
 
+    setTag("username", username);
+
     Optional<User> optionalUser = Optional.empty();
```
mmontgomery-square, line 86: For diagnosing failed login attempts.

server/src/main/java/keywhiz/service/resources/automation/v2/ClientResource.java
```diff
@@ -92,9 +94,16 @@
...
+  }
+
+  private Response doCreateClient(AutomationClient automationClient,
+      CreateClientRequestV2 request) {
     String creator = automationClient.getName();
     String client = request.name();
 
+    setTag("client", client);
+
     clientDAOReadWrite.getClientByName(client).ifPresent((c) -> {
```
mmontgomery-square, line 107: The most frequent errors we get for non-batch secret requests are for creating groups and clients.

server/src/main/java/keywhiz/Tracing.java
isemaya-square: What's the issue with mentioning DataDog? Is it that we want to avoid being tied to a specific library/project?

server/src/main/java/keywhiz/Tracing.java
isemaya-square: No way! TIL

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
isemaya-square: Should we make a TODO with a ticket? Then again, I'm not sure if we actually take action on TODOs anyway...

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
```diff
@@ -130,9 +131,11 @@ public BatchSecretDeliveryResource(@Readonly SecretController secretController,
     logger.info("Client {} granted access to {}.", client.getName(), clientAccessibleSecrets.stream().map(s -> s.name()).collect(toList()));
     try {
       // This is only possible if all secrets are both existing AND accessible to the client
-      return existingSecrets.stream()
+      List<SecretDeliveryResponse> secrets = existingSecrets.stream()
               .map(SecretDeliveryResponse::fromSecret)
               .collect(toList());
+      setTag("nSecrets", secrets.size());
```
isemaya-square, line 137: nit: should we say "numberOfSecrets" or "numSecrets" to make it clearer? I guess "nSecrets" isn't unclear, but why not spell it out in full?

isemaya-square (approved):
LGTM overall! 

Question though: how did you determine which methods should be traced? For example, for `GroupResource`, why only tag `createGroup` and say, not `secretsForGroup`? (I'm assuming it's because we care about new info (like groups) being added rather than called, and if someone (or rather, some entity) is making an API call to get secrets or groups, then presumably, that entity was already authN/Z'd?)

mweissbacher (approved):

server/src/main/java/keywhiz/Tracing.java
mmontgomery-square: Yeah, I look at DataDog as a Square-specific dependency whereas Keywhiz is supposed to be more generic due to its open-sourciness. OTOH I don't want to unnecessarily handicap ourselves.

server/src/main/java/keywhiz/Tracing.java
mmontgomery-square: It's because the protocol is interpreted as a label, which you tend not to see too much of, and everything after the slashes is interpreted as a comment. Of course, since labels need to be unique within a method, you can't just go crazy, and this is definitely an abuse of syntax, but still fun trivia.

server/src/main/java/keywhiz/service/resources/BatchSecretDeliveryResource.java
mmontgomery-square: I created an issue and linked it to the original issue for this task so that it wouldn't get lost.

mmontgomery-square:
> LGTM overall! I don't see anything that should block you.
> 
> Question though: how did you determine which methods should be traced? For example, for `GroupResource`, why only tag `createGroup` and say, not `secretsForGroup`? (I'm assuming it's because we care about new info (like groups) being added rather than called, and if someone (or rather, some entity) is making an API call to get secrets or groups, then presumably, that entity was already authN/Z'd?)
> 
> So in other words, could we be missing tags that could be helpful?

I'm sure I didn't cover everything. This was intended to be a foot-in-the-door change to demonstrate a working integration and also cover the most common use cases (which I determined by looking at our current traffic - typical errors, slow methods, etc.). I didn't want to saturate the codebase with tracing that we don't need, which is in line with DataDog's own recommendations (sprinkle additional data/traces as needed to diagnose specific issues rather than a shotgun approach). If we do need additional traces/tags (which I'm sure we will), it should be trivial to add it now that the basic integration work is done.

PR #892, opened 2021-07-16
Instruction on how to provide secret content

ruiwsquare:
When creating a new secret with add CLI, the user needs to enter the secret content then Ctrl+D to signal end of input. Otherwise, the command cannot finish and secret cannot be created.

### DIFF
--- a/cli/src/main/java/keywhiz/cli/commands/AddAction.java
+++ b/cli/src/main/java/keywhiz/cli/commands/AddAction.java
@@ -139,6 +139,7 @@ public class AddAction implements Runnable {
 
   private byte[] readSecretContent() {
     try {
+      System.out.print("Enter secret content, Ctrl+D when done: ");
       byte[] content = ByteStreams.toByteArray(stream);
       if (content.length == 0) {
         throw new RuntimeException("Secret content empty!");


cli/src/main/java/keywhiz/cli/commands/AddAction.java
```diff
@@ -139,6 +139,7 @@ private void createAndAssignSecret(String secretName, String description, byte[]
 
   private byte[] readSecretContent() {
     try {
+      System.out.print("Enter secret content, Ctrl+D when done: ");
       byte[] content = ByteStreams.toByteArray(stream);
       if (content.length == 0) {
         throw new RuntimeException("Secret content empty!");
```
mmontgomery-square, line 145: What effect does this have if you pipe the secret in? Is there a way we can only emit this if we're actually waiting for user input?

mweissbacher:
How about expanding documentation instead of printing? Hard to guarantee this would be side-effect free when used by other scripts

cli/src/main/java/keywhiz/cli/commands/AddAction.java
ruiwsquare: Great question, Michael. I tested with following CLI about piping: `echo "xxx" | java -jar "$KEYWHIZ_CLI_JAR" -U "$KEYWHIZ_SERVER_URL" "$@" --devTrustStore add --name secretName secret`. The behavior is that the CLI will finish with success. A better implementation will check whether stream is waiting for user input or data is already available. However, there seems no reliable method to check for the status: https://stackoverflow.com/questions/5826070/java-runtime-process-check-if-waiting-for-input

ruiwsquare:
The problem with expanding document is that the instruction is only useful when creating a secret and the secret content is not yet available in stream. If we provide it anywhere else, it could be hard to be connected with the specific situation. The worst case for the print is that there may be a small confusion when input is piped. The program still function correctly. How about .

mweissbacher (changes requested):
We can't know all situations how this tool is used, it could break compatibility for some users and could require a major version release. The system has been used on our team for years without this use-case being an issue. If you want to add examples to the documentation that works for me, but modifying output is risky. I'm not in favor of accepting this PR in its current form.

ruiwsquare:
Sure. Created a new PR: https://github.com/square/keywhiz/pull/893

PR #893, opened 2021-07-16
Instruction how to provide secret content manually

ruiwsquare:


mweissbacher:
@ruiwsquare the README file contains top-level information about the project, this note might better fit in the --help output of the CLI

ruiwsquare:
Thanks, @mweissbacher. I moved the change to help.

mweissbacher:
Where you added the comments is relevant to secrets but also clients and groups. Those don't have secrets contents so the ctrl+d remark doesn't apply to "add" generally, only to secrets. Another concern is that the lines in this view are broken up below 80 chars, that doesn't seem to be the case with this PR. 

More generally: ctrl+d is standard UNIX way of signaling end of transmission see e.g.: [here](https://en.wikipedia.org/wiki/End-of-Transmission_character), so this is a broader convention that is not specific to Keywhiz. I'm not sure where this could fit in the documentation, but this doesn't seem to be a good spot.

@mmontgomery-square said he will look into where and whether this could fit.

ruiwsquare:
Sure. Thanks for the review. Happy Friday and enjoy the weekend.

PR #904, opened 2021-07-28
Adding secret owner to API

mmontgomery-square:


mmontgomery-square:
Although lengthy, thematically this PR is simple:

* In secret-ish structures, the String property "owner" should come just after the existing "name" property.
* Long method invocations are transposed for readability.
* Tests for serialized structures with and without the new property are added to ensure backwards compatibility.

mmontgomery-square:
Whoops, looks like I ran the API sub-build rather than the full build locally. Hold on...

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -39,6 +39,7 @@ public SecretSeries map(SecretsRecord r) {
     return SecretSeries.of(
         r.getId(),
         r.getName(),
+        null,
         r.getDescription(),
         new ApiDate(r.getCreatedat()),
         r.getCreatedby(),
```
mmontgomery-square, line 45: This is intentionally null. The change to actually update the database schema is separate, and then there'll be a subsequent change to actually start using the new column. This PR is specifically just to update the public API and supporting model classes.

isemaya-square: Got it, so in the future, this will be `r.getOwner()`?

server/src/test/java/keywhiz/commands/DropDeletedSecretsCommandTest.java
```diff
@@ -40,7 +40,7 @@
 
   // an undeleted secret
   private SecretSeries series1 =
-      SecretSeries.of(1, "secret1", "desc1", january18, "creator", january19, "updater", null, null,
+      SecretSeries.of(1, "secret1", null, "desc1", january18, "creator", january19, "updater", null, null,
```
isemaya-square, line 43: Is the owner field set to `null` in these tests until the database schema is updated (as described in a comment above)?

isemaya-square (approved):
Had a question about setting the `owner` field to null in unit tests in the latter half of the PR, but otherwise LGTM.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
mmontgomery-square: Yup!

server/src/test/java/keywhiz/commands/DropDeletedSecretsCommandTest.java
mmontgomery-square: It depends on the test. For most of these the values are irrelevant and were just copied from other tests. Without builder support (which I was debating on adding) it makes it difficult to create mostly-empty POJOs for testing. For integration tests where there's actual calls to a local database then yeah, it totally makes sense to provide a value (assuming it's relevant to the test), but that'll come when we have our updated schema.

PR #915, opened 2021-08-11
Optimizing fetching list of secret names which a group has access to

mmontgomery-square:
GroupResource was unnecessarily fetching secret contents individually when it only needed the names from the secret series

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -529,6 +529,12 @@ public class AclDAO {
         .execute();
   }
 
+  public ImmutableSet<SecretSeries> getSecretSeriesFor(Group group) {
+    checkNotNull(group);
+
+    return dslContext.transactionResult(configuration -> getSecretSeriesFor(configuration, group));
+  }
+
   protected ImmutableSet<SecretSeries> getSecretSeriesFor(Configuration configuration, Group group) {
     List<SecretSeries> r = DSL.using(configuration)
         .select(SECRETS.fields())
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -529,6 +529,12 @@ protected void evictClient(Configuration configuration, long clientId, long grou
         .execute();
   }
 
+  public ImmutableSet<SecretSeries> getSecretSeriesFor(Group group) {
+    checkNotNull(group);
+
+    return dslContext.transactionResult(configuration -> getSecretSeriesFor(configuration, group));
+  }
```
mmontgomery-square, line 536: The API for fetching a group's info needs to get a list of the names of all the secrets which that group has access to. The names live on the series and not the contents, however, there was no API exposed to fetch just the series themselves.

server/src/main/java/keywhiz/service/resources/automation/v2/GroupResource.java
```diff
@@ -141,8 +142,8 @@ public GroupDetailResponseV2 groupInfo(@Auth AutomationClient automationClient,
     Group group = groupDAOReadOnly.getGroup(name)
         .orElseThrow(NotFoundException::new);
 
-    Set<String> secrets = aclDAOReadOnly.getSanitizedSecretsFor(group).stream()
-        .map(SanitizedSecret::name)
+    Set<String> secrets = aclDAOReadOnly.getSecretSeriesFor(group).stream()
+        .map(SecretSeries::name)
         .collect(toSet());
```
mmontgomery-square, line 147: Easy peasy.

server/src/main/java/keywhiz/service/daos/AclDAO.java
drjerry: This all looks good, and I'd hope the Keywhiz maintainers would ask for at least a smoke test added for any new publicly exposed method.

mmontgomery-square: It's covered by the existing test suite for the Group resource, which populates and then queries (via this API) an in-memory database.

drjerry: TIL, thanks.

mbyczkowski (approved):

PR #916, opened 2021-08-12
Unifying Injector creation

mmontgomery-square:
Keywhiz constructs the injector differently between the actual service and tests, and some pieces of the system (commands) manually assemble dependencies. This change implements a unified construction path so that Injector behavior is consistent across the codebase.

The initial motivation for this change is needing to add a new dependency to a command and discovering that it wasn't using Guice at all but rather manual constructor calls.


server/src/main/java/keywhiz/Environments.java
```diff
@@ -0,0 +1,22 @@
...
+    Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+
+    return new Environment(
+        bootstrap.getApplication().getName(),
+        bootstrap.getObjectMapper().copy(),
+        validator,
+        bootstrap.getMetricRegistry(),
+        bootstrap.getClassLoader());
+  }
+}
```
mmontgomery-square, line 22: In DropWizard, it appears that non-server commands have access to Bootstrap but not the full Environment. However, it's easy to construct the latter from the former, as our test runner demonstrates. This just repackages that logic into non-test code for use by commands so that they can make use of the same Injector as the service itself.

server/src/main/java/keywhiz/JooqHealthCheck.java
```diff
@@ -35,7 +35,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(JooqHealthCheck.class);
 
-  enum OnFailure {
+  public enum OnFailure {
     RETURN_UNHEALTHY, LOG_ONLY
   }
```
mmontgomery-square, line 40: This is referenced by a Module in a different package. Although, Keywhiz has a pattern of locating modules in the same folder as their dependencies (e.g. CryptoModule), so perhaps it makes sense to move the module out of keywhiz.inject and into keywhiz.

server/src/main/java/keywhiz/KeywhizService.java
```diff
@@ -119,10 +120,7 @@ public void setInjector(Injector injector) {
   }
 
   private void doRun(KeywhizConfig config, Environment environment) {
-    if (injector == null) {
-      logger.debug("No existing guice injector; creating new one");
-      injector = Guice.createInjector(new ServiceModule(config, environment));
-    }
+    ensureInjectorCreated(config, environment);
```
mmontgomery-square, line 124: I wanted to be able to test this behavior in isolation without all the other side effects of running the service, so I moved it into a helper method.

server/src/main/java/keywhiz/ServiceModule.java
```diff
@@ -62,18 +66,18 @@ public ServiceModule(KeywhizConfig config, Environment environment) {
...
-    // TODO(justin): Consider https://github.com/HubSpot/dropwizard-guice.
-    bind(Environment.class).toInstance(environment);
-    bind(Configuration.class).toInstance(config);
-    bind(KeywhizConfig.class).toInstance(config);
+    install(new ContextModule(config, environment));
+    install(new CookieModule(config.getCookieKey()));
+    install(new CryptoModule(config.getDerivationProviderClass(), config.getContentKeyStore()));
+    install(new DaoModule());
+    install(new StrictGuiceModule());
   }
```
mmontgomery-square, line 81: This is mostly just organization:

1. All bind() statements are in a single block, likewise install()s
2. Bindings for Environment, Configuration, and KeywhizConfig are moved to a dedicated module (ContextModule)
3. Dao-related bindings are in their own module
4. In order to circumvent some subtle bugs that came up in testing (Guice will, by default, instantiation class with no-arg constructors, which masked an issue where KeywhizConfig wasn't actually loading properly), we've turned on some stricter binding settings which are located inside StrictGuiceModule. As a consequence, classes with no-argument constructors that we want injected (ClientCertificateFilter, SessionMeResource, etc.) need explicit bindings.

server/src/main/java/keywhiz/ServiceModule.java
```diff
@@ -82,32 +86,6 @@ public ServiceModule(KeywhizConfig config, Environment environment) {
...
-      KeywhizConfig config) {
-    DataSourceFactory dataSourceFactory = config.getReadonlyDataSourceFactory();
-    ManagedDataSource dataSource = dataSourceFactory.build(environment.metrics(), "db-readonly");
-    environment.lifecycle().manage(dataSource);
-
-    environment.healthChecks().register("db-readonly-health",
-        new JooqHealthCheck(dataSource, RETURN_UNHEALTHY));
-
-    return dataSource;
-  }
```
mmontgomery-square, line 109: The only thing that differs from the main service (DropWizard's built-in "server" command) and our hand-rolled commands is that they define their data sources differently (different names, and only the service data sources have health checks). In order to seamlessly support both use cases we extract the service-specific data source bindings into a dedicated module so that they're only bound on the service code path.

server/src/main/java/keywhiz/commands/AddUserCommand.java
```diff
@@ -33,15 +35,25 @@ public AddUserCommand() {
...
+    ManagedDataSource dataSource = config.getDataSourceFactory()
+        .build(new MetricRegistry(), "add-user-datasource");
+
+    Injector injector = InjectorFactory.createInjector(
+        config,
+        Environments.fromBootstrap(bootstrap),
+        dataSource);
+
+    return injector.getInstance(UserDAO.class);
   }
```
mmontgomery-square, line 58: Rather than assembling dependencies by hand, we now take advantage of the same Injector that the main service uses (data source bindings aside) via the InjectorFactory.

server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
```diff
@@ -142,35 +138,18 @@ public DropDeletedSecretsCommand(SecretDAO secretDAO) {
...
-        new SecretSeriesMapper(bootstrap.getObjectMapper()),
-        null
-    );
-
-    return new SecretDAO(
-        dslContext,
-        secretContentDAOFactory,
-        secretSeriesDAOFactory,
-        null
-    );
```
mmontgomery-square, line 173: Icky manual injection replaced with Guice.

server/src/main/java/keywhiz/inject/ContextModule.java
```diff
@@ -0,0 +1,23 @@
...
+    this.config = config;
+    this.environment = environment;
+  }
+
+  @Override protected void configure() {
+    // TODO(justin): Consider https://github.com/HubSpot/dropwizard-guice.
+    bind(Environment.class).toInstance(environment);
+    bind(Configuration.class).toInstance(config);
+    bind(KeywhizConfig.class).toInstance(config);
+  }
```
mmontgomery-square, line 22: Small, single-focus modules are easier to reason about, compose, and test.

server/src/main/java/keywhiz/inject/DataSourceModule.java
```diff
@@ -0,0 +1,20 @@
...
+  public DataSourceModule(ManagedDataSource readWrite, ManagedDataSource readonly) {
+    this.readWrite = readWrite;
+    this.readonly = readonly;
+  }
+
+  @Override protected void configure() {
+    bind(ManagedDataSource.class).toInstance(readWrite);
+    bind(ManagedDataSource.class).annotatedWith(Readonly.class).toInstance(readonly);
+  }
+}
```
mmontgomery-square, line 20: This module is used by non-server commands to provide data source bindings.

Technically, with some extra coding on the service's run method (installation of health checks outside of modules) we could use this for both server and non-server commands.

server/src/main/java/keywhiz/inject/InjectorFactory.java
```diff
@@ -0,0 +1,24 @@
...
+        new ServiceDataSourceModule(),
+        new ServiceModule(config, environment));
+  }
+
+  public static Injector createInjector(KeywhizConfig config, Environment environment, ManagedDataSource dataSource) {
+    return Guice.createInjector(
+        new DataSourceModule(dataSource, dataSource),
+        new ServiceModule(config, environment));
+  }
+}
```
mmontgomery-square, line 24: Our handy factory class is used by the service ("server" command), other commands, and by tests. This ensures that our construction is unified for all use cases.

As mentioned earlier, we could get rid of the overload by doing a little more work inside the KeywhizService doRun() method as far as assembling health checks, eliminating the need for ServiceDataSourceModule.

server/src/main/java/keywhiz/inject/ServiceDataSourceModule.java
```diff
@@ -0,0 +1,40 @@
...
+      KeywhizConfig config) {
+    DataSourceFactory dataSourceFactory = config.getReadonlyDataSourceFactory();
+    ManagedDataSource dataSource = dataSourceFactory.build(environment.metrics(), "db-readonly");
+    environment.lifecycle().manage(dataSource);
+
+    environment.healthChecks().register("db-readonly-health",
+        new JooqHealthCheck(dataSource, RETURN_UNHEALTHY));
+
+    return dataSource;
+  }
```
mmontgomery-square, line 39: This module is responsible for the service's data source bindings. If we performed the lifecycle and health check logic inside KeywhizService.doRun(), this module would go away. I should note that, per Guice best practices, these kinds of actions (initialization, lifecycle management, etc.) should not be performed within modules anyway, they're just for side-effect-free wiring.

server/src/main/java/keywhiz/inject/StrictGuiceModule.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.google.inject.AbstractModule;
+
+public class StrictGuiceModule extends AbstractModule {
+  @Override protected void configure() {
+    binder().disableCircularProxies();
+    binder().requireAtInjectOnConstructors();
+    binder().requireExactBindingAnnotations();
+  }
+}
```
mmontgomery-square, line 11: Here's what these do:

1. Disallow circular references in our dependency graph
2. Disallow using no-argument constructors for instantiation unless explicitly requested (eliminates an annoying class of subtle bugs)
3. Requires strict matching of binding annotations, addressing a bug in Guice's original design which is kept around for backwards compatibility

server/src/main/java/keywhiz/service/crypto/CryptoModule.java
```diff
@@ -54,7 +55,9 @@ public CryptoModule(String derivationProviderClass, KeyStoreConfig keyStoreConfi
     this.keyStoreConfig = keyStoreConfig;
   }
 
-  @Override protected void configure() {}
+  @Override protected void configure() {
+    bind(SecureRandom.class).toProvider(SecureRandom::new);
+  }
```
mmontgomery-square, line 60: Previously, Guice would silently create a new SecureRandom whenever needed because it has a no-argument constructor. While helpful, this can mask silent bugs, one of which I encountered while putting this PR together, so it's not just an academic concern. Now we have to explicitly tell Guice that we want to create new SecureRandoms via the no-arg constructor.

server/src/main/java/keywhiz/service/daos/ClientMapper.java
```diff
@@ -16,6 +16,7 @@
 
 package keywhiz.service.daos;
 
+import com.google.inject.Inject;
 import java.util.Optional;
 import keywhiz.api.ApiDate;
 import keywhiz.api.model.Client;
```
mmontgomery-square, line 22: Whoops, obviously this can go away.

server/src/main/java/keywhiz/service/daos/DaoModule.java
```diff
@@ -0,0 +1,9 @@
+package keywhiz.service.daos;
+
+import com.google.inject.AbstractModule;
+
+public class DaoModule extends AbstractModule {
+  @Override protected void configure() {
+    bind(ClientMapper.class).toProvider(ClientMapper::new);
+  }
+}
```
mmontgomery-square, line 9: ClientMapper is package-private and has an implicit constructor. Rather than increase visibility or add @Inject to the constructor, and to support testing, I created a same-package module that explicitly binds the constructor.

server/src/main/java/keywhiz/service/daos/UserDAO.java
```diff
@@ -27,7 +28,7 @@
 public class UserDAO {
   private final DSLContext dslContext;
 
-  public UserDAO(DSLContext dslContext) {
+  @Inject public UserDAO(DSLContext dslContext) {
     this.dslContext = checkNotNull(dslContext);
   }
```
mmontgomery-square, line 33: This was being manually constructed, now it's injected.

server/src/test/java/keywhiz/KeywhizTestRunner.java
```diff
@@ -49,35 +40,7 @@ public KeywhizTestRunner(Class<?> klass) throws InitializationError {
...
-        objectMapper,
-        validator,
-        bootstrap.getMetricRegistry(),
-        bootstrap.getClassLoader());
-
-    Injector injector = Guice.createInjector(new ServiceModule(config, environment));
-
-    service.setInjector(injector);
-    return injector;
-  }
```
mmontgomery-square, line 80: This logic is now stored in a dedicated class since it's used in multiple places now and is no longer test runner-specific.

server/src/test/java/keywhiz/KeywhizTestRunner.java
```diff
@@ -104,8 +67,9 @@ static Injector createInjector() {
       jooqContext.truncate(Secrets.SECRETS).execute();
     } catch(DataAccessException e) {}
 
-    Object object = injector.getInstance(getTestClass().getJavaClass());
-    MockitoAnnotations.initMocks(object);
-    return object;
+    Object test = getTestClass().getJavaClass().getDeclaredConstructor().newInstance();
+    injector.injectMembers(test);
+    MockitoAnnotations.initMocks(test);
+    return test;
```
mmontgomery-square, line 73: Now that we have strict bindings, Guice no longer allows us to request instances of our test classes because their default constructors aren't marked with @Inject. Rather than modify every test class, we just modify how we create and inject them.

server/src/test/java/keywhiz/inject/DataSourceModuleTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class DataSourceModuleTest {
+  @Test
+  public void injectsDataSources() {
+    class Holder {
+      @Inject ManagedDataSource readWrite;
+      @Inject @Readonly ManagedDataSource readonly;
+    }
+
+    Holder holder = new Holder();
```
mmontgomery-square, line 21: Anonymous inner classes (one of Java's lesser-known features) are a fantastic way to test dependency injection in a convincing way: you can declare fields with various annotations exactly as you would in your regular classes.

server/src/test/java/keywhiz/test/ServiceContext.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.validation.Validation;
+import javax.validation.Validator;
+import keywhiz.KeywhizConfig;
+import keywhiz.KeywhizService;
+
+public class ServiceContext {
+  private final KeywhizService service;
+  private final Bootstrap<KeywhizConfig> bootstrap;
+  private final KeywhizConfig config;
+  private final Environment environment;
```
mmontgomery-square, line 21: For testing, these four classes are often required and their construction is intertwined. Encapsulating them all together in a single class made creation and consumption much easier and reduced boilerplate scattered about the codebase.

server/src/test/java/keywhiz/inject/DataSourceModuleTest.java
isemaya-square: TIL!

server/src/test/java/keywhiz/service/crypto/CryptoModuleTest.java
```diff
@@ -0,0 +1,39 @@
...
+public class CryptoModuleTest {
+  @Test
+  public void secureRandomIsExplicitlyBound() {
+    KeywhizConfig config = ServiceContext.create().getConfig();
+
+    Injector injector = Guice.createInjector(
+        new CryptoModule(
+            config.getDerivationProviderClass(),
+            config.getContentKeyStore()),
+        new StrictGuiceModule());
```
isemaya-square, line 23: How did you know to include `StrictGuiceModule` here?

server/src/test/java/keywhiz/service/crypto/CryptoModuleTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    assertNotNull(injector.getInstance(SecureRandom.class));
+  }
+
+  @Test
+  public void secureRandomIsNonSingleton() {
```
isemaya-square, line 29: Is this so that `SecureRandom` starts with a different seed every time?

isemaya-square (approved):
Had some questions about how you knew what to include in your tests (and wrote some questions out of curiosity), but didn't spot any red flags.

server/src/main/java/keywhiz/inject/ContextModule.java
mmontgomery-square: I kept it just in case. My guess is that whatever that link points to is some sort of first-class Guice-DropWizard integration where we wouldn't have to make these bindings ourselves, and while that does sound like an improvement, I didn't want to pursue it but I also didn't want to lose track of it. Of course, the right thing to do would be to create an internal task and remove the comment, so, maybe I'll just do that.

server/src/main/java/keywhiz/inject/StrictGuiceModule.java
mmontgomery-square: Yeah, I had a problem where there was a missing binding/logic for loading KeywhizConfig from the yaml file. Because KeywhizConfig is just a data blob with a no-argument constructor Guice happily created a new (empty) one for me, which caused all sorts of bizarre error messages. I've worked extensively with Guice in the past so I knew what to look for, but it's the kind of thing that can leave you running in circles. I decided that it would be best to not have to deal with those sorts of problems entirely at the cost of having to specify 2-3 additional bindings.

server/src/main/java/keywhiz/inject/InjectorFactory.java
```diff
@@ -0,0 +1,25 @@
...
+import keywhiz.KeywhizConfig;
+import keywhiz.ServiceDataSourceModule;
+import keywhiz.ServiceModule;
+
+public final class InjectorFactory {
+  private InjectorFactory() {}
+
+  public static Injector createInjector(KeywhizConfig config, Environment environment) {
+    return Guice.createInjector(
+        new ServiceDataSourceModule(),
```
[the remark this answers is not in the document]
mmontgomery-square: It just means that the module will get references to dependencies "indirectly" via Guice. Modules are a lot less magical than they seem.

So, let's say you have some provider method in your module:

```
@Provides Bar bar(Foo foo) {
  return new Bar(foo);
}
```

Guice will locate the binding (instructions on how to create an instance) for Foo, spine one up, and pass it into this method as an argument. A different but mostly-equivalent style would be:

```
Module MyModule extends AbstractModule {
  private final Foo foo;

  public MyModule(Foo foo) {
    this.foo = foo;
  }

  @Provides Bar bar() {
    return new Bar(foo);
  }
}
```

So, instead of having Guice create the Foo instance on your behalf, you're directly supplying an instance yourself. The downside here is that the Foo is just a field, so it's effectively a singleton. If you want some other pattern (every new Bar also gets a new Foo), the first style is more flexible because it leaves that decision to someone else.

Generally, I would always recommend the first approach and reserve passing in constructor arguments for things which cannot be created by Guice (in our case, DropWizard-ish configuration), usually because they're instantiated by a framework before the Injector is ever created.

server/src/test/java/keywhiz/EnvironmentsTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class EnvironmentsTest {
+  @Test
+  public void createsEnvironmentFromBootstrap() {
+    KeywhizService keywhiz = new KeywhizService();
+    Bootstrap<KeywhizConfig> bootstrap = new Bootstrap<>(keywhiz);
+    Environment environment = Environments.fromBootstrap(bootstrap);
+
+    assertEquals(keywhiz.getName(), environment.getName());
+    assertNotSame(bootstrap.getObjectMapper(), environment.getObjectMapper());
```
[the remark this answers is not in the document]
mmontgomery-square: I actually don't know for certain. In the original code it creates a copy of the ObjectMapper. My guess would be that they want a second instance with the same configuration as the first but they need it to be separate due to threading etc. and using the copy() method was the easiest way to achieve this.

server/src/test/java/keywhiz/service/crypto/CryptoModuleTest.java
mmontgomery-square: I needed to verify that, with the restrictions turned on, we could still inject instances of SecureRandom. Basically, I wanted the Injector in this test to faithfully reproduce the behavior of the service's actual Injector.

server/src/test/java/keywhiz/service/crypto/CryptoModuleTest.java
mmontgomery-square: This is actually mostly for documentation. Because we didn't specify any binding for SecureRandom, and because it has a no-argument constructor, Guice was creating a new instance every time. Now, if this new instance was then given to a singleton then it's also effectively a singleton (Guice will never need to create a second instance) but in any case the behavior was implicit. Now, it's explicit, and this test proves that. If we actually wanted the SecureRandom to be a singleton then we'll need to modify the code accordingly, but it was never a singleton to begin with.

PR #919, opened 2021-08-16
Persist secret owner

mmontgomery-square:


api/src/main/java/keywhiz/api/model/SecretContent.java
```diff
@@ -31,12 +31,30 @@
...
+      String encryptedContent,
+      String hmac,
+      ApiDate createdAt,
+      @Nullable String createdBy,
+      ApiDate updatedAt,
+      @Nullable String updatedBy,
+      ImmutableMap<String, String> metadata,
+      long expiry) {
+
+    return builder()
```
mmontgomery-square, line 46: For convenience, especially in constructing test data, this class now supports a builder. There are no additional fields.

api/src/main/java/keywhiz/api/model/SecretSeries.java
```diff
@@ -73,4 +76,27 @@ public static SecretSeries of(
   public abstract Optional<String> type();
   public abstract ImmutableMap<String, String> generationOptions();
   public abstract Optional<Long> currentVersion();
+
+  public static Builder builder() {
+    return new AutoValue_SecretSeries.Builder();
+  }
+
+  public abstract Builder toBuilder();
```
mmontgomery-square, line 84: This class also gains builder support.

server/src/main/java/keywhiz/commands/DropDeletedSecretsCommand.java
```diff
@@ -43,7 +43,7 @@ public DropDeletedSecretsCommand() {
 
   @VisibleForTesting
   public DropDeletedSecretsCommand(SecretDAO secretDAO) {
-    super("drop-deleted-secrets", "PERMANENTLY REMOVES database records for deleted secrets");
+    this();
     this.secretDAO = secretDAO;
   }
```
mmontgomery-square, line 48: This change is unrelated to the rest of the PR, but I noticed that the constructors were replicating behavior as far as calling the super class with the same string constants. By having one constructor lean on (invoke) the other we no longer need to duplicate the string values.

server/src/main/java/keywhiz/service/config/Readwrite.java
```diff
@@ -0,0 +1,9 @@
+package keywhiz.service.config;
+
+import com.google.inject.BindingAnnotation;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@BindingAnnotation
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Readwrite {}
```
mmontgomery-square, line 9: We already have an annotation for read-only bindings, adding a read-write annotation helps simplify code in that we can directly inject a read-write instance rather than injecting a factory and invoking the readwrite() method on it.

server/src/main/java/keywhiz/service/daos/DaoModule.java
```diff
@@ -1,9 +1,48 @@
...
+  GroupDAO readonlyGroupDAO(GroupDAO.GroupDAOFactory factory) {
+    return factory.readonly();
+  }
+
+  @Provides
+  @Singleton
+  @Readwrite
+  GroupDAO readwriteGroupDAO(GroupDAO.GroupDAOFactory factory) {
+    return factory.readwrite();
+  }
```
mmontgomery-square, line 26: These direct bindings allow for simplification and hide the factories from our code and tests.

server/src/main/java/keywhiz/service/daos/SecretController.java
```diff
@@ -309,13 +320,31 @@ public SecretBuilder withType(String type) {
...
+            name,
+            owner,
+            encryptedSecret,
+            hmac,
+            creator,
+            metadata,
+            expiry,
+            description,
+            type,
             generationOptions);
```
mmontgomery-square, line 333: The owner field here is new.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -78,18 +79,32 @@
...
+      SecretContentDAOFactory secretContentDAOFactory,
+      SecretSeriesDAOFactory secretSeriesDAOFactory,
+      GroupDAO.GroupDAOFactory groupDAOFactory,
+      ContentCryptographer cryptographer) {
     this.dslContext = dslContext;
     this.secretContentDAOFactory = secretContentDAOFactory;
     this.secretSeriesDAOFactory = secretSeriesDAOFactory;
+    this.groupDAOFactory = groupDAOFactory;
     this.cryptographer = cryptographer;
   }
```
mmontgomery-square, line 93: The dependency on the Group DAO is new because now Secrets care about owners (Groups).

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -109,6 +124,8 @@ public long createSecret(String name, String encryptedSecret, String hmac,
       SecretContentDAO secretContentDAO = secretContentDAOFactory.using(configuration);
       SecretSeriesDAO secretSeriesDAO = secretSeriesDAOFactory.using(configuration);
 
+      Long ownerId = getOwnerId(configuration, owner);
+
       Optional<SecretSeries> secretSeries = secretSeriesDAO.getSecretSeriesByName(name);
       long secretId;
       if (secretSeries.isPresent()) {
```
mmontgomery-square, line 131: Our API deals with owners via their names, but internally we need to record the ID (a long), so we introduce a lookup method to perform that translation.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -575,36 +614,60 @@ public void dangerPermanentlyRemoveSecretsDeletedBeforeDate(DateTime deletedBefo
     }
   }
 
+  private Long getOwnerId(Configuration configuration, String owner) {
+    GroupDAO groupDAO = groupDAOFactory.using(configuration);
+    Optional<Group> group = groupDAO.getGroup(owner);
+    return group.isPresent() ? group.get().getId() : null;
+  }
```
mmontgomery-square, line 621: Oops, I think this is a bug: if the caller specifies an owner which doesn't exist, we should fail rather than silently succeed.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -20,26 +20,36 @@
...
-  @Inject public SecretSeriesMapper(ObjectMapper mapper) {
+  @Inject public SecretSeriesMapper(
+      ObjectMapper mapper,
+      @Readonly GroupDAO groupDAO) {
     this.mapper = mapper;
+    this.groupDAO = groupDAO;
   }
 
   public SecretSeries map(SecretsRecord r) {
+    String owner = getOwner(r);
```
mmontgomery-square, line 47: Design-wise, this was the crux of the change: where and how should we perform the translation from the database's model of ownership, which is a long foreign key to the group table, to the API model of ownership which is the group's name? Fortunately, that's exactly the job of this class, at least on the read path. However, it did require injecting a dependency that we didn't have before, which was the original motivator for cleaning up the DI code in a previous change.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -20,26 +20,36 @@
...
+
     return SecretSeries.of(
         r.getId(),
         r.getName(),
-        null,
+        owner,
         r.getDescription(),
         new ApiDate(r.getCreatedat()),
         r.getCreatedby(),
```
mmontgomery-square, line 55: Concretely, this is the major behavior change on the read path: if the database contains a value for the owner it will now be represented in our API model.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -50,6 +60,25 @@ public SecretSeries map(SecretsRecord r) {
...
+      throw new IllegalStateException(
+          String.format(
+              "Unable to find owner for secret [%s] (ID %s): group ID %s not found",
+              r.getName(),
+              r.getId(),
+              ownerId));
+    }
+
+    return maybeGroup.get().getName();
+  }
```
mmontgomery-square, line 80: We support missing (null) owners, but we don't support references to owners which don't exist. Normally we would enforce this at the database level via a foreign key constraint but our testing setup's use of truncate precludes this (and I haven't had time to dig in to find a better solution).

server/src/test/java/keywhiz/KeywhizRule.java
```diff
@@ -0,0 +1,21 @@
...
+  public KeywhizRule(Object test) {
+    this.test = test;
+  }
+
+  @Override public Statement apply(Statement statement, Description description) {
+    Injector injector = KeywhizTests.createInjector();
+    injector.injectMembers(test);
+    return statement;
+  }
+}
```
mmontgomery-square, line 21: Oops, I wrote this but never ended up using it. Will delete.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
isemaya-square: Fantastic - otherwise, having to provide a group id for the owner (rather than the name) in the API would be confusing/painful. Nice work!

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -50,6 +60,25 @@ public SecretSeries map(SecretsRecord r) {
         r.getCurrent());
   }
 
+  private String getOwner(SecretsRecord r) {
```
isemaya-square, line 63: Might be worth being specific here (and in general in this PR) that this is returning the owner name, and NOT the owner id. So renaming the method to`getOwnerName` would save some mental processing for the reader and clarify that we're getting the group name from the secrets record, which contains an owner id.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
isemaya-square, line 47: Similar to comment below, changing this to `String ownerName = getOwnerName(r)` makes it clearer that we are using `groupDAO` to find the owner name from the owner id.

isemaya-square (commented):
Might be worth considering renaming instances of  `owner` to either `ownerId` or `ownerName`, depending on the context, both in methods and in unit tests.

I think that would be helpful in making it clear when the api model is used and when the database model is used.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
mmontgomery-square: Right, for backwards compatibility we allow owners to be null since we can't atomically update Keywhiz, the Keywhiz database, and Secrets Authority together. However, going forward it may make sense to lock that down since it won't make sense to create secrets without owners (I think), at least from the Console/Secrets Authority path (note to self: I should figure out if there are other secret creation paths which we'll need to support and what changes may be necessary there [maybe add a counter every time a secret is created without an owner?]). But, we also *do not* allow you to specify an owner which doesn't exist. I *think* it makes sense for the caller to ensure that the owner exists beforehand by making the appropriate group fetch/create calls rather than have the secret creation API silently create a group if none exists (also the problem with that option is that we won't have all the fields necessary to properly create the group if we're limited to whatever the secret creation API accepts as parameters).

server/src/test/java/keywhiz/service/daos/SecretControllerTest.java
```diff
@@ -77,6 +77,116 @@
...
+        .withOwner(owner)
+        .create();
+
+    assertEquals(owner, created.getOwner());
+  }
+
+  @Test
+  public void createsOrUpdatesNewSecretWithOwner() {
+    String owner = UUID.randomUUID().toString();
+
```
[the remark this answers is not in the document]
mmontgomery-square: Which space? Not sure what you're referring to.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
mmontgomery-square: Great suggestion, I'll see what I can do to distinguish the two in the code.

server/src/test/java/keywhiz/service/daos/SecretControllerTest.java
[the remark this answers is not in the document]
isemaya-square: Oh hmm, I was looking at how in the previous test, there was no space between `ownerName` and `groupDAO.createGroup`, but there was in this case. Very minor nit!

server/src/test/java/keywhiz/service/daos/SecretControllerTest.java
```diff
@@ -79,112 +79,112 @@
 
   @Test
   public void createsSecretWithOwner() {
-    String owner = UUID.randomUUID().toString();
-    groupDAO.createGroup(owner, "creator", "description", ImmutableMap.of());
+    String ownerName = UUID.randomUUID().toString();
+    groupDAO.createGroup(ownerName, "creator", "description", ImmutableMap.of());
```
isemaya-square, line 83: I wonder if it makes sense to overload the `createGroup` method with just three args if the fourth `metadata` map is going to be empty often like this (or would this just be for testing?)

isemaya-square (approved):

api/src/main/java/keywhiz/api/model/SecretSeries.java
```diff
@@ -46,20 +46,23 @@ public static SecretSeries of(
...
-        updatedAt,
-        nullToEmpty(updatedBy),
-        Optional.ofNullable(type),
-        options,
-        Optional.ofNullable(currentVersion));
+
+    return builder()
+        .id(id)
+        .name(name)
+        .owner(owner)
```
isemaya-square, line 56: We want to keep this as "owner" because this is part of the external facing API, right?

server/src/test/java/keywhiz/service/daos/SecretControllerTest.java
mmontgomery-square: Typically what I'll do for tests is have some helper method that takes in just the one or two parameters I care about and then defaults the others, but I just didn't do that here. Maybe there wasn't enough duplication to make it worthwhile (yet).

api/src/main/java/keywhiz/api/model/SecretSeries.java
mmontgomery-square: Yeah, and I think that makes sense because the owner name and ID are only distinct internally because of the database model, externally there is only the owner name (the internal data ID is never exposed) so leaving it as just owner makes more sense. Although, we could still call it ownerName, but we'd have to change some APIs (not the end of the world, just more work).

isemaya-square: Yeah, that makes sense. If the API only uses the owner name, then we can use "owner" as a shorthand without causing any confusion.

PR #931, opened 2021-08-30
Log owner during secret creation

mmontgomery-square:
We already log the owner in the audit logs, but we should put it in the regular logs as well

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
+++ b/server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
@@ -211,7 +211,16 @@ public class SecretsResource {
   @Consumes(APPLICATION_JSON)
   public Response createSecret(@Auth User user, @Valid CreateSecretRequestV2 request) {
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -211,7 +211,16 @@ protected SanitizedSecret retrieveSecret(@Auth User user, String name) {
...
+    {
+      String ownerPart = request.owner() == null
+          ? "with no owner"
+          : String.format("with owner '%s'", request.owner());
+      logger.info(
+          "User '{}' creating secret '{}' {}.",
+          user,
+          request.name(),
+          ownerPart);
+    }
```
mmontgomery-square, line 223: Maybe I'm overthinking it, but I didn't want a missing owner to be represented by 'null' in the output, since that looks a lot like a string containing the word null and not a null value.

isemaya-square (approved):

PR #932, opened 2021-09-01
Fixing two other locations where owner was not being passed into secr…

mmontgomery-square:
…et builder

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/AclDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/AclDAO.java
@@ -578,17 +578,24 @@ public class AclDAO {
     private final SecretSeriesDAOFactory secretSeriesDAOFactory;
     private final ClientMapper clientMapper;
     private final GroupMapper groupMapper;
-    private final SecretSeriesMapper secretSeriesMapper;
+    private final SecretSeriesMapper.SecretSeriesMapperFactory secretSeriesMapperFactory;
     private final SecretContentMapper secretContentMapper;
     private final RowHmacGenerator rowHmacGenerator;
     private final KeywhizConfig config;
 
-    @Inject public AclDAOFactory(DSLContext jooq, @Readonly DSLContext readonlyJooq, ClientDAOFactory clientDAOFactory,
-                                 GroupDAOFactory groupDAOFactory, SecretContentDAOFactory secretContentDAOFactory,
-                                 SecretSeriesDAOFactory secretSeriesDAOFactory, ClientMapper clientMapper,
-                                 GroupMapper groupMapper, SecretSeriesMapper secretSeriesMapper,
-                                 SecretContentMapper secretContentMapper, RowHmacGenerator rowHmacGenerator,
-                                 KeywhizConfig config) {
+    @Inject public AclDAOFactory(
+        DSLContext jooq,
+        @Readonly DSLContext readonlyJooq,
+        ClientDAOFactory clientDAOFactory,
+        GroupDAOFactory groupDAOFactory,
+        SecretContentDAOFactory secretContentDAOFactory,
+        SecretSeriesDAOFactory secretSeriesDAOFactory,
+        ClientMapper clientMapper,
+        GroupMapper groupMapper,
+        SecretSeriesMapper.SecretSeriesMapperFactory secretSeriesMapperFactory,
+        SecretContentMapper secretContentMapper,
+        RowHmacGenerator rowHmacGenerator,
+        KeywhizConfig config) {
       this.jooq = jooq;
       this.readonlyJooq = readonlyJooq;
       this.clientDAOFactory = clientDAOFactory;
@@ -597,29 +604,56 @@ public class AclDAO {
       this.secretSeriesDAOFactory = secretSeriesDAOFactory;
       this.clientMapper = clientMapper;
       this.groupMapper = groupMapper;
-      this.secretSeriesMapper = secretSeriesMapper;
+      this.secretSeriesMapperFactory = secretSeriesMapperFactory;
       this.secretContentMapper = secretContentMapper;
       this.rowHmacGenerator = rowHmacGenerator;
       this.config = config;
     }
 
     @Override public AclDAO readwrite() {
-      return new AclDAO(jooq, clientDAOFactory, groupDAOFactory, secretContentDAOFactory,
-          secretSeriesDAOFactory, clientMapper, groupMapper, secretSeriesMapper, secretContentMapper,
-          rowHmacGenerator, config);
+      return new AclDAO(
+          jooq,
+          clientDAOFactory,
+          groupDAOFactory,
+          secretContentDAOFactory,
+          secretSeriesDAOFactory,
+          clientMapper,
+          groupMapper,
+          secretSeriesMapperFactory.using(jooq),
+          secretContentMapper,
+          rowHmacGenerator,
+          config);
     }
 
     @Override public AclDAO readonly() {
-      return new AclDAO(readonlyJooq, clientDAOFactory, groupDAOFactory, secretContentDAOFactory,
-          secretSeriesDAOFactory, clientMapper, groupMapper, secretSeriesMapper, secretContentMapper,
-          rowHmacGenerator, config);
+      return new AclDAO(
+          readonlyJooq,
+          clientDAOFactory,
+          groupDAOFactory,
+          secretContentDAOFactory,
+          secretSeriesDAOFactory,
+          clientMapper,
+          groupMapper,
+          secretSeriesMapperFactory.using(readonlyJooq),
+          secretContentMapper,
+          rowHmacGenerator,
+          config);
     }
 
     @Override public AclDAO using(Configuration configuration) {
       DSLContext dslContext = DSL.using(checkNotNull(configuration));
-      return new AclDAO(dslContext, clientDAOFactory, groupDAOFactory, secretContentDAOFactory,
-          secretSeriesDAOFactory, clientMapper, groupMapper, secretSeriesMapper, secretContentMapper,
-          rowHmacGenerator, config);
+      return new AclDAO(
+          dslContext,
+          clientDAOFactory,
+          groupDAOFactory,
+          secretContentDAOFactory,
+          secretSeriesDAOFactory,
+          clientMapper,
+          groupMapper,
+          secretSeriesMapperFactory.using(dslContext),
+          secretContentMapper,
+          rowHmacGenerator,
+          config);
     }
   }
 }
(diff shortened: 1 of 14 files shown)


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -388,30 +390,48 @@ public long dangerPermanentlyRemoveRecordsForGivenIDs(List<Long> ids) {
...
     }
 
     @Override public SecretSeriesDAO readwrite() {
-      return new SecretSeriesDAO(jooq, objectMapper, secretSeriesMapper, rowHmacGenerator);
+      return new SecretSeriesDAO(
+          jooq,
+          objectMapper,
+          secretSeriesMapperFactory.using(jooq),
+          rowHmacGenerator);
     }
```
mmontgomery-square, line 418: Integration testing revealed a subtle bug in some previous changes I'd made regarding automatically injecting a read-only instance of a GroupDAO into the SecretSeriesMapper. The symptom was that integration tests were timing out waiting for a database connection, and the integration tests run using a connection pool with a single connection. Specifically, they were timing out inside the SecretSeriesMapper while trying to look up the owner for secrets which had an owner. The root cause was that, in Jooq, a context (DSLContext) is roughly equivalent to a database connection, and every database query within the same context will share the same connection, more or less. By injecting a read-only GroupDAO up-front into the single, global SecretSeriesMapper we were essentially forcing Jooq to perform the owner lookup on a connection other than the one that was currently being used to look up the secrets. By introducing a factory for the SecretSeriesMapper we can scope it such that it shares the same context, and thus the same connection, of whatever query is currently executing.

Speaking of scoping, a lot (all?) of the DAO factory boilerplate would go away if we adopted Guice's child injectors. Basically, on every high-level database operation, create a child injector which binds DSLContext to the current context, and then inject DAOs etc. willy-nilly. I'm going to pursue this route in a follow-up PR but I didn't want to overcomplicate what is already a decent-sized change.

server/src/main/java/keywhiz/service/daos/SecretSeriesMapper.java
```diff
@@ -36,9 +36,9 @@
...
   private final GroupDAO groupDAO;
 
-  @Inject public SecretSeriesMapper(
+  public SecretSeriesMapper(
       ObjectMapper mapper,
-      @Readonly GroupDAO groupDAO) {
+      GroupDAO groupDAO) {
     this.mapper = mapper;
     this.groupDAO = groupDAO;
   }
```
mmontgomery-square, line 44: Rather than always using the read-only context, the injected GroupDAO re-uses the current DSLContext.

server/src/main/java/keywhiz/service/resources/automation/AutomationSecretResource.java
```diff
@@ -113,9 +113,14 @@ public AutomationSecretResponse createSecret(
...
-        automationClient.getName(), request.expiry)
-        .withDescription(nullToEmpty(request.description));
+    SecretController.SecretBuilder builder = secretController
+        .builder(
+            request.name,
+            request.content,
+            automationClient.getName(),
+            request.expiry)
+        .withDescription(nullToEmpty(request.description))
+        .withOwnerName(request.owner);
```
mmontgomery-square, line 123: This was one of the bugs: we weren't passing in the owner to the builder.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -123,9 +123,14 @@ public Response createSecret(@Auth AutomationClient automationClient,
...
-        .builder(name, request.content(), automationClient.getName(), request.expiry())
+        .builder(
+            name,
+            request.content(),
+            automationClient.getName(),
+            request.expiry())
         .withDescription(request.description())
         .withMetadata(request.metadata())
+        .withOwnerName(request.owner())
         .withType(request.type());
```
mmontgomery-square, line 134: This was the other bug: again, not passing in the owner to the builder.

server/src/test/java/keywhiz/service/daos/DaoModuleTest.java
```diff
@@ -100,11 +102,13 @@ public void injectsReadwriteSecretContentDAO() {
...
-    when(groupDAOFactory.readonly()).thenReturn(dao);
+  public void doesNotInjectSecretSeriesMapper() {
+    assertThrows(ConfigurationException.class, () -> getInstance(SecretSeriesMapper.class));
+  }
 
-    assertNotNull(getInstance(SecretSeriesMapper.class));
+  @Test
+  public void injectsSecretSeriesMapperFactory() {
+    assertNotNull(getInstance(SecretSeriesMapper.SecretSeriesMapperFactory.class));
   }
```
mmontgomery-square, line 112: Because the SecretSeriesMapper must be scoped to a DSLContext we can no longer inject it directly. However, we can directly inject the factory. This gives it similar behavior to existing DAOs.

server/src/test/java/keywhiz/service/daos/SecretSeriesMapperTest.java
```diff
@@ -17,9 +19,17 @@
...
+  @Inject private SecretSeriesMapper.SecretSeriesMapperFactory mapperFactory;
+  @Inject private DSLContext jooq;
   @Inject @Readwrite private GroupDAO groupDAO;
 
+  private SecretSeriesMapper mapper;
+
+  @Before
+  public void before() {
+    mapper = mapperFactory.using(jooq);
+  }
```
mmontgomery-square, line 31: Since we can no longer directly inject the SecretSeriesMapper, we need a bit more setup code to keep our test working.

server/src/test/java/keywhiz/service/resources/admin/SecretsResourceTest.java
```diff
@@ -310,9 +310,7 @@ public void deleteErrorsOnNotFound() {
   }
 
   @Test(expected = ConflictException.class)
-  public void triesToCreateDuplicateSecret() throws Exception {
-    SecretController.SecretBuilder secretBuilder = mock(SecretController.SecretBuilder.class);
-    when(secretController.builder("name", "content", user.getName(), 0)).thenReturn(secretBuilder);
+  public void triesToCreateDuplicateSecret() {
     DataAccessException exception = new DataAccessException("");
     doThrow(exception).when(secretBuilder).create();
```
mmontgomery-square, line 315: Unrelated cleanup, but: this test did not need to declare any thrown exception (so I removed it) and it was duplicating some builder mocking which was being done in the setup method (also removed).

server/src/test/java/keywhiz/service/resources/automation/AutomationSecretResourceIntegrationTest.java
```diff
@@ -47,6 +54,26 @@
...
+        ownerName,
+        "description",
+        "content",
+        ImmutableMap.of(),
+        0);
+    createSecret(request);
+
+    AutomationSecretResponse details = getSecretByName(secretName);
+    assertEquals(ownerName, details.owner());
+  }
```
mmontgomery-square, line 75: This integration test verifies that we can create a secret with an owner via the deprecated automation API.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -66,6 +68,23 @@
...
+    CreateSecretRequestV2 request = CreateSecretRequestV2.builder()
+        .name(secretName)
+        .owner(ownerName)
+        .content(encode("foo"))
+        .build();
+    create(request);
+
+    SecretDetailResponseV2 details = lookup(secretName);
+    assertEquals(ownerName, details.owner());
+  }
```
mmontgomery-square, line 86: This test verifies that we can create a secret with an owner via the V2 automation API.

server/src/test/java/keywhiz/service/resources/admin/SecretsResourceTest.java
isemaya-square: I think we have a lot of unit tests unnecessarily throwing an exception...

server/src/main/java/keywhiz/service/daos/AclDAO.java
```diff
@@ -578,17 +578,24 @@ protected void evictClient(Configuration configuration, long clientId, long grou
...
+        ClientDAOFactory clientDAOFactory,
+        GroupDAOFactory groupDAOFactory,
+        SecretContentDAOFactory secretContentDAOFactory,
+        SecretSeriesDAOFactory secretSeriesDAOFactory,
+        ClientMapper clientMapper,
+        GroupMapper groupMapper,
+        SecretSeriesMapper.SecretSeriesMapperFactory secretSeriesMapperFactory,
+        SecretContentMapper secretContentMapper,
+        RowHmacGenerator rowHmacGenerator,
+        KeywhizConfig config) {
```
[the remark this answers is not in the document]
isemaya-square: Do you mean in later in this PR? :)

Was there a specific reason or motivation for doing so? (asking as someone with no context on the factory pattern, so bear with me...)

isemaya-square (commented):
LGTM, though I don't think you explained the reason for replacing the SecretSeriesMapper with a factory.

server/src/main/java/keywhiz/service/daos/AclDAO.java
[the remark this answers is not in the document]
mmontgomery-square: It all comes down to scoping. Before, the mapper didn't have any database dependencies so it didn't have to care about whether there were any transactions or connections currently in use. Now, it has to make a call into the groups DAO in order to fetch the name of the owning group (if any), which means that it needs to piggyback off of any existing connection/transaction in order to prevent resource leaks. So, every time we need an instance of the mapper we need to pass in the current database "scope", which is what the factory does for us.

There are fancier ways to accomplish this in Guice, either via auto-generated factories or child injectors (my preference). I was planning on putting together a PR with the latter today if I had time.

server/src/test/java/keywhiz/service/resources/admin/SecretsResourceTest.java
mmontgomery-square: Sometimes I'd just slap "throws Exception" onto tests just to prevent any compilation issues around thrown exceptions, especially as test change over time, but IDEs also tend to nag about unnecessary throws declarations (like this one) so I figure getting rid of nag highlights is a win for reducing cognitive load. One less thing to worry about.

isemaya-square (approved):

server/src/main/java/keywhiz/service/daos/AclDAO.java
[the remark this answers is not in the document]
isemaya-square: Ah ok. That does make me wonder though, how did we get by without database dependencies before?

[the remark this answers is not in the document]
mmontgomery-square: The way it works is that it's using a Java Stream to read the raw records from the database, and the translation class is responsible for mapping that into our own model class. Since the translator was starting from the raw record it didn't have to perform any database operations of its own.

I have a nagging feeling that there's a more idiomatic way to accomplish what I'm trying to do (perhaps a subclass of the Jooq entity which includes a field that's populated via a join), and I'll be doing some more exploration of this later this week, but this works for now.

PR #942, opened 2021-09-22
Adding command for backfilling ownership

mmontgomery-square:


isemaya-square (approved):

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -0,0 +1,165 @@
...
+
+    new Backfill(jooq, batchSize, delay).run();
+  }
+
+  private static int getBatchSize(Namespace namespace) {
+    Integer batchSize = namespace.getInt(Args.BATCH_SIZE);
+    return batchSize == null
+        ? 1
+        : batchSize;
+  }
```
mmontgomery-square, line 70: In real-world use this should be unnecessary since we've supplied a default for the argument. However, this also requires us to fake out the parser's behavior during testing, which is annoying. So, I decided to make these methods a bit more robust against missing values than they otherwise would be.

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -0,0 +1,165 @@
...
+    String delayToken = namespace.getString(Args.DELAY);
+    return delayToken == null
+        ? Duration.ZERO
+        : Duration.parse(delayToken);
+  }
+
+  private static class Backfill implements Runnable {
+    private final DSLContext jooq;
+    private final int batchSize;
+    private final Duration delay;
```
mmontgomery-square, line 82: Because almost every method for backfilling requires access to the DSLContext I decided to introduce a helper class in order to scope it as a field.

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    private Cursor<Record1<Long>> allSecretsWithoutOwners() {
+      return jooq
+          .select(SECRETS.ID)
+          .from(SECRETS)
+          .where(SECRETS.OWNER.isNull())
+          .fetchSize(batchSize)
+          .fetchLazy();
+    }
```
mmontgomery-square, line 143: Originally I had a "forSharing()" step in here, with the idea to explicitly tell MySQL not to hold any locks. However, during testing the database yelled at me saying this was unsupported syntax. Odd. In any case it's removed, but the intent of the cursor is that we can scroll through results without having to hold a ton of locks. I need to do a bit more reading on the MySQL side to see if there's another hint we can provide.

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandTest.java
```diff
@@ -0,0 +1,119 @@
...
+    long group3Id = insertGroup();
+
+    grantAccess(secretId, group1Id);
+    grantAccess(secretId, group2Id);
+    grantAccess(secretId, group3Id);
+
+    backfill();
+
+    assertOwnerEquals(secretId, group1Id);
+  }
```
mmontgomery-square, line 43: This is good for some basic coverage, but it doesn't really handle cases like secrets which already have owners, secrets with no access grants, or multiple batches. I'll add a couple more tests to handle those.

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private static void waitOneMillisecond() {
+    try {
+      TimeUnit.MILLISECONDS.sleep(1);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    }
+  }
```
mmontgomery-square, line 118: This is to ensure that access grants are always at least one millisecond apart so that we can reason about which one came first (i.e. we don't want two access grants with identical creation times, since that's what we're sorting by).

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandTest.java
isemaya-square: Actually yeah, good point, I'm going to revoke my stamp until you add in those tests :P

isemaya-square (changes requested):
More testing (like mentioned in comments)

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
isemaya-square: Sounds good. Also minor nit: since this particular ternary operation is very short and simple, I think we could get away with putting everything on one line.

isemaya-square (approved):
Nice testing!

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -0,0 +1,165 @@
...
+  public BackfillOwnershipCommand() {
+    super("backfill-ownership", "Backfills secret ownership");
+  }
+
+  @Override public void configure(Subparser subparser) {
+    subparser.addArgument("--batchSize")
+        .dest(Args.BATCH_SIZE)
+        .type(Integer.class)
+        .setDefault(1)
+        .help("Number of records to process in one batch");
```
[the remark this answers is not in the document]
mmontgomery-square: Done!

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -0,0 +1,165 @@
...
+
+    subparser.addArgument("--delay")
+        .dest(Args.DELAY)
+        .type(String.class)
+        .setDefault("PT0S")
+        .help("Delay between batches in ISO 8601 duration format e.g. PT0.1s");
```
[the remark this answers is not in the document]
mmontgomery-square: Done!

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandTest.java
mmontgomery-square: I've added additional tests, and also added an API for the backfill that returns the number of batches processed so that we can verify that that's working as intended.

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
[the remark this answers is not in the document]
mmontgomery-square: For the record, the help output for DropWizard commands automatically includes the default value in the description, so this ended up just adding some unneeded redundancy.

PR #943, opened 2021-09-23
Allow ownership backfill command to accept configuration file

mmontgomery-square:


server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -33,17 +33,19 @@ public BackfillOwnershipCommand() {
   }
 
   @Override public void configure(Subparser subparser) {
+    super.configure(subparser);
+
     subparser.addArgument("--batchSize")
```
mmontgomery-square, line 38: There are multiple base classes for DropWizard commands, and you extend from them depending on the features you want. In our case we need to access the app's configuration, so we extend from the configuration-aware base class. That class operates by overriding the configure() method to add in a command line argument which is the path to the configuration file. That's great, but, if you also need to override the configure() method to add in your own command line arguments you need to remember to call super() in order to preserve the configuration file behavior.

server/src/main/java/keywhiz/commands/BackfillOwnershipCommand.java
```diff
@@ -33,17 +33,19 @@ public BackfillOwnershipCommand() {
...
-        .help("Number of records to process in one batch (default is 1)");
+        .help("Number of records to process in one batch");
 
     subparser.addArgument("--delay")
         .dest(Args.DELAY)
         .type(String.class)
         .setDefault("PT0S")
-        .help("Delay between batches in ISO 8601 duration format e.g. PT0.1s (default is no delay)");
+        .help("Delay between batches in ISO 8601 duration format");
   }
```
mmontgomery-square, line 49: It turns out that DropWizard's built-in command help system automatically prints out the default value for an argument so there's no need to provide it manually.

sq-smits (approved):

mmontgomery-square:
For the record, sq-smits is part of our team, we're just waiting for the GitHub membership to be updated.

PR #961, opened 2021-11-05
Expose secret renaming operation on automation V2 API

mmontgomery-square:


server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -70,6 +72,37 @@
...
+
+  @Test public void cannotRenameSecretToExistingSecret() throws Exception {
+    String oldName = UUID.randomUUID().toString();
+    createSecret(oldName);
+
+    String newName = UUID.randomUUID().toString();
+    createSecret(newName);
+
+    assertEquals(500, renameSecret(oldName, newName));
+  }
```
mmontgomery-square, line 94: Upon further reflection, I think this error code is inconsistent with current behavior. Trying to create the same secret twice results in a ConflictException (error code 409), and I believe we should be doing the same here.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -198,6 +198,21 @@ public Response createOrUpdateSecret(@Auth AutomationClient automationClient,
...
   }
 
+  @Timed @ExceptionMetered
+  @Path("{oldName}/rename/{newName}")
+  @POST
+  @Consumes(APPLICATION_JSON)
+  public Response renameSecret(
+      @Auth AutomationClient automationClient,
+      @PathParam("oldName") String oldName,
+      @PathParam("newName") String newName) {
```
mmontgomery-square, line 208: I'm not terribly familiar with @Consumes, but the precedent seems to be that even if the POST method takes no input other than URL parameters it still declares that it consumes JSON. On the client side, the request type is still application/json (or whatever) but the body is empty.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -66,6 +68,37 @@
     mutualSslClient = TestClients.mutualSslClient();
   }
 
+  //---------------------------------------------------------------------------------------
+  // renameSecret
+  //---------------------------------------------------------------------------------------
+
+  @Test public void renamesSecret() throws Exception {
```
isemaya-square, line 75: nit: test name is pretty similar to the helper method`renameSecret`, maybe `successfullyRenamesSecret` instead?

isemaya-square (approved):

PR #964, opened 2021-11-10
Switching secret rename response code to 201 Created to be more REST-y

mmontgomery-square:


server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -210,7 +210,8 @@ public Response renameSecret(
         .orElseThrow(NotFoundException::new);
     secretDAO.renameSecretById(secret.series().id(), newName, automationClient.getName());
 
-    return Response.ok().build();
+    UriBuilder uriBuilder = UriBuilder.fromResource(SecretResource.class).path(newName);
+    return Response.created(uriBuilder.build()).build();
```
mmontgomery-square, line 214: Rather than just saying "ok", we indicate that a new resource (the renamed secret) was created and return the URL to that resource.

sq-smits (approved):

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -77,7 +77,7 @@
     createSecret(oldName);
 
     String newName = UUID.randomUUID().toString();
-    assertEquals(200, renameSecret(oldName, newName));
```
lavanyaharinarayan, line 80: Should we also test that the URL is as expected?

lavanyaharinarayan (approved):

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
mmontgomery-square: Yeah, we probably should, I'll get to that shortly.

PR #972, opened 2021-11-30
Renaming a secret now updates the row HMAC

mmontgomery-square:


server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -101,7 +102,19 @@
...
+    private final String configEntry;
+
+    RowHmacCheck(String configEntry) {
+      this.configEntry = configEntry;
+    }
+
+    public String getConfigEntry() {
+      return configEntry;
+    }
   }
```
mmontgomery-square, line 118: For testing, we need a way to override the configured HMAC check configuration so that mismatches fail. Since enum-to-string translation is now two-way I decided to store the config file tokens directly with the enums themselves in order to simplify things and increase maintainability.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -85,13 +86,33 @@ long createSecretSeries(
...
+  @VisibleForTesting
+  long createSecretSeries(
+      long id,
+      String name,
+      Long ownerId,
+      String creator,
+      String description,
+      @Nullable String type,
+      @Nullable Map<String, String> generationOptions,
+      long now) {
```
mmontgomery-square, line 110: This is an override of the existing method (which now delegates to this one) with the only difference being that the ID is passed in rather than generated. In order to write a "known-answer test" for the row HMAC we need total control of the inputs (the ID and name), and as it stood the ID was being generated randomly. This seemed more straightforward than trying to futz with the random number generator to spit out some pre-determined output.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -131,8 +152,7 @@ void updateSecretSeries(
     }
 
     try {
-      String rowHmac = rowHmacGenerator.computeRowHmac(
-          SECRETS.getName(), List.of(name, secretId));
+      String rowHmac = computeRowHmac(secretId, name);
```
mmontgomery-square, line 155: The logic to compute the HMAC is repeated in a couple places so I introduced a helper method to reduce duplication.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -348,8 +368,10 @@ public void deleteSecretSeriesById(long id) {
   }
 
   public void renameSecretSeriesById(long secretId, String name, String creator, long now) {
+    String rowHmac = computeRowHmac(secretId, name);
     dslContext.update(SECRETS)
         .set(SECRETS.NAME, name)
+        .set(SECRETS.ROW_HMAC, rowHmac)
         .set(SECRETS.UPDATEDBY, creator)
         .set(SECRETS.UPDATEDAT, now)
         .where(SECRETS.ID.eq(secretId))
```
mmontgomery-square, line 377: This is our bug fix: the row HMAC is now calculated and updated as part of the renaming process.

server/src/test/java/keywhiz/service/daos/AclDAOTest.java
```diff
@@ -90,6 +94,26 @@
...
+        ApiDate.now().toEpochSecond());
+
+    KeywhizConfig.RowHmacCheck originalHmacCheck = keywhizConfig.getRowHmacCheck();
+    try {
+      keywhizConfig.setRowHmacCheck(KeywhizConfig.RowHmacCheck.ENFORCED);
+      aclDAO.getSanitizedSecretsFor(client1);
+    } finally {
+      keywhizConfig.setRowHmacCheck(originalHmacCheck);
+    }
+  }
```
mmontgomery-square, line 115: This test initially failed before the bug was fixed. This is a higher-level unit test that mimics the actual calls that were in the initial bug report.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -47,6 +50,51 @@
...
+        "abc",
+        null,
+        null,
+        null,
+        null,
+        null,
+        ApiDate.now().toEpochSecond());
+
+    assertEquals(wellKnownId, id);
+  }
```
mmontgomery-square, line 78: This test is probably silly, it's on the same level as a getter test. Maybe I'm just being paranoid.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -47,6 +50,51 @@
...
+        null,
+        null,
+        null,
+        null,
+        ApiDate.now().toEpochSecond());
+
+    assertEquals(
+        "649F23AF8C88F32AAEDD848EB5F42A8770BBE59B6DBE20E66DDA2AB915E28876",
+        getRowHmac(wellKnownId));
+  }
```
mmontgomery-square, line 96: This known-answer test actually caught a bug where I had reversed the order of the inputs to the HMAC (secret series ID and name) during refactoring. Regardless, I wanted to add this as a warning in case future maintainers inadvertently (or otherwise) modify how the HMAC is calculated.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -201,18 +214,20 @@ public String getDerivationProviderClass() {
...
-        return RowHmacCheck.ENFORCED;
-      case "logging":
-        return RowHmacCheck.DISABLED_BUT_LOG;
-      case "disabled":
-        return RowHmacCheck.DISABLED;
-      default:
-        throw new IllegalArgumentException(
-            String.format("%s is an invalid rowHmacCheck parameter", rowHmacCheck)
-        );
+    for (RowHmacCheck e : RowHmacCheck.values()) {
```
isemaya-square, line 217: Nice fix!

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
isemaya-square: Might be good to leave this cautionary tale as a comment :P

isemaya-square (approved):

PR #979, opened 2021-12-03
Fixing serialization of RowHmacCheck enum

mmontgomery-square:


server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -33,7 +33,6 @@
 import keywhiz.service.config.ClientAuthConfig;
 import keywhiz.service.config.KeyStoreConfig;
 import keywhiz.service.config.Templates;
-import keywhiz.service.crypto.RowHmacGenerator;
 import org.hibernate.validator.constraints.Length;
 import org.hibernate.validator.constraints.NotEmpty;
```
mmontgomery-square, line 37: Unrelated to this change, but this import was unused so I removed it.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -90,7 +89,7 @@
   private String statusCacheExpiry;
 
   @JsonProperty
-  private String rowHmacCheck;
+  private RowHmacCheck rowHmacCheck;
```
mmontgomery-square, line 92: Jackson supports properties which allow direct mapping from strings to enum values, so we can now store the property natively as an enum.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -102,19 +101,12 @@
...
-    public String getConfigEntry() {
-      return configEntry;
-    }
+    @JsonProperty("disabled")
+    DISABLED,
+    @JsonProperty("logging")
+    DISABLED_BUT_LOG,
+    @JsonProperty("enforced")
+    ENFORCED;
   }
```
mmontgomery-square, line 110: Short. And. Sweet.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -214,20 +206,12 @@ public String getDerivationProviderClass() {
...
-    );
+    return rowHmacCheck;
   }
 
   @VisibleForTesting
-  public void setRowHmacCheck(RowHmacCheck check) {
-    rowHmacCheck = check.getConfigEntry();
+  public void setRowHmacCheck(RowHmacCheck rowHmacCheck) {
+    this.rowHmacCheck = rowHmacCheck;
   }
```
mmontgomery-square, line 215: As a positive side-effect, the accessors are now completely straightforward.

server/src/test/java/keywhiz/KeywhizConfigTest.java
```diff
@@ -18,6 +18,24 @@
...
+  public void parsesRowHmacCheckEnforced() {
+    KeywhizConfig config = loadConfig("row-hmac-check-enforced.yaml");
+    assertEquals(KeywhizConfig.RowHmacCheck.ENFORCED, config.getRowHmacCheck());
+  }
+
+  @Test
+  public void parsesRowHmacCheckLogging() {
+    KeywhizConfig config = loadConfig("row-hmac-check-logging.yaml");
+    assertEquals(KeywhizConfig.RowHmacCheck.DISABLED_BUT_LOG, config.getRowHmacCheck());
+  }
```
mmontgomery-square, line 37: This test was initially failing and reproduces the specific error that we encountered.

server/src/test/resources/configs/row-hmac-check-disabled.yaml
```diff
@@ -0,0 +1,157 @@
+server:
+  applicationConnectors:
+    - type: resources-https
+      port: 4445
+      keyStorePath: dev_and_test_keystore.p12
```
mmontgomery-square, line 5: Unfortunately (depending on how you look at it), the large number of required values in our configuration file means that testing serialization of a single property requires a large amount of boilerplate.

server/src/main/java/keywhiz/KeywhizConfig.java
isemaya-square: Does  the`@JsonProperty` annotation remove the need for getters and setters?

server/src/test/resources/configs/row-hmac-check-disabled.yaml
```diff
@@ -0,0 +1,157 @@
...
+  # password:
+
+migrationsDir:
+  db/mysql/migration
+
+statusCacheExpiry: PT3S
+
+backupExportKeys:
+  test: |
+    -----BEGIN PGP PUBLIC KEY BLOCK-----
```
isemaya-square, line 75: is this value different for each yaml?

isemaya-square (approved):
LGTM. My only question is if there are any common values between the three yaml files and if it were possible to combine them (if not no worries, even if there are common values, I wonder if creating a common yaml will just make things confusing)

server/src/main/java/keywhiz/KeywhizConfig.java
mmontgomery-square: For the enum, yes. We were using the configEntry field to model the serialized value of each enum value, but this can be accomplished more cleanly by using the Jackson annotation.

server/src/test/resources/configs/row-hmac-check-disabled.yaml
mmontgomery-square: I just copied an existing config file and modified the field in question, so everything else is identical.

mmontgomery-square:
> 

Unfortunately the include directives aren't a standard yaml feature (that appears to be some in-house pre-processing) so DropWizard (and by extension, the tests) expects the configuration file to be fully self-contained. When I first introduced these configuration tests I tried to do something more clever to reduce duplication but didn't have any luck.

However... now that I think about it, we could probably have a standard "valid" configuration that parses and then try to merge/overwrite it with our test-specific changes. But, that would involve doing some yaml manipulation and I don't know if that would involve additional dependencies, etc. Probably worth looking into in the future.

PR #984, opened 2021-12-06
Adding API for manually backfilling the row HMAC for individual secrets

mmontgomery-square:


server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceTest.java
```diff
@@ -0,0 +1,72 @@
...
+  private SecretSeriesDAO secretSeriesDAO;
+
+  @Before
+  public void before() {
+    secretDAO = secretDAOFactory.readwrite();
+    secretSeriesDAO = secretSeriesDAOFactory.readwrite();
+  }
+
+  @Test
+  public void backfillsNonNullSecretHmac() {
```
mmontgomery-square, line 37: I had another test which attempted to verify that it would also backfill a null HMAC, however, it turns out that that isn't allowed by the database schema.

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -45,6 +47,32 @@ public BackfillRowHmacResource(DSLContext jooq, RowHmacGenerator rowHmacGenerato
...
+  public void backfillSecretRowHmacByName(@PathParam("secretName") String secretName) {
+    Optional<SecretsRecord> maybeRow = jooq.selectFrom(SECRETS)
+        .where(SECRETS.NAME.eq(secretName))
+        .fetchOptional();
+
+    if (!maybeRow.isPresent()) {
+      throw new NotFoundException();
+    }
+
+    SecretsRecord row = maybeRow.get();
```
mbyczkowski, line 67: Should we return the new hmac or some information that the HMAC has changed?

mmontgomery-square: AFAICT the row-level HMACs are an internal implementation detail and are never exposed via any of the existing APIs/models, so I didn't want to be the first to take that step. Would logging work? I don't see any obvious HTTP response codes that we could use to distinguish changed/not changed.

log/src/main/java/keywhiz/log/EventTag.java
```diff
@@ -13,6 +13,7 @@
   SECRET_READCONTENT,
   SECRET_RENAME,
   SECRET_UPDATECURRENT,
+  SECRET_BACKFILLHMAC,
```
mmontgomery-square, line 16: Following the existing naming precedent here.

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,11 +50,65 @@
...
+  @POST
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  public void backfillSecretRowHmacByName(
+      @Auth AutomationClient automationClient,
+      @PathParam("secretName") String secretName,
+      @QueryParam("force") boolean force) {
+    Optional<SecretsRecord> maybeRow = jooq.selectFrom(SECRETS)
+        .where(SECRETS.NAME.eq(secretName))
+        .fetchOptional();
```
mmontgomery-square, line 79: Originally I was thinking that authentication was handled at the class level, but it's not: I've now added @Auth AutomationClient to enforce that this is an authenticated, high-privilege caller (with tests to verify).

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,11 +50,65 @@
...
+        .put("newHmac", newHmac)
+        .build();
+
+    auditLog.recordEvent(new Event(
+        Instant.now(),
+        EventTag.SECRET_BACKFILLHMAC,
+        automationClient.getName(),
+        secretName,
+        extraInfo));
   }
```
mmontgomery-square, line 112: Now with audit logging! Note that we use Objects.toString() so that null becomes "null" since ImmutableMap doesn't allow null values.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -50,6 +50,17 @@
...
+  @Test public void setsRowHmacByName() {
+    String secretName = randomName();
+    createSecretSeries(secretName);
+
+    String newHmac = UUID.randomUUID().toString();
+    secretSeriesDAO.setRowHmacByName(secretName, newHmac);
+
+    String updatedHmac = jooqContext.fetchOne(SECRETS, SECRETS.NAME.eq(secretName)).getRowHmac();
+    assertEquals(newHmac, updatedHmac);
+  }
```
mmontgomery-square, line 62: This test verifies the basic update-hmac functionality at the DAO level.

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceClientIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class BackfillRowHmacResourceClientIntegrationTest {
+  private static final ObjectMapper mapper =
+      KeywhizService.customizeObjectMapper(Jackson.newObjectMapper());
+  private static final Base64.Encoder encoder = Base64.getEncoder();
+
+  OkHttpClient mutualSslClient;
+
+  @ClassRule public static final RuleChain chain = IntegrationTestRule.rule();
```
mmontgomery-square, line 33: We talked about naming in the previous commit, but I've taken it even further and added "Integration" to the name. Given that the test using the IntegrationTestRule, I feel like this is justified (and more clear than omitting it).

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceClientIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Before public void setUp() {
+    mutualSslClient = TestClients.mutualSslClient();
+  }
+
+  @Test public void backfillSecretHmacRequiresAuthentication() throws Exception {
+    String secretName = UUID.randomUUID().toString();
+    assertEquals(401, backfillSecretHmac(TestClients.unauthenticatedClient(), secretName, null).code());
+  }
```
mmontgomery-square, line 42: This test uses a non-authenticated client and verifies that the response is a 401 Unauthorized. To facilitate this, I added an override which allows us to specify the client to be used rather than always using the authenticated client.

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,11 +50,65 @@
...
+    if (!maybeRow.isPresent()) {
+      throw new NotFoundException();
+    }
+
+    SecretsRecord row = maybeRow.get();
+    String oldHmac = row.getRowHmac();
+
+    if (oldHmac != null && !force) {
+      return;
+    }
```
mmontgomery-square, line 90: This is easy to overlook, but we're introducing a "force" parameter so that we can support two use cases:

1. An actual backfill (only populating the HMAC if it's missing) via force = false
2. Re-calculating/over-writing an existing HMAC via force = true

I wanted to do this so that we could follow the spirit of the existing APIs in this resource (which are strictly backfill-only) while also enabling our "repair" use case.

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceClientIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+    assertEquals(204, backfillSecretHmac(secretName, true).code());
+  }
+
+  // The HMAC is internal (never exposed to the client) so we can only really check
+  // whether the call succeeds or not.
+  @Test public void backfillsSecretHmacWithForceFalse() throws Exception {
+    String secretName = UUID.randomUUID().toString();
+    createSecret(secretName);
+    assertEquals(204, backfillSecretHmac(secretName, false).code());
+  }
```
mmontgomery-square, line 66: Covering our three flavors of "force" (true, false, and missing)

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceIntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+    resource.backfillSecretRowHmacByName(automationClient, secretName, true);
+
+    String updatedHmac = getHmac(secretName);
+    assertNotEquals(newHmac, updatedHmac);
+    assertEquals(originalHmac, updatedHmac);
+  }
+
+  @Test
+  public void backfillSecretRowHmacByNameDoesNotBackfillNonNullHmacIfForceIsFalse() {
```
mmontgomery-square, line 62: It does what it says.

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceTest.java
```diff
@@ -0,0 +1,103 @@
...
+  private ArgumentCaptor<Event> eventCaptor;
+
+  @Before
+  public void before() {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void backfillSecretHmacByNameHandlesNullOldHmac() {
+    SecretsRecord record = new SecretsRecord();
```
mmontgomery-square, line 43: Although technically not a possibility in the really-real world because null HMACs are no longer allowed by the database schema, this test arose after discovering that the ImmutableMap used to track the before-and-after HMACs in the audit log entry barfed if the old HMAC was null, which means it would fail to satisfy the actually-backfilling-a-null-HMAC use case.

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceTest.java
```diff
@@ -0,0 +1,103 @@
...
+    String newHmac = UUID.randomUUID().toString();
+
+    when(rowHmacGenerator.computeRowHmac(any(), any())).thenReturn(newHmac);
+
+    SecretsRecord record = new SecretsRecord();
+    record.setId(1L);
+    record.setName("foo");
+    record.setRowHmac(oldHmac);
+
+    DSLContext context = TestDSLContexts.returning(record);
```
mmontgomery-square, line 74: The DSLContext's builder-ish API isn't amenable to trivial mocking but I'm happy with this simple alternative, which is basically shorthand for when-select-then-return-thing.

server/src/test/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResourceTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    Event event = eventCaptor.getValue();
+    assertEquals(EventTag.SECRET_BACKFILLHMAC, event.getType());
+    assertEquals(clientName, event.getUser());
+    assertEquals(secretName, event.getObjectName());
+
+    Map<String, String> extraInfo = event.getExtraInfo();
+    assertEquals(oldHmac, extraInfo.get("oldHmac"));
+    assertEquals(newHmac, extraInfo.get("newHmac"));
+  }
```
mmontgomery-square, line 95: Verifying the contents of the audit log entry.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -182,6 +182,13 @@ public int setExpiration(long secretContentId, Instant expiration) {
...
+  public int setRowHmacByName(String secretName, String hmac) {
+    return dslContext.update(SECRETS)
+        .set(SECRETS.ROW_HMAC, hmac)
+        .where(SECRETS.NAME.eq(secretName))
+        .execute();
+  }
+
   public int setHmac(long secretContentId, String hmac) {
     return dslContext.update(SECRETS_CONTENT)
         .set(SECRETS_CONTENT.CONTENT_HMAC, hmac)
```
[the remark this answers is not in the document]
captiosus: maybe we could add a comment or change the name of this function to clarify

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -38,11 +50,65 @@
...
+
+    String newHmac = rowHmacGenerator.computeRowHmac(
```
captiosus, line 92: would probably be a good time to standardize computing the row hmac instead of having arbitrary parameters and repeating this everywhere

captiosus (approved):
LGTM

server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
mmontgomery-square: I looked into this, but it's a slightly bigger change than I was expecting. Calculating the HMAC on persistence is done against one model, but calculating the HMAC on load is done against the low-level row model, so we'd have to change around the guts a bit if we want to have, say, a single utility method on the RowHmacGenerator class. OTOH I was also thinking that we could add methods to the model classes themselves. But yeah, a follow-up change is to add the owner to the row HMAC so I was planning on incorporating the refactoring work as part of that work.

PR #1054, opened 2022-04-14
Compare HMACs with MessageDigest to avoid timing attacks

lavanyaharinarayan:
String.equals is vulnerable to timing attacks.

### DIFF
--- a/server/src/main/java/keywhiz/service/crypto/RowHmacGenerator.java
+++ b/server/src/main/java/keywhiz/service/crypto/RowHmacGenerator.java
@@ -42,7 +42,7 @@ public class RowHmacGenerator {
    *
    * Uses MessageDigest to prevent timing attacks.
    * */
-  public boolean compareHmacs(String left, String right) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/crypto/RowHmacGenerator.java
```diff
@@ -36,6 +37,18 @@ public String computeRowHmac(String table, List<Object> fields) {
     return cryptographer.computeHmacWithSecretKey(hmacContent.getBytes(UTF_8), hmacKey);
   }
 
+  /**
+   * Checks whether two HMACs are equal.
+   *
+   * Uses MessageDigest to prevent timing attacks.
+   * */
+  public boolean compareHmacs(String left, String right) {
```
jdtw, line 45: I don't know Java well, so I might be totally off base here, but can/should this method be static since it doesn't need to use any of the class members?

jdtw (approved):

PR #1055, opened 2022-04-15
added Action interface

violetd12:


server/src/main/java/keywhiz/service/permissions/Action.java
```diff
@@ -0,0 +1,23 @@
...
+  // for secrets and permissions, it will be a soft deletion
+  public static final String DELETE = "DELETE";
+
+  public static final String READ = "READ";
+
+  // remove clients to groups or remove secrets to groups
+  public static final String REMOVE = "REMOVE";
+
+  public static final String UPDATE = "UPDATE";
+}
```
mmontgomery-square, line 23: For the actual constants themselves, I would remove the comments. This class doesn't "know" how it will be consumed and the problem is that, if that changes down the road, that change will most likely not be reflected here. For example, information about what kind of deletion is performed is better off coming from the actual implementation (and who knows, we may have different kinds of deletion based on the object type) - we shouldn't make any assumptions about that ahead of time, even if that's officially part of our plan.

server/src/main/java/keywhiz/service/permissions/Action.java
```diff
@@ -0,0 +1,23 @@
+package keywhiz.service.permissions;
+
+/**
+ * Action class defines the set of actions that can be performed on an object
+ * */
+public final class Action {
```
mmontgomery-square, line 6: Typically I'm very anti-comment but given that action is a very generic term I think this is helpful.

mmontgomery-square (approved):

PR #1059, opened 2022-04-20
added PermissionCheck interface with placeholder KeywhizPrincipal int…

violetd12:
…erface

### DIFF
--- a/server/src/main/java/keywhiz/service/permissions/KeywhizPrincipal.java
+++ b/server/src/main/java/keywhiz/service/permissions/KeywhizPrincipal.java
@@ -1,0 +1,4 @@
+package keywhiz.service.permissions;
+
+public interface KeywhizPrincipal {
+}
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/permissions/KeywhizPrincipal.java
```diff
@@ -0,0 +1,4 @@
+package keywhiz.service.permissions;
+
+public interface KeywhizPrincipal {
+}
```
mmontgomery-square, line 4: Now I'm wondering if we should put this into the auth package, or leave it here and call it "Actor". Let's leave it as-is and we can figure out the naming later.

mmontgomery-square (approved):

isemaya-square:
Nit: I like to add the CISM ticket, either to the title or to the description so it's easy to reference what task the PR is associated with.

isemaya-square (approved):

PR #1068, opened 2022-05-20
Added AlwaysAllowPermissionCheck

violetd12:
Added a permission check functionality.

### DIFF
--- a/server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
+++ b/server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
@@ -1,0 +1,97 @@
+package keywhiz.service.permission;
+
+import com.codahale.metrics.MetricRegistry;
+import java.util.Objects;
+import keywhiz.service.permissions.Action;
+import keywhiz.service.permissions.AlwaysAllowDelegatingPermissionCheck;
+import keywhiz.service.permissions.KeywhizPrincipal;
+import keywhiz.service.permissions.PermissionCheck;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.when;
+
+public class AlwaysAllowDelegatingPermissionCheckTest {
+
+  @Rule public MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock PermissionCheck delegate;
+
+  private MetricRegistry metricRegistry;
+  private AlwaysAllowDelegatingPermissionCheck alwaysAllow;
+
+  private static KeywhizPrincipal principal;
+  private static Objects target;
+
+  private static final String ISALLOWED_SUCCESS_METRIC_NAME = "keywhiz.service.permissions.AlwaysAllowDelegatingPermissionCheck.isAllowed.success.histogram";
+  private static final String ISALLOWED_FAILURE_METRIC_NAME = "keywhiz.service.permissions.AlwaysAllowDelegatingPermissionCheck.isAllowed.failure.histogram";
+  private static final String CHECKALLOWEDORTHROW_SUCCESS_METRIC_NAME = "keywhiz.service.permissions.AlwaysAllowDelegatingPermissionCheck.checkAllowedOrThrow.success.histogram";
+  private static final String CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME = "keywhiz.service.permissions.AlwaysAllowDelegatingPermissionCheck.checkAllowedOrThrow.exception.histogram";
+
+  @Before
+  public void setUp() {
+    metricRegistry = new MetricRegistry();
+    alwaysAllow = new AlwaysAllowDelegatingPermissionCheck(metricRegistry, delegate);
+  }
+
+  @Test public void isAllowedReturnsTrueWhenDelegateReturnsTrue() {
+    when(delegate.isAllowed(any(), any(), any())).thenReturn(true);
+
+    boolean permitted = alwaysAllow.isAllowed(principal, Action.ADD, target);
+
+    assertThat(permitted);
+
+    assertThat(metricRegistry.histogram(ISALLOWED_SUCCESS_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(ISALLOWED_SUCCESS_METRIC_NAME).getSnapshot().getMean()).isEqualTo(1);
+
+    assertThat(metricRegistry.histogram(ISALLOWED_FAILURE_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(ISALLOWED_FAILURE_METRIC_NAME).getSnapshot().getMean()).isEqualTo(0);
+  }
+
+  @Test public void isAllowedReturnsTrueWhenDelegateReturnsFalse() {
+    when(delegate.isAllowed(any(), any(), any())).thenReturn(false);
+
+    boolean permitted = alwaysAllow.isAllowed(principal, Action.ADD, target);
+
+    assertThat(permitted);
+
+    assertThat(metricRegistry.histogram(ISALLOWED_SUCCESS_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(ISALLOWED_SUCCESS_METRIC_NAME).getSnapshot().getMean()).isEqualTo(0);
+
+    assertThat(metricRegistry.histogram(ISALLOWED_FAILURE_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(ISALLOWED_FAILURE_METRIC_NAME).getSnapshot().getMean()).isEqualTo(1);
+  }
+
+  @Test public void CheckAllowedOrThrowReturnsVoidWhenDelegateReturnsVoid() {
+    doNothing().when(delegate).checkAllowedOrThrow(any(), any(), any());
+
+    alwaysAllow.checkAllowedOrThrow(principal, Action.ADD, target);
+
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_SUCCESS_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_SUCCESS_METRIC_NAME).getSnapshot().getMean()).isEqualTo(1);
+
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME).getSnapshot().getMean()).isEqualTo(0);
+  }
+
+  @Test public void CheckAllowedOrThrowReturnsVoidWhenDelegateThrowException() {
+    doThrow(RuntimeException.class).when(delegate).checkAllowedOrThrow(any(), any(), any());
+
+    alwaysAllow.checkAllowedOrThrow(principal, Action.ADD, target);
+
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_SUCCESS_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_SUCCESS_METRIC_NAME).getSnapshot().getMean()).isEqualTo(0);
+
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME).getSnapshot().getMean()).isEqualTo(1);
+  }
+}
+
--- a/server/src/main/java/keywhiz/KeywhizService.java
+++ b/server/src/main/java/keywhiz/KeywhizService.java
@@ -39,6 +39,7 @@ import keywhiz.commands.PreviewMigrateCommand;
 import keywhiz.inject.InjectorFactory;
 import keywhiz.service.filters.CookieRenewingFilter;
 import keywhiz.service.filters.SecurityHeadersFilter;
+import keywhiz.service.permissions.PermissionCheck;
 import keywhiz.service.providers.AuthResolver;
 import keywhiz.service.providers.AutomationClientAuthFactory;
 import keywhiz.service.providers.ClientAuthFactory;
@@ -139,6 +140,8 @@ public class KeywhizService extends Application<KeywhizConfig> {
     jersey.register(new AuthResolver.Binder(injector.getInstance(ClientAuthFactory.class),
         injector.getInstance(AutomationClientAuthFactory.class),
         injector.getInstance(UserAuthFactory.class)));
+
+    jersey.register(injector.getInstance(PermissionCheck.class));
 
     logger.debug("Registering resources");
     jersey.register(injector.getInstance(BackfillRowHmacResource.class));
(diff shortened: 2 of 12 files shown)


server/src/main/java/keywhiz/service/permissions/AlwaysAllowPermissionCheck.java
```diff
@@ -0,0 +1,40 @@
...
+  @Counted
+  public boolean isAllowed(KeywhizPrincipal source, String action, Object target) {
+    boolean hasPermission = this.delegate.isAllowed(source, action, target);
+    if (!hasPermission) {
+      denied();
+    }
+    if (hasPermission) {
+      logger.info(String.format("In delegate - %s wants to %s %s returns true", source, action, target));
+    } else {
+      logger.info(String.format("In delegate - %s wants to %s %s returns false", source, action, target));
```
mmontgomery-square, line 29: Since the class name is already included in the log line, we can probably drop the "In delegate - " prefix. Also, since the sources and targets might not always have super obvious string representations (their types won't be included), it'll be more clear if you format the message using fields like so:

Actor: %s Action: %s Target: %s Result: %s

Something like that. Finally, since hasPermission is a boolean you can use that directly via a fourth %s rather than using an if/else and printing "true" and "false" manually.

server/src/main/java/keywhiz/service/permissions/AlwaysAllowPermissionCheck.java
```diff
@@ -0,0 +1,40 @@
...
+    this.metricRegistry = metricRegistry;
+  }
+
+  @Counted
+  public boolean isAllowed(KeywhizPrincipal source, String action, Object target) {
+    boolean hasPermission = this.delegate.isAllowed(source, action, target);
+    if (!hasPermission) {
+      denied();
+    }
+    if (hasPermission) {
```
mmontgomery-square, line 26: Personal preference, but I like to put a blank line between blocks (anything with a pair of curly braces) to give them some visual space.

server/src/main/java/keywhiz/service/permissions/AlwaysAllowPermissionCheck.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+    return true;
+  }
+
+  @Counted
+  private void denied() {}
+
+  public void checkAllowedOrThrow(KeywhizPrincipal source, String action, Object target) {
+    this.delegate.isAllowed(source, action, target);
+  }
```
mmontgomery-square, line 39: I would add some logging here that matches what we have for the isAllowed method.

server/src/main/java/keywhiz/service/permissions/PermissionCheckProvider.java
```diff
@@ -0,0 +1,21 @@
...
+    this.metricRegistry = metricRegistry;
+  }
+
+  @Override
+  public PermissionCheck get() {
+    PermissionCheck alwaysFail = new AlwaysFailPermissionCheck();
+    PermissionCheck alwaysAllow = new AlwaysAllowPermissionCheck(this.metricRegistry, alwaysFail);
+    return alwaysAllow;
+  }
+}
```
mmontgomery-square, line 21: Ok, I had to read around a bit to see how this is used. Binding to a `Provider<T>` and having `@ Provides T someMethod()` on a module are totally equivalent as far as creating instances of T, however, I think since our logic is simple (so far) it might be more clear to move this to an `@Provides` method on the module so that the logic/binding is all in one place.

server/src/test/java/keywhiz/service/permission/PermissionCheckTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PermissionCheckTest {
```
mmontgomery-square, line 21: Since PermissionCheck is an interface it doesn't have any behavior to test (default methods aside), so it probably doesn't need its own unit test. Looking at the test names, this is probably supposed to be AlwaysAllowPermissionCheckTest.

server/src/test/java/keywhiz/service/permission/PermissionCheckTest.java
```diff
@@ -0,0 +1,51 @@
...
+  PermissionCheckImpl permissionCheck;
+  String add = Action.ADD;
+  String create = Action.CREATE;
+  String delete = Action.DELETE;
+
+  @Before
+  public void setUp() {
+    permissionCheckProvider = new PermissionCheckProvider(metricRegistry);
+    permissionCheck = new PermissionCheckImpl(permissionCheckProvider);
+  }
```
mmontgomery-square, line 38: I don't think this indirection is necessary, we can just create a new AlwaysAllowPermissionCheck by passing in a mock delegate and then control the behavior of the mock.

server/src/test/java/keywhiz/service/permission/PermissionCheckTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Rule public MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock MetricRegistry metricRegistry;
+  @Mock KeywhizPrincipal principal;
+  @Mock Object target;
+  Provider<PermissionCheck> permissionCheckProvider;
+  PermissionCheckImpl permissionCheck;
+  String add = Action.ADD;
+  String create = Action.CREATE;
+  String delete = Action.DELETE;
```
mmontgomery-square, line 32: I would just use the constants directly in the code instead of storing them in locals, that way it's obvious that they're constants when you see them in use.

server/src/test/java/keywhiz/service/permission/PermissionCheckTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PermissionCheckTest {
+
+  @Rule public MockitoRule mockito = MockitoJUnit.rule();
+
+  @Mock MetricRegistry metricRegistry;
+  @Mock KeywhizPrincipal principal;
+  @Mock Object target;
```
mmontgomery-square, line 27: Mocks are an easy way to create "don't care" instances of objects (such as the first two here) but for the target I might just create a new Object(). @Mock implies that you're actually going to mock some behavior but for Object it's just a thing we're passing around, there's no behavior to mock.

server/src/test/java/keywhiz/service/permission/PermissionCheckTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Test public void IsAllowedAlwaysReturnsTrue() {
+    boolean permitted_add = permissionCheck.isAllowed(principal, add, target);
+    assertThat(permitted_add).isEqualTo(true);
+
+    boolean permitted_create = permissionCheck.isAllowed(principal, create, target);
+    assertThat(permitted_create).isEqualTo(true);
+
+    boolean permitted_delete = permissionCheck.isAllowed(principal, delete, target);
+    assertThat(permitted_delete).isEqualTo(true);
+  }
```
mmontgomery-square, line 49: Make sure you follow the variable naming convention for Java (camelCase).

As far as how to implement this test, the behavior you want to verify is that the permission check always returns true regardless of the result of the underlying permission check (which could be true or false, or it could throw an exception). So you'd want to have a test for each of those conditions, and the easiest way to do that is to use a @ Mock PermissionCheck as the delegate to the AlwaysAllowPermissionCheck and then control what the output of the mock is.

I like to think of tests in terms of truth tables, so something like:

* Delegate: T Result: T
* Delegate: F Result: T
* Delegate: X Result: T

(where X is exception, although I don't know if we actually discussed what we should do in that case)

Basically, what are all the possible combinations of inputs and expected outputs, and is there a test for each combination?

server/src/main/java/keywhiz/service/permissions/PermissionCheckImpl.java
```diff
@@ -0,0 +1,21 @@
+package keywhiz.service.permissions;
+
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+
+public class PermissionCheckImpl implements PermissionCheck{
```
mmontgomery-square, line 6: I don't think you'll actually end up needing this class, it's just a delegate with no additional behavior.

server/src/main/java/keywhiz/service/permissions/AlwaysAllowPermissionCheck.java
```diff
@@ -0,0 +1,40 @@
+package keywhiz.service.permissions;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.annotation.Counted;
+import com.google.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AlwaysAllowPermissionCheck implements PermissionCheck{
```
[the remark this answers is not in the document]
violetd12: > you could imagine having an AlwaysAllow check that delegates to another check (like we have) and an AlwaysAllow check that doesn't (it just returns true)
Would this name change be future-proofing ourselves? or which part of the design would we just have a function that returns true

server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -0,0 +1,73 @@
...
+    metricRegistry.register("permissionCheckDeniedCount", new Counter());
+    alwaysAllow = new AlwaysAllowDelegatingPermissionCheck(metricRegistry, alwaysFail);
+  }
+
+  @Test public void delegateIsAllowedReturnsTrue() {
+    when(alwaysFail.isAllowed(principal, Action.ADD, target)).thenReturn(true);
+
+    boolean permitted = alwaysAllow.isAllowed(principal, Action.ADD, target);
+
+    assertThat(permitted).isEqualTo(true);
```
mmontgomery-square, line 46: assertThat(boolean) returns a BooleanAssertion which has an isTrue() method, I think that would be a (very slight) improvement over an equality check (more of a style thing).

server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -0,0 +1,73 @@
...
+    assertThat(metricRegistry.counter("permissionCheckCount").getCount()).isEqualTo(1);
+    assertThat(metricRegistry.counter("permissionCheckDeniedCount").getCount()).isEqualTo(0);
+  }
+
+  @Test public void delegateIsAllowedReturnsFalse() {
+    when(alwaysFail.isAllowed(principal, Action.ADD, target)).thenReturn(false);
```
mmontgomery-square, line 52: Mockito includes an `any()` matcher which is a nice way of communicating to the reader "I don't care about the actual values of the parameters". For cases like this where the desired behavior is to return false no matter what it can make the intent more clear.

server/src/main/java/keywhiz/service/permissions/PermissionCheckMetricsNames.java
```diff
@@ -0,0 +1,7 @@
+package keywhiz.service.permissions;
+
+public final class PermissionCheckMetricsNames {
+  private PermissionCheckMetricsNames() {}
+
+  public static final String PermissionCheckHistogram = "PermissionCheckHistogram";
+}
```
mmontgomery-square, line 7: Having a direct discussion re: metric naming strategy.

server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    boolean permitted = alwaysAllow.isAllowed(principal, Action.ADD, target);
+
+    assertThat(permitted);
+    assertThat(metricRegistry.histogram(PermissionCheckMetricsNames.PermissionCheckHistogram).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(PermissionCheckMetricsNames.PermissionCheckHistogram).getSnapshot().getMean()).isEqualTo(1);
+  }
+
+  @Test public void delegateIsAllowedReturnsFalse() {
+    when(delegate.isAllowed(any(), any(), any())).thenReturn(false);
```
mmontgomery-square, line 51: Totally a style thing, no action needed on your part:

Personally, I find mock programming to be pretty noisy, and I think all the when/any/thenReturn etc. can obscure the underlying intent or behavior. In this case, I might replace this line with a helper method like:

```
isAllowedReturns(true);
```

Short and sweet, without all the Mockito syntax cruft, and (I think) the behavior is totally obvious. Or, maybe we should throw a "delegate" in there too.

Same goes for other stuff in tests, too: if you have a ton of mostly-identical setup/execution code that only varies in a couple different ways, a helper method can keep the top-level test code concise and readable. My experience is that a lot of people think of testing code as a sort of second-class citizen where you don't need to be as tidy as actual application code, but I believe tests should be held to the same readability standards since they're essentially documentation-as-code.

Speaking of which, I like to use long test names that include not just the "situation" (as you've done here) but also the expected behavior. Something like "isAllowedReturnsTrueWhenDelegateReturnsFalse" is very literal and verbose, but something like "allowsWhenDelegateDenies" sounds a bit more human.

server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    boolean permitted = alwaysAllow.isAllowed(principal, Action.ADD, target);
+
+    assertThat(metricRegistry.histogram(PermissionCheckMetricsNames.PermissionCheckHistogram).getCount()).isEqualTo(1);
+    assertThat(metricRegistry.histogram(PermissionCheckMetricsNames.PermissionCheckHistogram).getSnapshot().getMean()).isEqualTo(0);
```
mmontgomery-square, line 56: I've seen multiple major outages have delayed detection because someone messed up a merge and a critical metric line was erased and the metric stopped being emitted. A lot of people might turn up their noses at unit testing metrics, but if those metrics are what allows you to monitor critical behavior then I think it's also critical that they be tested, if only to confirm that they're still being emitted.

server/src/test/java/keywhiz/service/permission/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -30,42 +29,71 @@ public class AlwaysAllowDelegatingPermissionCheckTest {
...
+  String checkAllowedOrThrowExceptionMetricName;
 
   @Before
   public void setUp() {
     metricRegistry = new MetricRegistry();
     alwaysAllow = new AlwaysAllowDelegatingPermissionCheck(metricRegistry, delegate);
+    isAllowedSuccessMetricName = MetricRegistry.name(AlwaysAllowDelegatingPermissionCheck.class, "isAllowed", "success", "histogram");
+    isAllowedFailureMetricName = MetricRegistry.name(AlwaysAllowDelegatingPermissionCheck.class, "isAllowed", "failure", "histogram");
+    checkAllowedOrThrowSuccessMetricName = MetricRegistry.name(AlwaysAllowDelegatingPermissionCheck.class, "checkAllowedOrThrow", "success", "histogram");
+    checkAllowedOrThrowExceptionMetricName = MetricRegistry.name(AlwaysAllowDelegatingPermissionCheck.class, "checkAllowedOrThrow", "exception", "histogram");
```
mmontgomery-square, line 44: Ah, something super subtle here:

So, we're using the same method of constructing the metric names as the actual source code (invoking MetricRegistry.name()). Now, imagine there's a bug that's introduced into that method which changes how the metric names are built: since the test and the code both use it they won't notice a difference, but the actual emitted metric will be different (which could throw off some alarms). I remember one of the first things I learned at a previous role was that if you're going to do assertions against the final value of a constant expression, you should actually use the literal result as your expected value rather than reconstructing it.

In other words, if MetricRegistry.name("foo", "bar") results in the metric name "foo.bar", you should declare this somewhere in the test:

```
fooBarMetricName = "foo.bar";
```

That way the expected result is hard-coded and immune to changes in how the code works, so you'll know right away if something breaks.

Separately, if these are constants that are used across multiple tests there's no need to keep re-declaring them in @ Before (which is the per-test setup method). You could declare them as plain ol' private static final constants, or you could make them private static and use a static method marked with @ BeforeClass to assign them once for the whole test class. A method (needs to be static, no args, public, returns void) marked with @ BeforeClass is called once before any of the tests in the current test class are run.

mmontgomery-square (approved):

PR #1079, opened 2022-06-10
added HelloWorld testing annotation (will remove later) + AutomationClientPermissionCheck

violetd12:


server/src/main/java/keywhiz/service/permissions/AutomationClientDelegatingPermissionCheck.java
```diff
@@ -0,0 +1,76 @@
...
+import com.codahale.metrics.MetricRegistry;
+import com.google.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AutomationClientDelegatingPermissionCheck implements PermissionCheck{
+  private static final Logger logger = LoggerFactory.getLogger(
+      AutomationClientDelegatingPermissionCheck.class);
+
+  private PermissionCheck delegate;
```
mmontgomery-square, line 12: So for this one I don't think we're going to need a delegate, this can just be a "terminal" permission check. I guess I would say that a permission check either has permission logic of its own (which this one does) or relying on a delegate's logic, but not both. The pass-through and always-allow are in the second category since they don't have any (non-trivial) permission check logic of their own.

server/src/main/java/keywhiz/service/permissions/AutomationClientDelegatingPermissionCheck.java
```diff
@@ -0,0 +1,76 @@
...
+      hasPermission = delegate.isAllowed(source, action, target);
+    }
+
+    int hasPermissionSuccessMetricInt = hasPermission ? 1 : 0;
+    String successMetricName = MetricRegistry.name(AutomationClientDelegatingPermissionCheck.class, "isAllowed", "success", "histogram");
+    metricRegistry.histogram(successMetricName).update(hasPermissionSuccessMetricInt);
+
+    int hasPermissionFailureMetricInt = hasPermission ? 0 : 1;
+    String failureMetricName = MetricRegistry.name(AutomationClientDelegatingPermissionCheck.class, "isAllowed", "failure", "histogram");
+    metricRegistry.histogram(failureMetricName).update(hasPermissionFailureMetricInt);
```
mmontgomery-square, line 36: Just to shorten things up a bit, you could move all the metrics stuff into an emitMetrics() helper method. That's actually one of the really nice things about the built-in metrics annotations is that they really hide all of this boilerplate and allow the method to just focus on the business logic.

server/src/main/java/keywhiz/service/permissions/AutomationClientDelegatingPermissionCheck.java
```diff
@@ -0,0 +1,76 @@
...
+  private boolean isAutomation(KeywhizPrincipal source) {
+    //TODO(violet): authenticate source as automation, return true for now
+
+    return true;
+  }
+
+  @Override
+  public void checkAllowedOrThrow(KeywhizPrincipal source, String action, Object target) {
+    String successMetricName = MetricRegistry.name(AutomationClientDelegatingPermissionCheck.class, "checkAllowedOrThrow", "success", "histogram");
+    String exceptionMetricName = MetricRegistry.name(AutomationClientDelegatingPermissionCheck.class, "checkAllowedOrThrow", "exception", "histogram");
```
mmontgomery-square, line 54: If we gave up on having custom metric names for each entry point ("isAllowed" vs. "checkAllowedOrThrow"), could we just use the default implementation of this method (which just delegates to isAllowed?)

violetd12: what if we move all the custom metrics emission into a method defined in PermissionCheck interface directly.

violetd12: would have to add MetricRegistry as a parameter...could be difficult

PR #1086, opened 2022-06-16
Checking in generated model source to eliminate build-time dependency on database

mmontgomery-square:


model/pom.xml
```diff
@@ -96,19 +96,29 @@
...
+              <goal>add-source</goal>
             </goals>
+            <configuration>
+              <sources>
+                <source>src-generated/main/java</source>
+              </sources>
+            </configuration>
           </execution>
         </executions>
+      </plugin>
```
mmontgomery-square, line 116: GitHub really whiffed on the diff here. Maven doesn't have first-class support for multiple source directories, so since we'd like to keep the hand-written and generated code in separate folders we're using this plugin (which appears to be the standard approach to solving this problem).

I picked src-generated because it will appear alphabetically next to the usual src folder but also contains the word "generated".

model/pom.xml
```diff
@@ -124,16 +134,6 @@
...
-        <executions>
-          <execution>
-            <phase>generate-sources</phase>
-            <goals>
-              <goal>migrate</goal>
-            </goals>
-          </execution>
-        </executions>
-
         <configuration>
```
mmontgomery-square, line 137: Rather than tying these goals into the standard build lifecycle via the executions tag, we're omitting that section so that they can only be invoked manually.

model/pom.xml
```diff
@@ -215,6 +204,7 @@
             </database>
             <target>
               <packageName>keywhiz.jooq</packageName>
+              <directory>src-generated/main/java</directory>
             </target>
```
mmontgomery-square, line 208: By default, Jooq jumps generated code under the target/ folder, which is the standard location for Maven's build outputs. However, since it's generally considered a best practice to never check in anything in that folder, we're redirecting Jooq's output to our new src-generated folder.

model/regenerate.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -
+
+mvn \
+validate \
+org.codehaus.mojo:sql-maven-plugin:execute \
+org.flywaydb:flyway-maven-plugin:migrate \
+org.jooq:jooq-codegen-maven:generate
```
mmontgomery-square, line 7: This is our new script which can be manually invoked to re-generate the Java-to-database glue code. I believe validate is necessary because without it Maven will not attempt to ensure that all declared dependencies (including the build plugins themselves) are available locally in the case where you're invoking a build plugin directly.

server/pom.xml
```diff
@@ -11,6 +11,67 @@
...
+            <executions>
+              <execution>
+                <id>create-keywhiz-database</id>
+                <phase>none</phase>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
```
mmontgomery-square, line 31: It turns out that about half of the Keywhiz tests are actually integration tests that rely on being able to connect to a local MySQL database. It also turns out that the server sub-project implicitly relies on the model sub-project (which builds first because the server sub-project depends on it) to create the test database rather than handling this directly. Now that that's no longer the case, the database setup "code" has been copied over to this POM.

That's all fine and dandy, but there are some cases (containerized builds, for example, thus the profile name) where we don't want to have a database dependency at all. Although we can attain this for the most part by bypassing tests via passing `-Dmaven.test.skip=true` to Maven on the command line, it turns out this property is really more of a by-convention thing rather than something first-class that actually de-activates all testing-related lifecycle phases. In other words, it doesn't affect our plugin in any way, which means the plugin will run and fail the build if the database is unreachable. And so, in order to gives ourselves an out, we can use a profile to manually disable the plugin (phase `none`) and then select that profile from the command line via `-P container`.

server/pom.xml
```diff
@@ -215,6 +276,31 @@
...
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>sql-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>create-keywhiz-database</id>
+            <phase>process-test-classes</phase>
+            <goals>
+              <goal>execute</goal>
+            </goals>
+          </execution>
```
mmontgomery-square, line 290: The ID is an addition beyond the original configuration in model/pom.xml. It's necessary so that we can match the profile configuration with this specific execution.

Also, unlike Ant, Maven doesn't have an explicit notion of before/after hooks for each lifecycle phase, so if we need to do some pre-test setup we just need to arbitrarily pick one of the preceding phases and slap this thing in there.

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandIntegrationTest.java
```diff
@@ -13,16 +13,21 @@
...
 import org.jooq.Record1;
+import org.junit.ClassRule;
 import org.junit.Test;
+import org.junit.rules.RuleChain;
 import org.junit.runner.RunWith;
 
 @RunWith(KeywhizTestRunner.class)
-public class BackfillOwnershipCommandTest {
+public class BackfillOwnershipCommandIntegrationTest {
+  @ClassRule public static final RuleChain chain = IntegrationTestRule.rule();
```
mmontgomery-square, line 29: There were two tests (both backfills) which weren't properly configured to use/migrate the database and depended on execution order to succeed. Explicitly adding the integration test rule (which in turn runs Flyway's migration logic) fixed that issue. Also, since we only needed to touch two tests, I decided to rename them to be actual integration tests to be more clear (and note that there's another 150 or so which should get the same treatment, but that's a fight for another day).

model/regenerate.sh
mbyczkowski: would be nice if we a) added to README to explain that this should be run occasionally (e.q. when we bump jooq and/or change schema) and b) added a step in CI where we attempt to run the code generation and ensure there's no file diff.

model/pom.xml
mbyczkowski: seems reasonable 👍

server/src/test/java/keywhiz/commands/BackfillOwnershipCommandIntegrationTest.java
mbyczkowski: good catch!

server/pom.xml
mbyczkowski: maybe add a small comment about "phase `none`" that it disables the plugin?

server/pom.xml
```diff
@@ -11,6 +11,67 @@
...
+                <phase>none</phase>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>
+        no-custom-mysql-port
```
mbyczkowski, line 34: is the `no-custom-mysql-port` just copied from a different place or entirely new?

model/regenerate.sh
eknapp-square: Could it be a test? Have the test regen the files and diff. If the test fails, it prints instructions for how to fix.

server/pom.xml
mmontgomery-square: That's copied from the model sub-project. I figured the flexibility made sense to preserve even though I don't believe we're taking advantage of it.

server/pom.xml
mmontgomery-square: Done!

model/regenerate.sh
mmontgomery-square: Are you referring to Travis CI or our internal build CI? That might be doable, assuming we use a similar setup to the server sub-project where we're using a profile.

server/pom.xml
mbyczkowski: could be just some old cruft

mbyczkowski (approved):

PR #1088, opened 2022-06-21
Embedding maven.test.skip property into container profile

mmontgomery-square:


server/pom.xml
```diff
@@ -14,6 +14,9 @@
   <profiles>
     <profile>
       <id>container</id>
+      <properties>
+        <maven.test.skip>true</maven.test.skip>
+      </properties>
       <build>
```
mmontgomery-square, line 20: We're setting this on the command line internally but really it should be folded into this profile so that it's self-contained.

mbyczkowski (approved):

PR #1089, opened 2022-06-23
added OwnershipPermissionCheck

violetd12:
TODO: tests


server/src/main/java/keywhiz/service/permissions/OwnershipPermissionCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    this.metricRegistry = metricRegistry;
+    this.aclDAO = aclDAOFactory.readwrite();
+  }
+
+  public boolean isAllowed(Object source, String action, Object target) {
+    boolean hasPermission = false;
+
+    if (isClient(source) && isSecret(target)) {
+      Set<Group> clientGroups = aclDAO.getGroupsFor((Client) source);
+      String secretOwner =((Secret) target).getOwner();
```
mmontgomery-square, line 34: We should really get Checkstyle going for this project, but in the meantime make sure you always put a space on both sides of an equal sign.

violetd12: there is. just realized that the style guide warned me about the if statement simplification for the help functions. but not the spacing

server/src/test/java/keywhiz/service/permission/OwnershipPermissionCheckTest.java
```diff
@@ -0,0 +1,196 @@
...
+    metricRegistry = new MetricRegistry();
+    when(aclDAOFactory.readwrite()).thenReturn(aclDAO);
+    ownershipPermissionCheck = new OwnershipPermissionCheck(metricRegistry, aclDAOFactory);
+  }
+
+  @Test
+  public void testIsAllowedWhenSourceIsNotAClient() {
+    boolean permitted = ownershipPermissionCheck.isAllowed(USER, Action.CREATE, SECRET_WITH_OWNER);
+
+    assertThat(permitted).isEqualTo(false);
```
violetd12, line 66: Had to make this explicitly equal to false because it appears that `assertThat(permitted)` always returns true

PR #1091, opened 2022-06-27
Implement AnyPermissionCheck

john-shieh:
https://jira.sqprod.co/browse/CISM-3874

### DIFF
--- a/server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
+++ b/server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
@@ -1,0 +1,33 @@
+package keywhiz.service.permissions;
+
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AnyPermissionCheck implements PermissionCheck {
+
+  private static final Logger logger = LoggerFactory.getLogger(AnyPermissionCheck.class);
+
+  private List<PermissionCheck> subordinateChecks;
+
+  public AnyPermissionCheck(List<PermissionCheck> subordinateChecks) {
+    this.subordinateChecks = subordinateChecks;
+  }
+
+  public boolean isAllowed(Object source, String action, Object target) {
+    boolean hasPermission = false;
+
+    for (PermissionCheck subordinateCheck : subordinateChecks) {
+      if (subordinateCheck.isAllowed(source, action, target)) {
+        hasPermission = true;
+        break;
+      }
+    }
+
+    logger.info(
+        String.format("isAllowed Actor: %s, Action: %s, Target: %s, Result: %s", source, action, target,
+            hasPermission));
+
+    return hasPermission;
+  }
+}
--- a/server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
+++ b/server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
@@ -1,5 +1,6 @@
 package keywhiz.service.permissions;
 
+import com.google.common.collect.ImmutableList;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -11,7 +12,7 @@ public class AnyPermissionCheck implements PermissionCheck {
   private List<PermissionCheck> subordinateChecks;
 
   public AnyPermissionCheck(List<PermissionCheck> subordinateChecks) {
-    this.subordinateChecks = subordinateChecks;
+    this.subordinateChecks = ImmutableList.copyOf(subordinateChecks);
   }
 
   public boolean isAllowed(Object source, String action, Object target) {
--- a/server/src/test/java/keywhiz/service/permission/AnyPermissionCheckTest.java
+++ b/server/src/test/java/keywhiz/service/permission/AnyPermissionCheckTest.java
@@ -1,0 +1,74 @@
+package keywhiz.service.permission;
+
+import java.util.Collections;
+import java.util.List;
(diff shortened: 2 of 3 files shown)


server/src/test/java/keywhiz/service/permission/AnyPermissionCheckTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static org.mockito.Mockito.when;
+
+public class AnyPermissionCheckTest {
+
+  @Rule public MockitoRule mockito = MockitoJUnit.rule();
+  @Mock private PermissionCheck delegate1;
+  @Mock private PermissionCheck delegate2;
+
+  private static Object source;
+  private static Object target;
```
mmontgomery-square, line 27: I'm torn on whether they should be initialized (if I'm reading correctly, they are not and thus remain null). I can't think of a case where either should legitimately be null, but OTOH maybe the composite should be "dumb" and just strictly delegate to the subordinates (great term, by the way)? I'd be fine leaving as-is.

mmontgomery-square (approved):

server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
```diff
@@ -0,0 +1,33 @@
...
+
+    for (PermissionCheck subordinateCheck : subordinateChecks) {
+      if (subordinateCheck.isAllowed(source, action, target)) {
+        hasPermission = true;
+        break;
+      }
+    }
+
+    logger.info(
+        String.format("isAllowed Actor: %s, Action: %s, Target: %s, Result: %s", source, action, target,
```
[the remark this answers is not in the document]
john-shieh: Here is the original reasoning that I found: https://github.com/square/keywhiz/pull/1068#discussion_r878573997

server/src/test/java/keywhiz/service/permission/AnyPermissionCheckTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test public void isAllowedReturnsTrueWithEmptyCheckList() {
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(Collections.emptyList());
+
+    assertFalse(anyPermissionCheck.isAllowed(source, Action.ADD, target));
+  }
+
+  @Test public void isAllowedReturnsFalseWhenAllDelegatesReturnFalse() {
+    when(delegate1.isAllowed(any(), any(), any())).thenReturn(false);
+    when(delegate2.isAllowed(any(), any(), any())).thenReturn(false);
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(List.of(delegate1, delegate2));
```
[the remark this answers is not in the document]
john-shieh: Good point. I've created a separate function for it.

server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
```diff
@@ -0,0 +1,34 @@
...
+  private List<PermissionCheck> subordinateChecks;
+
+  public AnyPermissionCheck(List<PermissionCheck> subordinateChecks) {
+    this.subordinateChecks = ImmutableList.copyOf(subordinateChecks);
+  }
+
+  public boolean isAllowed(Object source, String action, Object target) {
+    boolean hasPermission = false;
+
+    for (PermissionCheck subordinateCheck : subordinateChecks) {
```
isemaya-square, line 21: The functionality here looks great, so this is just food for thought/thinking about alternative ways to accomplish the same task: if all you're looking for is that one of the subordinate checks returns true (rather than a specific one) you could replace this with a steam like:

```
boolean hasPermission  = subordinateChecks.stream().anyMatch(subordinateCheck
                     -> subordinateCheck.isAllowed(source, action, target));
```

(look like the tradeoff here is that streams may not short circuit: https://stackoverflow.com/questions/32192846/java-8-stream-short-circuit#:~:text=The%20operations%20allMatch%20%2C%20noneMatch%20and,using%20these%20may%20not%20terminate.)

I personally like streams because I find them to be an intuitive way to think about logic - don't feel obligated to change this at all, especially since the method is pretty simple - everyone has their own coding style!

lavanyaharinarayan (approved):

server/src/test/java/keywhiz/service/permission/AnyPermissionCheckTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Test public void isAllowedReturnsTrueWhenOneDelegateReturnsTrue() {
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(createDelegatesList(true, false));
+
+    assertTrue(anyPermissionCheck.isAllowed(source, Action.ADD, target));
+  }
+
+  @Test public void checkAllowedOrThrowThrowsExceptionWithEmptyCheckList() {
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(Collections.emptyList());
+
+    assertThatThrownBy(() -> anyPermissionCheck.checkAllowedOrThrow(source, Action.ADD, target))
```
isemaya-square, line 50: nit: if you want, you can also check for the exception message (I believe you can call `.hasMessage(...)` after `.isInstanceOf`

isemaya-square (approved):

server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
john-shieh: I hadn't thought about using streams but will keep that in mind for next time!

PR #1092, opened 2022-06-28
added inferring secret owner option in config. added inferring secret owner upon creation

violetd12:
TODO: 

- decide what to do when the client belongs to 0 group
- decide what to do when the client belongs to multiple groups
- tests


server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -153,30 +166,6 @@ public class SecretResourceTest {
   //---------------------------------------------------------------------------------------
   // createOrUpdateSecret
   //---------------------------------------------------------------------------------------
-
```
violetd12, line 156: the test config file now specifies that the owner should be inferred so this test does not make sense any more

api/src/main/java/keywhiz/api/automation/v2/CreateOrUpdateSecretRequestV2.java
```diff
@@ -33,6 +34,10 @@ public static Builder builder() {
     public abstract Builder type(String type);
     public abstract Builder expiry(long expiry);
 
+    public CreateOrUpdateSecretRequestV2.Builder owner(String owner) {
```
violetd12, line 37: Originally, owner couldn't be updated with this function

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -153,30 +166,6 @@ public class SecretResourceTest {
...
-        .content(encode("bar"))
-        .description(UUID.randomUUID().toString())
-        .build();
-
-    assertEquals(201, createOrUpdate(updateRequest, secretName).code());
-
-    SecretDetailResponseV2 updatedDetails = lookup(secretName);
-    assertNull(updatedDetails.owner());
-  }
-
```
violetd12, line 179: the test config file now specifies that the owner should be inferred so this test does not make sense any more

violetd12: I will mark it as @Ignored actually

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -100,6 +100,9 @@ public class KeywhizConfig extends Configuration {
   @JsonProperty
   private Long maximumSecretSizeInBytesInclusive;
 
+  @JsonProperty
+  private boolean inferSecretOwnerUponCreation;
```
mmontgomery-square, line 104: We want to be careful about using a boolean here, since we'll be locked into two different options forever. How about something like:

```
enum NewSecretOwnershipStrategy {
  INFER_FROM_CLIENT,
  NONE
}

```

api/src/main/java/keywhiz/api/automation/v2/CreateOrUpdateSecretRequestV2.java
```diff
@@ -33,6 +34,10 @@ public static Builder builder() {
...
+      return owner(owner);
+    }
```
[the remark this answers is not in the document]
violetd12: having trouble making this work with public abstract. I don't quite understand how to make the change compile. I get the error `[AutoValueBuilderWhatProp] Method owner does not correspond to a property method of keywhiz.api.automation.v2.CreateOrUpdateSecretRequestV2`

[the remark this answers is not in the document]
violetd12: But since this part is not the most important change for this PR, I will revert it for now and circle back to it later

server/src/test/java/keywhiz/IntegrationTestRule.java
```diff
@@ -29,4 +29,11 @@ public static RuleChain rule() {
         .outerRule(new MigrationsRule())
         .around(new DropwizardAppRule<>(KeywhizService.class, configPath));
   }
+
+  public static RuleChain rule(String configFile) {
```
violetd12, line 33: overload constructor so that I can specify different config files for testing `createSecret` in SecretResource

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
```diff
@@ -0,0 +1,94 @@
...
+import okhttp3.Response;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SecretResourceConfigNewSecretOwnershipStrategyInferTest {
```
violetd12, line 24: Currently, this test file as well as `SecretResourceConfigNewSecretOwnershipStrategyNoneTest.java` only test `createSecret` function. When `createOrUpdate` function is also modified to infer secret owner, these two test files will include those tests as well. Open to naming suggestions for these test files.

mmontgomery-square:
Duh, I had a review all put together and then forgot to post it :/ I'll go over this again now.

server/src/test/java/keywhiz/KeywhizConfigTest.java
```diff
@@ -48,6 +51,27 @@ public void missingMaximumSecretLengthIsNull() {
     assertNull(config.getMaximumSecretSizeInBytesInclusive());
   }
 
+  @Test
+  public void parseMissingNewSecretOwnershipStrategy() {
+    KeywhizConfig config = loadConfig("new-secret-ownership-strategy-missing.yaml");
+    assertThat(config.getNewSecretOwnershipStrategy()).isEqualTo(
+        KeywhizConfig.NewSecretOwnershipStrategy.NONE);
+  }
```
mmontgomery-square, line 59: You might want to add "DefaultsToNone" to the end of the test name just document what the expected behavior is.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
mmontgomery-square: As much as I like long, descriptive names for things, that is certainly a mouthful. Maybe SecretResourceInferOwnerTest and SecretResourceNoOwnerTest?

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+  OkHttpClient mutualSslClient;
+
+  @ClassRule public static final RuleChain chain = IntegrationTestRule.rule("keywhiz-test-newSecretOwnershipStrategyInfer.yaml");
+
+  @Before public void setUp() {
+    mutualSslClient = TestClients.mutualSslClient();
+    secretResourceTest = new SecretResourceTest();
+    secretResourceTest.mutualSslClient = mutualSslClient;
+  }
```
mmontgomery-square, line 40: Creating an instance of a test inside another test is definitely thinking outside the box, but that's a pattern that's usually avoided. If there are related tests that share helper methods/resources then typically you can either have a common abstract base class that they inherit from (say, AbstractSecretResourceTestBase, and IIRC that class can have @ Before etc. methods to properly initialize its own fields, although I'd have to double-check the JUnit docs on that). The other one would be to move the shared utility method to a dedicated class (SecretResourceTestHelper?) and have them both use that instead.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
```diff
@@ -0,0 +1,94 @@
...
+    assertThat(response.code()).isEqualTo(RESPONSE_CODE_CREATED);
+
+    SecretDetailResponseV2 details = secretResourceTest.lookup(secretName);
+    assertThat(details.name()).isEqualTo(secretName);
+    assertThat(details.owner()).isEqualTo(ownerName);
+  }
+
+  private String randomString() {
+    return UUID.randomUUID().toString();
+  }
```
mmontgomery-square, line 83: I must have written this exact same method over 1,000 times in my career.

Random data is nice because it ensures that you won't have any collisions if you're inserting things into, say, a persistent database (integration testing against staging, for example) no matter how many times you run the tests. However, since all GUIDs more or less look like gibberish, it can sometimes be handy to add a human-readable prefix to it to indicate that it's coming from an integration test (doesn't really apply here, but something I've run into before).

mmontgomery-square (commented):
This is really fantastic work, thank you for being so thorough!

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
violetd12: Ah ok I will look around other files and see how I can refactor the tests. I am fairly certain that creating a new instance of another test class is a common thing among several other tests .

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceConfigNewSecretOwnershipStrategyInferTest.java
violetd12: makes sense. I wrote the method also by mirroring the original SecretResourceTest but seems like a little extra here since we are not really worried about collision and all we want is some name to write assertions

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceInferOwnerTest.java
```diff
@@ -0,0 +1,95 @@
...
+        .content(encode(SECRET_CONTENT))
+        .build();
+
+    Response response = secretResourceTest.create(request);
+    assertThat(response.code()).isEqualTo(RESPONSE_CODE_CREATED);
+
+    SecretDetailResponseV2 details = secretResourceTest.lookup(secretName);
+    assertThat(details.name()).isEqualTo(secretName);
+    assertThat(details.owner()).isEqualTo(SECRET_OWNER);
+  }
```
isemaya-square, line 80: Even if you're still not sure what functionality you'll have when `clientGroups.size() != 1` (in `SecretResource`), you could still write tests for them, even if they currently fail!

violetd12: added!

mmontgomery-square (approved):

PR #1098, opened 2022-07-01
try this to run tests in GH Actions

mbyczkowski:


isemaya-square (approved):

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,31 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
```
mmontgomery-square, line 10: Even though it's kinda obvious that the first step is just checking out the code, I might add a `name` entry for it just to be consistent with the others.

mmontgomery-square (approved):

PR #1102, opened 2022-07-12
Add permission checks to SecretResource

john-shieh:
https://jira.sqprod.co/browse/CISM-3876

Instead of adding permission checks at the DAO level with overloaded methods like in https://github.com/square/keywhiz/pull/1094, I decided to move all the permission checks to the Resource level so the principal (automation client) didn't need to be passed down. It also simplified the code because most of the resource methods already retrieved Secrets from the DAO, so a permission check could be added right after.

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
+++ b/server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
@@ -46,6 +46,7 @@ import keywhiz.api.model.SanitizedSecretWithGroupsListAndCursor;
 import keywhiz.api.model.Secret;
 import keywhiz.api.model.SecretContent;
 import keywhiz.api.model.SecretRetrievalCursor;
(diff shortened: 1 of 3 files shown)


server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -118,6 +123,8 @@ public class SecretResource {
   @Consumes(APPLICATION_JSON)
   public Response createSecret(@Auth AutomationClient automationClient,
       @Valid CreateSecretRequestV2 request) {
+    permissionCheck.checkAllowedOrThrow(automationClient, Action.CREATE);
```
mmontgomery-square, line 126: It occurs to me that we could be more specific by including the type of thing being created, but we can deal with that later if/when we expand to other resource types.

mmontgomery-square (approved):

PR #823, opened 2021-03-26
quoting reserved mysql phrase

pouyan021:
not quoting the groups makes this script not runnable

PR #1105, opened 2022-07-13
Bump MySQL in CI to 8.0

mbyczkowski:
This is a follow-up on #823

### DIFF
--- a/server/src/main/java/keywhiz/utility/DSLContexts.java
+++ b/server/src/main/java/keywhiz/utility/DSLContexts.java
@@ -21,7 +21,7 @@ import java.sql.SQLException;
 import javax.sql.DataSource;
 import org.jooq.DSLContext;
 import org.jooq.SQLDialect;
-import org.jooq.conf.RenderNameStyle;
+import org.jooq.conf.RenderQuotedNames;
 import org.jooq.conf.Settings;
 import org.jooq.impl.DSL;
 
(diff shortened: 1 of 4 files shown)


server/src/main/java/keywhiz/utility/DSLContexts.java
```diff
@@ -49,6 +49,6 @@ public static DSLContext databaseAgnostic(DataSource dataSource) throws SQLExcep
     return DSL.using(dataSource, dialect,
             new Settings()
                 .withRenderSchema(false)
-                .withRenderNameStyle(RenderNameStyle.AS_IS));
+                .withRenderQuotedNames(RenderQuotedNames.EXPLICIT_DEFAULT_QUOTED));
```
mmontgomery-square, line 52: According to the documentation, these flags are independent of each other:

https://www.jooq.org/doc/latest/manual/sql-building/dsl-context/custom-settings/settings-name-style/

Should we be adding a new line rather than replacing the existing one? I guess it's not clear to me what problem un-setting the existing setting is solving.

mmontgomery-square: Reading a bit further, AS_IS and EXPLICIT_DEFAULT_QUOTED are both (supposed to be) the default values.

mmontgomery-square (approved):

PR #1106, opened 2022-07-13
Add permission checks to backfill row hmac resource

john-shieh:
https://jira.sqprod.co/browse/CISM-3876

### DIFF
--- a/server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
+++ b/server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
@@ -27,6 +27,8 @@ import keywhiz.log.AuditLog;
 import keywhiz.log.Event;
 import keywhiz.log.EventTag;
 import keywhiz.service.crypto.RowHmacGenerator;
+import keywhiz.service.permissions.Action;
+import keywhiz.service.permissions.PermissionCheck;
 import org.jooq.DSLContext;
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/resources/automation/v2/BackfillRowHmacResource.java
```diff
@@ -74,6 +79,8 @@ public void backfillSecretRowHmacByName(
       @Auth AutomationClient automationClient,
       @PathParam("secretName") String secretName,
       @QueryParam("force") boolean force) {
+    permissionCheck.checkAllowedOrThrow(automationClient, Action.UPDATE);
```
john-shieh, line 82: I'm not sure how we should model permission checks for this resource since each method acts on database rows. The current check is simple, but coarse, similar to https://github.com/square/keywhiz/blob/6d151a884e3041c745a7e4a12320c0eb6e384191/server/src/main/java/keywhiz/service/resources/automation/v2/ClientResource.java#L148-L154. For more granularity, SecretRecord, SecretContentRecord, and ClientRecord have corresponding mappers, so they could be converted and passed into the permission check although SecretContent doesn't have a direct owner. The accessgrant and membership tables don't have mappers, but maybe their components could be converted separately and checked (e.g. check the group and secret for every accessgrant record).

mmontgomery-square (approved):

PR #1117, opened 2022-08-04
Introducing top-level permission check method for target type

mmontgomery-square:


server/src/main/java/keywhiz/service/permissions/AlwaysAllowDelegatingPermissionCheck.java
```diff
@@ -32,14 +33,41 @@ public boolean isAllowed(Object source, String action, Object target) {
...
 
+  @Override
+  public boolean isAllowedForTargetType(Object source, String action, Class<?> targetType) {
+    boolean hasPermission = delegate.isAllowedForTargetType(source, action, targetType);
+
+    emitHistogramMetrics(hasPermission);
+
+    logger.info(
+        String.format("isAllowedByType Actor: %s, Action: %s, Target type: %s, Result: %s", source, action, targetType,
+            hasPermission));
```
mmontgomery-square, line 44: I can see a PermissionCheckContext class in our future, where you set all of the fields for which you have data available via a builder and the toString() belongs to the context class so that we don't have to manually print it out here.

server/src/main/java/keywhiz/service/permissions/AlwaysAllowDelegatingPermissionCheck.java
```diff
@@ -32,14 +33,41 @@ public boolean isAllowed(Object source, String action, Object target) {
...
   @Override
   public void checkAllowedOrThrow(Object source, String action, Object target) {
     Boolean isPermitted;
     try {
       delegate.checkAllowedOrThrow(source, action, target);
       isPermitted = true;
     } catch (RuntimeException e) {
-      logger.error(String.format("checkAllowedOrThrow Actor: %s, Action: %s, Target: %s throws exception", source, action, target),e);
+      logger.error(String.format("checkAllowedOrThrow Actor: %s, Action: %s, Target: %s throws exception", source, action, target), e);
+      isPermitted = false;
```
mmontgomery-square, line 57: Minor: just adding a space after the comma and the end.

server/src/main/java/keywhiz/service/permissions/AnyPermissionCheck.java
```diff
@@ -3,7 +3,6 @@
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
-import javax.inject.Inject;
 import org.slf4j.Logger;
```
mmontgomery-square, line 6: This was unused.

server/src/main/java/keywhiz/service/permissions/AutomationClientPermissionCheck.java
```diff
@@ -6,7 +6,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class AutomationClientPermissionCheck implements PermissionCheck{
+public class AutomationClientPermissionCheck implements PermissionCheck {
   private static final Logger logger = LoggerFactory.getLogger(AutomationClientPermissionCheck.class);
```
mmontgomery-square, line 10: Minor: adding a space before the brace.

server/src/main/java/keywhiz/service/permissions/PermissionCheck.java
```diff
@@ -3,9 +3,7 @@
 public interface PermissionCheck {
   boolean isAllowed(Object source, String action, Object target);
 
-  default boolean isAllowed(Object source, String action) {
-    return isAllowed(source, action, null);
-  }
+  boolean isAllowedForTargetType(Object source, String action, Class<?> targetType);
 
   default void checkAllowedOrThrow(Object source, String action, Object target){
```
mmontgomery-square, line 8: This is the crux of the change.

For our Secret creation permission check (in a future PR), we need an overload which takes in the type (class literal) of the target as a third argument. Unfortunately, this means that our existing isAllowed(Object, String, Object) signature is a superset (because Object is a superclass of Class) of our new overload, which makes it hard to reason about dispatch and also confuses Mockito, leading to unexpected behavior and easy-to-introduce bugs.

So, as a solution, I decided to change the method name of the overload from isAllowed to isAllowedForTargetType to eliminate any chance of conflict. I also decided to get rid of the default implementations of this method and the original isAllowed as a hint to implementers that they needed to reason about both of these cases separately.

Finally, with our new overload, the existing two-argument isAllowed(Object, String) isn't used anywhere else, so I removed it. In hindsight, it really didn't make much sense to have it automatically delegate to isAllowed(Object, String, Object) since those are separate questions, and because it could also just as easily delegate to isAllowedForTargetType(Object, String, Class).

server/src/main/java/keywhiz/service/resources/automation/v2/ClientResource.java
```diff
@@ -98,7 +98,7 @@ public class ClientResource {
   @Consumes(APPLICATION_JSON)
   public Response createClient(@Auth AutomationClient automationClient,
       @Valid CreateClientRequestV2 request) {
-    permissionCheck.checkAllowedOrThrow(automationClient, Action.CREATE);
+    permissionCheck.checkAllowedForTargetTypeOrThrow(automationClient, Action.CREATE, Client.class);
```
mmontgomery-square, line 102: All of the two-argument isAllowed calls now use the "ForTargetType" method instead and supply a third argument. We already had a follow-up task for this work, but the need to implement our Secret creation permission check just moved it forward in the schedule.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -326,7 +326,7 @@ public Iterable<String> secretListing(@Auth AutomationClient automationClient,
   public Iterable<SanitizedSecret> secretListingV2(@Auth AutomationClient automationClient,
       @QueryParam("idx") Integer idx, @QueryParam("num") Integer num,
       @DefaultValue("true") @QueryParam("newestFirst") boolean newestFirst) {
-    permissionCheck.checkAllowedOrThrow(automationClient, Action.READ);
+    permissionCheck.checkAllowedForTargetTypeOrThrow(automationClient, Action.READ, SanitizedSecret.class);
```
mmontgomery-square, line 330: Our code makes a distinction between SanitizedSecret and regular ol' Secret, so I decided that the check should as well (even if the implementation just coalesces everything into Secret, although this code wouldn't know/shouldn't assume that).

Also, note that phase 1 doesn't support field-level permissions (but phase 2 will), so for cases where we're just returning a single field such as secret name we just check permissions for the entire Secret class, which could be overly restrictive.

server/src/test/java/keywhiz/service/permissions/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -24,8 +23,8 @@ public class AlwaysAllowDelegatingPermissionCheckTest {
   private MetricRegistry metricRegistry;
   private AlwaysAllowDelegatingPermissionCheck alwaysAllow;
 
-  private static Objects target;
-  private static Objects principal;
+  private static Object target = new Object();
+  private static Object principal = new Object();
```
mmontgomery-square, line 28: I ran into an issue where (before renaming our new isAllowed overload) Mockito got confused because null is a sort of supertype-of-all-types and it wouldn't match properly. Using non-null values fixed this and I think will help ensure more correct behavior going forward.

server/src/test/java/keywhiz/service/permissions/AlwaysAllowDelegatingPermissionCheckTest.java
```diff
@@ -78,7 +77,7 @@ public void setUp() {
     assertThat(metricRegistry.histogram(CHECKALLOWEDORTHROW_EXCEPTION_METRIC_NAME).getSnapshot().getMean()).isEqualTo(0);
   }
 
-  @Test public void CheckAllowedOrThrowReturnsVoidWhenDelegateThrowException() {
+  @Test public void checkAllowedOrThrowReturnsVoidWhenDelegateThrowException() {
     doThrow(RuntimeException.class).when(delegate).checkAllowedOrThrow(any(), any(), any());
```
mmontgomery-square, line 81: Minor: lower-casing method name.

server/src/main/java/keywhiz/service/permissions/AlwaysFailPermissionCheck.java
```diff
@@ -9,4 +9,6 @@ public AlwaysFailPermissionCheck() {}
   public boolean isAllowed(Object source, String action, Object target) {
```
john-shieh, line 9: Will the AlwaysFailPermissionCheck be used anywhere?

john-shieh (approved):

violetd12 (approved):

server/src/main/java/keywhiz/service/permissions/AlwaysFailPermissionCheck.java
mmontgomery-square: I think it's currently only used in tests, but I believe we originally implemented it for completeness (if we have always allow, we should have always deny, essentially).

PR #1118, opened 2022-08-05
Implementing permission check to allow clients to create secrets

mmontgomery-square:


api/src/main/java/keywhiz/api/model/Client.java
```diff
@@ -87,7 +87,7 @@ public Client(@JsonProperty("id") long id,
       @JsonProperty("enabled") boolean enabled,
       @JsonProperty("automationAllowed") boolean automationAllowed) {
     this.id = id;
-    this.name = checkNotNull(name);
+    this.name = checkNotNull(name, "Client name must not be null");
     this.description = nullToEmpty(description);
```
mmontgomery-square, line 91: This is functionally unrelated to the change, but I found the error messages helpful while debugging errors related to creating test data.

server/src/main/java/keywhiz/service/permissions/AllowClientSecretCreationPermissionCheck.java
```diff
@@ -0,0 +1,18 @@
...
+  public boolean isAllowed(Object source, String action, Object target) {
+    return false;
+  }
+
+  @Override
+  public boolean isAllowedForTargetType(Object source, String action, Class<?> targetType) {
+    return
+        Client.class.isAssignableFrom(source.getClass()) &&
+        Secret.class.isAssignableFrom(targetType);
+  }
```
mmontgomery-square, line 17: Technically the first line is identical to `source instanceof Client`, but I liked the symmetry with the second line where `isAssignableFrom()` is required because the class is a variable (instanceof is compile-time, isAssignableFrom is run-time).

server/src/main/java/keywhiz/service/permissions/AllowClientSecretCreationPermissionCheck.java
```diff
@@ -0,0 +1,18 @@
...
+import keywhiz.api.model.Secret;
+
+public class AllowClientSecretCreationPermissionCheck implements PermissionCheck {
+  @Override
+  public boolean isAllowed(Object source, String action, Object target) {
+    return false;
+  }
+
+  @Override
+  public boolean isAllowedForTargetType(Object source, String action, Class<?> targetType) {
```
isemaya-square, line 13: I'm probably missing a lot of context, but are we not use `action` at all?

mmontgomery-square: Oh man! I can't believe I missed that. Yes, it should totally include action. Derp derp derp

PR #1119, opened 2022-08-08
Wiring up allow-clients-to-create-secrets permission check

mmontgomery-square:


server/src/main/java/keywhiz/service/permissions/PermissionCheckModule.java
```diff
@@ -8,13 +8,31 @@
...
+      OwnershipPermissionCheck ownershipCheck,
+      AllowClientSecretCreationPermissionCheck clientSecretCreationCheck) {
+
+    List<PermissionCheck> permissionChecks = List.of(
+        ownershipCheck,
+        automationClientCheck,
+        clientSecretCreationCheck
+    );
+
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(metricRegistry, permissionChecks);
```
mmontgomery-square, line 27: I just realized that the any permission check is short-circuiting, so we'll never be able to see the client secret creation check in action in the logs since, in the current configuration, the automation client check will always pass. I'll re-order these.

server/src/main/java/keywhiz/service/permissions/PermissionCheckModule.java
```diff
@@ -8,13 +8,31 @@
 public class PermissionCheckModule extends AbstractModule {
 
   @Override
-  protected void configure() {}
+  protected void configure() {
+    bindToDefaultConstructor(AllowClientSecretCreationPermissionCheck.class);
+  }
```
mmontgomery-square, line 13: Elsewhere in the code we have a "strict Guice module" which disables automatically injecting classes with a no-argument constructor, which is a slightly error-prone default behavior of Guice (the bugs this can introduce are rare but very confusing). This includes synthetic (i.e. generated by the compiler) constructors, so even though our "allow clients to create secrets" permission check doesn't appear to have a constructor at all, it still gets caught up.

So, rather than create an empty constructor just to slap @ Inject on it, I decided to write a little utility method and bind the default constructor directly instead, which also satisfies the strict requirement.

server/src/test/java/keywhiz/service/permissions/PermissionCheckModuleTest.java
```diff
@@ -21,4 +24,27 @@ class Holder {
     createInjector().injectMembers(holder);
     assertNotNull(holder.permissionCheck);
   }
+
+  @Test
+  public void injectedPermissionCheckAllowsClientSecretCreation() {
+    PermissionCheck permissionCheck = createInjector().getInstance(PermissionCheck.class);
+    assertTrue(permissionCheck.isAllowedForTargetType(newClient(), Action.CREATE, Secret.class));
+  }
```
mmontgomery-square, line 32: This test failed before the new permission check was added to the permission check list.

server/src/main/java/keywhiz/service/permissions/PermissionCheckModule.java
violetd12: So when AllowClientSecretCreationPermissionCheck is used in for PermissionCheck provider method, it's created following this constructor?

mmontgomery-square: Yup! We're just telling Guice that when it needs an instance of this class, it should use this constructor.

violetd12 (approved):

server/src/main/java/keywhiz/service/permissions/PermissionCheckModule.java
```diff
@@ -8,13 +8,31 @@
...
+        ownershipCheck,
+        clientSecretCreationCheck,
+        automationClientCheck
+    );
+
+    PermissionCheck anyPermissionCheck = new AnyPermissionCheck(metricRegistry, permissionChecks);
     return anyPermissionCheck;
   }
+
+  private <T> void bindToDefaultConstructor(Class<T> clazz) {
```
jdtw, line 31: Heh. I like `clazz`. Is that idiomatic? (An old coworker always used to name his buffers `buffy` and I always got a kick out of it.)

PR #1120, opened 2022-08-10
Adding logging and metrics to client secret creation permission check

mmontgomery-square:


violetd12 (approved):

server/src/main/java/keywhiz/service/permissions/PermissionCheckModule.java
```diff
@@ -6,12 +6,6 @@
 import java.util.List;
 
 public class PermissionCheckModule extends AbstractModule {
-
-  @Override
-  protected void configure() {
-    bindToDefaultConstructor(AllowClientSecretCreationPermissionCheck.class);
-  }
```
mmontgomery-square, line 13: This was needed because the permission check didn't have any explicit constructor defined. Now that it does, since the constructor is annotated with @ Inject it's just wired up automatically.

PR #1124, opened 2022-08-18
add logging to AutomationV2 API

violetd12:
[CISM-4227. ](https://jira.sqprod.co/browse/CISM-4227)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -405,6 +405,9 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.10.1</version>
         <configuration>
+          <compilerArgs>
+            <arg>-parameters</arg>
+          </compilerArgs>
           <source>${java.version}</source>
           <target>${java.version}</target>
         </configuration>
@@ -505,6 +508,9 @@
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.10.1</version>
           <configuration>
+            <compilerArgs>
+              <arg>-parameters</arg>
+            </compilerArgs>
             <source>${java.version}</source>
             <target>${java.version}</target>
           </configuration>
--- a/server/src/main/java/keywhiz/service/resources/automation/v2/GroupResource.java
+++ b/server/src/main/java/keywhiz/service/resources/automation/v2/GroupResource.java
@@ -34,6 +34,7 @@ import keywhiz.api.model.SecretSeries;
 import keywhiz.log.AuditLog;
 import keywhiz.log.Event;
 import keywhiz.log.EventTag;
+import keywhiz.log.LogArguments;
 import keywhiz.service.daos.AclDAO;
 import keywhiz.service.daos.AclDAO.AclDAOFactory;
 import keywhiz.service.daos.GroupDAO;
@@ -84,6 +85,7 @@ public class GroupResource {
   @Timed @ExceptionMetered
   @POST
   @Consumes(APPLICATION_JSON)
+  @LogArguments
   public Response createGroup(@Auth AutomationClient automationClient,
       @Valid CreateGroupRequestV2 request) {
     permissionCheck.checkAllowedForTargetTypeOrThrow(automationClient, Action.CREATE, Group.class);
@@ -124,6 +126,7 @@ public class GroupResource {
   @Timed @ExceptionMetered
   @GET
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public Iterable<String> groupListing(@Auth AutomationClient automationClient) {
     permissionCheck.checkAllowedForTargetTypeOrThrow(automationClient, Action.READ, Group.class);
 
@@ -144,6 +147,7 @@ public class GroupResource {
   @GET
   @Path("{name}")
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public GroupDetailResponseV2 groupInfo(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadOnly.getGroup(name)
@@ -177,6 +181,7 @@ public class GroupResource {
   @GET
   @Path("{name}/secrets")
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public Set<SanitizedSecret> secretsForGroup(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadOnly.getGroup(name)
@@ -199,6 +204,7 @@ public class GroupResource {
   @GET
   @Path("{name}/secretsandgroups")
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public Set<SanitizedSecretWithGroups> secretsWithGroupsForGroup(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadOnly.getGroup(name)
@@ -232,6 +238,7 @@ public class GroupResource {
   @GET
   @Path("{name}/ownedsecretsandgroups")
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public Set<SanitizedSecretWithGroups> ownedSecretsWithGroupsForGroup(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadOnly.getGroup(name)
@@ -264,6 +271,7 @@ public class GroupResource {
   @GET
   @Path("{name}/clients")
   @Produces(APPLICATION_JSON)
+  @LogArguments
   public Set<Client> clientDetailForGroup(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadOnly.getGroup(name)
@@ -284,6 +292,7 @@ public class GroupResource {
   @Timed @ExceptionMetered
   @DELETE
   @Path("{name}")
+  @LogArguments
   public Response deleteGroup(@Auth AutomationClient automationClient,
       @PathParam("name") String name) {
     Group group = groupDAOReadWrite.getGroup(name)
--- a/log/pom.xml
+++ b/log/pom.xml
@@ -14,5 +14,19 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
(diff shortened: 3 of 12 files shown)


server/src/main/java/keywhiz/service/resources/automation/v2/AutomationAPIsLoggingImpl.java
```diff
@@ -0,0 +1,15 @@
...
+import org.aopalliance.intercept.MethodInvocation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AutomationAPIsLoggingImpl implements MethodInterceptor {
+  private static final Logger logger = LoggerFactory.getLogger(AutomationAPIsLogging.class);
+  public Object invoke(MethodInvocation invocation) throws Throwable {
+    logger.info(String.format("Function Name: %s, Function Arguments: %s",invocation.getMethod().getName(), Arrays.toString(invocation.getArguments())));
+    return invocation.proceed();
+  }
```
mmontgomery-square, line 14: I love how simple this is! I still think it would be worthwhile to try to use reflection to get the parameter names automatically, either from the compiled code or from the annotations. For the compiled code, It turns out that the names may actually be present, you can use invocation.getMethod().getParameters() to get at the parameters array and then use the methods isNamePresent() and getName() to grab the name. If it is present I would swap out the value in the arguments array with a Map.Entry containing the name and argument value.

server/src/main/java/keywhiz/service/resources/automation/v2/GroupResource.java
```diff
@@ -84,6 +84,7 @@ public class GroupResource {
   @Timed @ExceptionMetered
   @POST
   @Consumes(APPLICATION_JSON)
+  @AutomationAPIsLogging()
   public Response createGroup(@Auth AutomationClient automationClient,
```
mmontgomery-square, line 88: I think you can leave off the parentheses if your annotation doesn't require any values.

server/src/main/java/keywhiz/service/resources/automation/v2/AutomationAPIsLoggingImpl.java
mmontgomery-square: Ah, forgot one thing: if you use Arrays.deepToString() it will correctly handle nested arrays. I don't know if we actually have any methods with arrays in their signature but better to get the proper behavior in place now before we need it.

server/src/main/java/keywhiz/service/resources/automation/v2/LogArgumentsImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import org.slf4j.LoggerFactory;
+
+public class LogArgumentsImpl implements MethodInterceptor {
+  private static final Logger logger = LoggerFactory.getLogger(LogArguments.class);
+  public Object invoke(MethodInvocation invocation) throws Throwable {
+    Parameter[] parameters = invocation.getMethod().getParameters();
+    Object[] arguments = invocation.getArguments();
+    Stream<Map.Entry<Parameter, Object>> parameterArgumentPairs = Streams.zip(Arrays.stream(parameters), Arrays.stream(arguments), Map::entry);
+    logger.info(String.format("Function Name: %s, Function Arguments: %s",invocation.getMethod().getName(),
+        Arrays.deepToString(parameterArgumentPairs.toArray())));
```
violetd12, line 20: Can potentially group together L16-L18 and L20

pom.xml
```diff
@@ -405,6 +405,9 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.10.1</version>
         <configuration>
+          <compilerArgs>
+            <arg>-parameters</arg>
```
violetd12, line 409: `-parameters` option will include the parameter names when calling `getParameters()`

server/src/main/java/keywhiz/service/resources/automation/v2/LogArgumentsImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.stream.Stream;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LogArgumentsImpl implements MethodInterceptor {
+  private static final Logger logger = LoggerFactory.getLogger(LogArguments.class);
+  public Object invoke(MethodInvocation invocation) throws Throwable {
+    Object[] parameters = Arrays.stream(invocation.getMethod().getParameters()).map(parameter -> parameter.isNamePresent()? parameter.getName() : parameter.getType()).toArray();
```
violetd12, line 16: Alternatively, I can just do `getParameters()`, which includes the type of the parameter as well. See below for an example log.

`keywhiz.service.resources.automation.v2.LogArguments: Function Name: createGroup, Function Arguments: [keywhiz.api.model.AutomationClient automationClient=AutomationClient{name=client}, keywhiz.api.automation.v2.CreateGroupRequestV2 request=CreateGroupRequestV2{name=group6, description=, metadata={}}]`

server/src/main/java/keywhiz/service/resources/automation/v2/LogArgumentsImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.stream.Stream;
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LogArgumentsImpl implements MethodInterceptor {
+  private static final Logger logger = LoggerFactory.getLogger(LogArguments.class);
+  public Object invoke(MethodInvocation invocation) throws Throwable {
+    Object[] parameters = Arrays.stream(invocation.getMethod().getParameters()).map(parameter -> parameter.isNamePresent()? parameter.getName() : parameter).toArray();
```
violetd12, line 16: Alternatively, I can just do `getParameters()`, which includes the type of the parameter as well. See below for an example log.

`keywhiz.service.resources.automation.v2.LogArguments: Function Name: createGroup, Function Arguments: [keywhiz.api.model.AutomationClient automationClient=AutomationClient{name=client}, keywhiz.api.automation.v2.CreateGroupRequestV2 request=CreateGroupRequestV2{name=group6, description=, metadata={}}]`

pom.xml
mmontgomery-square: Nice! I was surprised that this information was available at all, and also didn't realize that a special compiler flag was necessary to gather it.

server/src/main/java/keywhiz/service/resources/automation/v2/LogArgumentsImpl.java
mmontgomery-square: If I'm reading this correctly, the map operation on line 16 will produce either the parameter name (type String) or the parameter object itself (type Parameter), and the toString() on Parameter includes the modifiers, type, name, and some other info according to the docs:

https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Parameter.html#toString--

If so, I think that's going to really blow up the log statement in the case where we don't have parameters available (which shouldn't ever really be the case given we control the compiler settings as well, but...).

I was thinking more of an iterative approach, like:

```
for (int i = 0; i < copyOfArguments.length; i++) {
  if (parameters[i].isNamePresent()) {
    copyOfArguments[i] = Map.entry(parameters[i].getName(), copyOfArguments[i]);
  }
}
```

I know we discussed trying to grab argument names from the various annotations on the parameters as a possible backup, but if you were planning on adding that in a future revision I think we could skip that for now since this approach is pretty awesome by itself. Fancy stuff!

server/src/main/java/keywhiz/service/resources/automation/v2/LogArgumentsModule.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.google.inject.AbstractModule;
+import com.google.inject.matcher.Matchers;
+
+public class LogArgumentsModule extends AbstractModule {
+  @Override
+  protected void configure() {
+    bindInterceptor(Matchers.any(), Matchers.annotatedWith(LogArguments.class), new LogArgumentsImpl());
+  }
+}
```
mmontgomery-square, line 11: Seeing code like this that's so powerful yet so concise makes me happy. Yay Guice!

log/src/main/java/keywhiz/log/LogArgumentsImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory;
+
+public class LogArgumentsImpl implements MethodInterceptor {
+  private static final Logger logger = LoggerFactory.getLogger(LogArguments.class);
+
+  public Object invoke(MethodInvocation invocation) throws Throwable {
+    Parameter[] parameters = invocation.getMethod().getParameters();
+    Object[] arguments = invocation.getArguments();
+
+    Object[] parameterArgumentPairs = new Object[arguments.length];
```
mmontgomery-square, line 18: I spent a little bit of time trying to figure out whether the `this` reference is considered to be the first argument in an instance method (it's not) and ended up learning about something called a receiver parameter, which I did not know was a thing:

https://riptutorial.com/java/example/23136/annotations-for--this--and-receiver-parameters

log/src/main/java/keywhiz/log/LogArgumentsImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+    for (int i = 0; i < arguments.length; i++) {
+      if (parameters[i].isNamePresent()) {
+        parameterArgumentPairs[i] = new AbstractMap.SimpleImmutableEntry<>(parameters[i].getName(), arguments[i]);
+      } else {
+        parameterArgumentPairs[i] = arguments[i];
+      }
+    }
+
+    logger.info(String.format("Function Name: %s, Function Arguments: %s",
```
mmontgomery-square, line 28: I would change Function to Method since it is very technically slightly more correct.

mmontgomery-square (approved):

PR #1154, opened 2022-10-28
Adding Client and Group owner fields to model

mmontgomery-square:


api/src/main/java/keywhiz/api/model/Group.java
```diff
@@ -71,6 +79,27 @@ public Group(@JsonProperty("id") long id,
     this.updatedAt = updatedAt;
     this.updatedBy = nullToEmpty(updatedBy);
     this.metadata = ImmutableMap.copyOf(metadata == null ? ImmutableMap.of() : metadata);
+    this.owner = owner;
+  }
+
+  public Group(long id,
```
graysonchao, line 85: does this need Json annotations too?

graysonchao (approved):
gosh, that really does touch a lot of files.

api/src/main/java/keywhiz/api/model/Group.java
mmontgomery-square: Interestingly, it had them originally, which caused Jackson to fail.

Jackson expects there to be a single constructor or factory method annotated with @JsonCreator, but if you leave that annotation off (which we were doing previously) then it scans the class for a constructor whose arguments have JSON-ish annotations. If you have multiple such constructors, it complains that it doesn't know which one to use and fails.

So, since this constructor is only intended for backwards compatibility, it doesn't have any JSON annotations because Jackson will never invoke it. I've also explicitly marked the other constructor with @JsonCreator so that we're a bit more well behaved.

mmontgomery-square:
> gosh, that really does touch a lot of files.

Yeah, it's unfortunate. One reason is because the old/non-v2 API uses constructors rather than builders, so every new field is a new constructor argument is a backwards-incompatible change. Also, these classes look like they've been more or less untouched since they were created, so there was some one-time overhead to add testing for the "original" definition so that we can have some assurance that we're not breaking things at the serialization level.

mmontgomery-square:
(forced merge since graysonchao's invite is still pending)

PR #1161, opened 2022-12-01
Adding owner support to GroupDAO

mmontgomery-square:


api/src/main/java/keywhiz/api/model/Group.java
```diff
@@ -59,7 +59,7 @@ public class Group {
   private final ImmutableMap<String, String> metadata;
 
   @JsonProperty
-  private final String owner;
+  private String owner;
```
mmontgomery-square, line 62: This was an oversight in the original model change: being able to mutate this property makes the DAO code much cleaner in that you don't have to recreate the same Group twice, once without an owner and then once with.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -49,7 +55,25 @@ private GroupDAO(DSLContext dslContext, GroupMapper groupMapper, ObjectMapper ma
...
+      String creator,
+      String description,
+      ImmutableMap<String, String> metadata) {
+    return createGroup(
+        name,
+        creator,
+        description,
+        metadata,
+        NO_OWNER);
+  }
```
mmontgomery-square, line 69: As much as I prefer to avoid the proliferation of overloads for backwards compatibility's sake, this is just temporary until we wire up the corresponding Resources (which sit one level higher).

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -93,22 +118,57 @@ public void deleteGroup(Group group) {
           .set(SECRETS.OWNER, (Long) null)
           .where(SECRETS.OWNER.eq(group.getId()))
           .execute();
+      DSL.using(configuration)
+          .update(GROUPS)
+          .set(GROUPS.OWNER, (Long) null)
+          .where(GROUPS.OWNER.eq(group.getId()))
+          .execute();
     });
```
mmontgomery-square, line 126: In the absence of enforced foreign keys, we're forced to manually clean up dangling references to a deleted group.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -93,22 +118,57 @@ public void deleteGroup(Group group) {
...
+    for (GroupsRecord record : records) {
+      if (record.getOwner() != null) {
+        Group group = groupsById.get(record.getId());
+        Group owner = groupsById.get(record.getOwner());
+        group.setOwner(owner.getName());
+      }
+    }
+
+    return ImmutableSet.copyOf(groupsById.values());
+  }
```
mmontgomery-square, line 155: This method returns all the groups in the database. I don't know why you'd want to do that (or why we'd allow it, especially without pagination), but as a minor optimization we're taking advantage of the fact that all groups are now in memory to populate the owner fields without having to go back to the database.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -93,22 +118,57 @@ public void deleteGroup(Group group) {
...
+
+    if (record.getOwner() != null) {
+      GroupsRecord ownerRecord = dslContext.fetchOne(GROUPS, GROUPS.ID.eq(record.getOwner()));
+      if (ownerRecord != null) {
+        group.setOwner(ownerRecord.getName());
+      }
+    }
+
+    return group;
   }
```
mmontgomery-square, line 172: I believe this is a lot cleaner than the prior change for secrets, which is probably worth revisiting.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -93,22 +118,57 @@ public void deleteGroup(Group group) {
...
+          .set(GROUPS.OWNER, (Long) null)
+          .where(GROUPS.OWNER.eq(group.getId()))
+          .execute();
     });
   }
 
   public Optional<Group> getGroup(String name) {
     GroupsRecord r = dslContext.fetchOne(GROUPS, GROUPS.NAME.eq(name));
-    return Optional.ofNullable(r).map(groupMapper::map);
+    return Optional.ofNullable(toGroup(r));
```
spennymac, line 131: nit: Is there a way to get the group and owner group in one query?

spennymac (approved):
Can you just add a  little description to the PR?

server/src/main/java/keywhiz/service/daos/GroupDAO.java
mmontgomery-square: After poking around at this for a couple hours, I believe I have misjudged Jooq as a result of an earlier programming error (it only falls back to string matching when proper aliasing is missing). It turns out that it's trivial but non-obvious to correctly perform a self-join. The new code is much cleaner and also efficient in that we're only fetching the name of the owner since that's the only field we care about. It also made me realize two things:

1. There are a ton of places in Keywhiz were we hydrate entire objects only to throw away every field but one (typically, the secret name). We could do so much better there.
2. I need to go back and re-implement the owner fetching for secrets.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -93,22 +120,70 @@ public void deleteGroup(Group group) {
...
+        .select(GROUPS.fields())
+        .select(GROUP_OWNERS.NAME)
+        .from(GROUPS)
+        .leftJoin(GROUP_OWNERS)
+        .on(GROUPS.OWNER.eq(GROUP_OWNERS.ID))
+        .where(GROUPS.NAME.eq(name))
+        .fetchOne();
+
+    return Optional.ofNullable(recordToGroup(record));
   }
```
mmontgomery-square, line 142: We can't rely on the convenience of a dumb mapper against a single table select, but I also don't think the new query is that much more complicated.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
```diff
@@ -39,6 +43,10 @@
 import static keywhiz.jooq.tables.Secrets.SECRETS;
 
 public class GroupDAO {
+  private static final Long NO_OWNER = null;
+
+  private static final Groups GROUP_OWNERS = GROUPS.as("owners");
```
mmontgomery-square, line 48: Aliasing the owners table was the key insight into getting the self-join to work as intended.

Note that introducing an alias for GROUPS simplifies the generated SQL quite a bit (without an alias the fields are all prefixed by the schema and the table name), and we also have the option of invoking .asterisk() rather than .fields() to emit a literal asterisk into the query for all that table's fields. However, this only saves us a few bytes of query text and has some negative consequences as far as Jooq's type safety, I believe.

server/src/main/java/keywhiz/service/daos/JooqSqlPrinter.java
```diff
@@ -0,0 +1,27 @@
...
+
+    if (executeContext.query() != null) {
+      System.out.println(dslContext.renderInlined(executeContext.query()));
+    }
+  }
+
+  public static void install(DSLContext dslContext) {
+    dslContext.configuration().set(new DefaultExecuteListenerProvider(new JooqSqlPrinter()));
+  }
+}
```
mmontgomery-square, line 27: Inspired by a code sample in Jooq's documentation, this allowed me to spit the raw SQL out to the console in order to see what the consequences of various changes/approaches were.

server/src/main/java/keywhiz/service/daos/GroupDAO.java
spennymac: Awesome @mmontgomery-square! Thanks for looking into this.

PR #1162, opened 2022-12-03
Adding API for batch secret creation/updates

mmontgomery-square:


testing/src/main/java/keywhiz/testing/JsonHelpers.java
```diff
@@ -56,8 +57,12 @@ private JsonHelpers() { /* singleton */ }
...
    */
-  public static String asJson(Object object) throws JsonProcessingException {
-    return MAPPER.writeValueAsString(object);
+  public static String asJson(Object object) {
+    try {
+      return MAPPER.writeValueAsString(object);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
   }
```
mmontgomery-square, line 66: Personally I feel that checked exceptions in test helper methods just contribute to API cruft, IDE nagging (wrap with try/catch, etc.), and general obnoxiousness.

mmontgomery-square:
The intent for codeowners was an either/or for the teams during the handoff transition, not both, so I've overridden the merge requirements.

PR #1163, opened 2022-12-04
Adding support for hard-deleting secrets

mmontgomery-square:


api/src/main/java/keywhiz/api/model/SecretDeletionMode.java
```diff
@@ -0,0 +1,17 @@
...
+
+  public static final String HARD = "hard";
+  public static final String SOFT = "soft";
+
+  private static final List<String> values = ImmutableList.of(HARD, SOFT);
+
+  public static List<String> values() {
+    return values;
+  }
+}
```
mmontgomery-square, line 17: This class is intended to be used by clients. Generally speaking, Java enums don't play well with API evolution (unknown values result in exceptions), so we prefer strings. However, it's still nice to have a constants class available.

api/src/main/java/keywhiz/api/model/SecretSeries.java
```diff
@@ -34,18 +36,18 @@
...
-      ApiDate createdAt,
-      @Nullable String createdBy,
-      ApiDate updatedAt,
-      @Nullable String updatedBy,
-      @Nullable String type,
-      @Nullable Map<String, String> generationOptions,
-      @Nullable Long currentVersion) {
+  @JsonCreator public static SecretSeries of(
+      @JsonProperty("id") long id,
+      @JsonProperty("name") String name,
```
mmontgomery-square, line 41: Unlike many other API classes, this class was apparently never directly returned as a result so it had none of the Jackson annotations required for serialization. Now that we have a new API which returns deleted secret series (if only to support testing), I've added annotations and tests to verify that (de)serialization works as intended.

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -559,10 +558,24 @@ public void setCurrentSecretVersionByName(String name, long versionId, String up
    * @param name of secret series to delete.
    */
   public void deleteSecretsByName(String name) {
+    deleteSecretsByName(name, SecretDeletionMode.SOFT);
+  }
+
+  public void deleteSecretsByName(String name, SecretDeletionMode mode) {
```
mmontgomery-square, line 564: Rather than touch a zillion files with a refactor, I just left the old signature intact as an overload.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -321,59 +321,99 @@ public ImmutableList<SecretSeries> getSecretSeriesBatched(int idx, int num, bool
...
+
+      SecretsRecord record = dslContext.select()
+          .from(SECRETS)
+          .where(SECRETS.ID.eq(id))
+          .forUpdate()
+          .fetchOneInto(SECRETS);
+
+      hardDeleteSecretSeries(dslContext, record);
+    });
+  }
```
mmontgomery-square, line 350: In the existing code there was a lot of duplication between the by-id and by-name methods, even though only the lookup logic was (slightly) different and the deletion logic was identical. I went ahead and split out the deletion logic into a dedicated helper method to reduce duplication in the new methods and also applied this same pattern to the old methods. Net result: much cleaner code.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -472,9 +512,9 @@ public static class SecretSeriesDAOFactory implements DAOFactory<SecretSeriesDAO
 
   // create a new name for the deleted secret, so that deleted secret names can be reused, while
   // still having a unique constraint on the name field in the DB
-  private String transformNameForDeletion(String name) {
+  private static String transformNameForDeletion(String name) {
     long now = OffsetDateTime.now().toEpochSecond();
-    return String.format(".%s.deleted.%d.%s", name, now, UUID.randomUUID().toString());
+    return String.format(".%s.deleted.%d.%s", name, now, UUID.randomUUID());
   }
```
mmontgomery-square, line 518: Fixing a couple IDE nags:

1. This is a utility method which doesn't rely on any state from the hosting object, so I've marked it static.
2. String.format() already coerces every argument into a String by invoking toString(), so manually invoking it ourselves isn't necessary.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -815,6 +817,17 @@ public Iterable<String> modifySecretGroups(@Auth AutomationClient automationClie
...
+  @Timed
+  @ExceptionMetered
+  @GET
+  @Path("/deleted/{name}")
+  @LogArguments
+  public List<SecretSeries> getDeletedSecretSeriesByName(
+    @Auth AutomationClient automationClient,
+    @PathParam("name") String name) {
+    return secretDAO.getSecretsWithDeletedName(name);
+  }
```
mmontgomery-square, line 829: Easy peasy. This functionality was already available but only exposed to the CLI, so exposing it via a REST endpoint was trivial.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -827,15 +840,22 @@ public Iterable<String> modifySecretGroups(@Auth AutomationClient automationClie
...
-  public Response deleteSecretSeries(@Auth AutomationClient automationClient,
-      @PathParam("name") String name) {
+  public Response deleteSecretSeries(
+      @Auth AutomationClient automationClient,
+      @PathParam("name") String name,
+      @QueryParam("deletionMode") String deletionMode) {
+
+    SecretDeletionMode mode = deletionMode == null
+        ? SecretDeletionMode.SOFT
+        : EnumUtils.getEnumIgnoreCase(SecretDeletionMode.class, deletionMode);
```
mmontgomery-square, line 850: In the client-facing API equivalent of this enum I used lowercase constants (since I imagined them being dumped directly into URLs as query parameters) but I followed the Java naming standard of all-caps for the actual Java enum. Since enum lookups are case-sensitive (gross), this Commons utility method prevents any case-related mixups.

server/src/main/java/keywhiz/service/resources/automation/v2/SecretResource.java
```diff
@@ -827,15 +840,22 @@ public Iterable<String> modifySecretGroups(@Auth AutomationClient automationClie
...
     permissionCheck.checkAllowedOrThrow(automationClient, Action.DELETE, secret);
 
     // Get the groups for this secret so they can be restored manually if necessary
     Set<String> groups = aclDAO.getGroupsFor(secret).stream().map(Group::getName).collect(toSet());
 
-    secretDAO.deleteSecretsByName(name);
+    secretDAO.deleteSecretsByName(name, mode);
 
     // Record the deletion in the audit log
     Map<String, String> extraInfo = new HashMap<>();
```
mmontgomery-square, line 861: It occurs to me that we should add the mode to the audit log entry.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTestHelper.java
```diff
@@ -35,10 +39,10 @@ public class SecretResourceTestHelper {
...
   }
 
-  Response create(CreateSecretRequestV2 request) throws IOException {
-    RequestBody body = RequestBody.create(JSON, mapper.writeValueAsString(request));
+  Response create(CreateSecretRequestV2 request) {
+    RequestBody body = RequestBody.create(JSON, toJson(request));
     Request post = clientRequest("/automation/v2/secrets").post(body).build();
-    return mutualSslClient.newCall(post).execute();
+    return execute(post);
   }
```
mmontgomery-square, line 46: Constantly being nagged about exception handling while writing tests drives me nuts, so I modified this and several other test helper methods in this class to use unchecked exceptions.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTestHelper.java
```diff
@@ -239,8 +260,49 @@ List<String> modifyGroups(String name, ModifyGroupsRequestV2 request) throws IOE
     });
   }
 
-  Response deleteSeries(String name) throws IOException {
+  Response deleteSeries(String name, String secretDeletionMode) {
+    String uri = UriBuilder.fromPath("/automation/v2/secrets")
+        .path(name)
+        .queryParam("deletionMode", secretDeletionMode)
+        .toString();
```
mmontgomery-square, line 267: We hand-assemble URIs using String.format() without proper escaping, which is sloppy. This is intended to provide a better example going forward.

server/src/main/java/keywhiz/service/daos/SecretDeletionMode.java
```diff
@@ -0,0 +1,6 @@
+package keywhiz.service.daos;
+
+public enum SecretDeletionMode {
```
spennymac, line 3: WDYT about a fromString, or something similar, static method that that defers to EnumUtils?

spennymac (changes requested):
LGTM, I just had one concern with EnumUtils.getEnumIgnoreCase returning a  null value if a client hits the rest endpoint with an unknown value. 

Note: I didn't invest too much time, but I was not able to run the tests locally to verify if this would cause a problem or not.

mmontgomery-square:
> LGTM, I just had one concern with EnumUtils.getEnumIgnoreCase returning a null value if a client hits the rest endpoint with an unknown value.
> 
> Note: I didn't invest too much time, but I was not able to run the tests locally to verify if this would cause a problem or not.

In the past I've always had a rule that every enum gets an UNKNOWN value whose job is to serve as a stand-in value for unrecognized enum values, or to serve as a default in the case where you're introducing, say, a new enum column to a table and there really isn't a sensible default from the meaningful values. But, might be overkill here. I didn't realize that the enum helper method returned null for unrecognized values, I think we'd probably want to throw an exception in that case. I'll see about a parse or similar method that encapsulates all this.

spennymac (approved):

PR #1167, opened 2022-12-05
Adding owner support to ClientDAO

mmontgomery-square:


api/src/main/java/keywhiz/api/model/Client.java
```diff
@@ -72,7 +72,7 @@ public class Client {
   private final boolean enabled;
 
   @JsonProperty
-  private final String owner;
+  private String owner;
```
mmontgomery-square, line 75: Same as the last PR: making the owner field writable simplifies the DAO layer code.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -138,26 +166,60 @@ public void sawClient(Client client, @Nullable Principal principal) {
...
+  }
+
+  private Optional<Client> getClient(Condition condition) {
+    Record record = dslContext
+        .select(CLIENTS.fields())
+        .select(CLIENT_OWNERS.NAME)
+        .from(CLIENTS)
+        .leftJoin(CLIENT_OWNERS)
+        .on(CLIENTS.OWNER.eq(CLIENT_OWNERS.ID))
+        .where(condition)
```
mmontgomery-square, line 187: I got a little fancier than the GroupDAO PR by parameterizing the where clause.

server/src/test/java/keywhiz/service/daos/GroupDAOTest.java
```diff
@@ -147,6 +153,23 @@ public class GroupDAOTest {
...
+
+    Client beforeWithOwner = getClientById(clientId);
+    assertEquals(ownerName, beforeWithOwner.getOwner());
+
+    Group owner = getGroupById(ownerId);
+    groupDAO.deleteGroup(owner);
+
+    Client afterWithoutOwner = getClientById(clientId);
+    assertNull(afterWithoutOwner.getOwner());
+  }
```
mmontgomery-square, line 171: It turns out that this test is unnecessary. Because we do a left join on the owning group and then only populate the owner name if the owner is present, a "dangling" owner ID field doesn't cause a problem. Of course, that's bad, because it masks an underlying issue. I think the solution is to add a sanity check in the record conversion code that will barf if the owner ID is present but the owning group is not.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -138,26 +166,69 @@ public void sawClient(Client client, @Nullable Principal principal) {
...
+    GroupsRecord ownerRecord = record.into(CLIENT_OWNERS);
+
+    boolean danglingOwner = clientRecord.getOwner() != null && ownerRecord.getId() == null;
+    if (danglingOwner) {
+      throw new IllegalStateException(
+          String.format(
+              "Owner %s for client %s is missing.",
+              clientRecord.getOwner(),
+              clientRecord.getName()));
+    }
```
mmontgomery-square, line 208: This logic fixes a bug I'd discussed in another comment: the left join will present a null owner name if the owner is missing regardless of whether the owner ID in the client is populated or not. Because we don't enforce referential integrity in the database, we have to hand-manage it here.

server/src/main/java/keywhiz/service/daos/ClientDAO.java
```diff
@@ -138,26 +166,69 @@ public void sawClient(Client client, @Nullable Principal principal) {
...
+        .from(CLIENTS)
+        .leftJoin(CLIENT_OWNERS)
+        .on(CLIENTS.OWNER.eq(CLIENT_OWNERS.ID))
+        .fetch();
+
+    Set<Client> clients = results.stream()
+        .map(this::recordToClient)
+        .collect(Collectors.toSet());
+
+    return ImmutableSet.copyOf(clients);
```
spennymac, line 231: You can map off of results instead if you want.

```suggestion
    List<Client> clients = dslContext
        .select(CLIENTS.fields())
        .select(CLIENT_OWNERS.NAME)
        .from(CLIENTS)
        .leftJoin(CLIENT_OWNERS)
        .on(CLIENTS.OWNER.eq(CLIENT_OWNERS.ID))
        .fetch()
        .map(this::recordToClient);

    return ImmutableSet.copyOf(clients);
```

spennymac (approved):
LGTM. I left a few suggestions.

PR #1168, opened 2022-12-08
Reducing expired secrets fetch cost by 25%

mmontgomery-square:


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -263,7 +264,7 @@ public ImmutableList<SecretSeries> getSecretSeries(@Nullable Long expireMaxTime,
     }
 
     SelectQuery<Record> select = dslContext
-          .select()
+          .select(SECRETS.fields())
           .from(SECRETS)
           .join(secretsContentTable)
           .on(SECRETS.CURRENT.equal(SECRETS_CONTENT.ID))
```
mmontgomery-square, line 270: When you issue a select() without any qualification, you pull in all the rows from all joined tables. However, if you look at the code just below the diff, you can see that we throw all the extra columns away and just map to a SecretSeries, which consists solely of columns from the SECRETS table.

Diffing a before/after query plan for our long-running expired secrets query showed a 25% reduction in query cost.

PR #1169, opened 2022-12-10
Populating secret ownership via join rather than subquery

mmontgomery-square:


spennymac (approved):

mmontgomery-square:
Update: I intentionally did not merge this because, while I believe it to be safe, it is also invasive, and I prioritized other more critical performance fixes over this one. It should still be merged eventually as I believe it to be a significant performance improvement.

PR #1170, opened 2022-12-12
Adding option to disable database validation

mmontgomery-square:


server/src/test/java/keywhiz/KeywhizServiceTest.java
```diff
@@ -7,8 +7,24 @@
...
+  public void doesNotValidateDatabaseIfDatabaseValidationDisabledInConfig() {
+    KeywhizConfig config = mock(KeywhizConfig.class);
+    when(config.shouldValidateDatabase()).thenReturn(false);
+
+    ServiceContext context = ServiceContext.create();
+    KeywhizService service = context.getService();
+    service.validateDatabase(config);
+
+    verify(config, never()).getDataSourceFactory();
+  }
```
mmontgomery-square, line 26: Ideally we'd have some sort of databaseValidator.validateDatabase() call that we could check for, but the way that the FlyWay stuff is instantiated doesn't lend itself well to that kind of check.

server/src/main/java/keywhiz/KeywhizConfig.java
```diff
@@ -246,6 +249,14 @@ public void setMaximumSecretSizeInBytesInclusive(Long maximumSecretSizeInBytesIn
     this.maximumSecretSizeInBytesInclusive = maximumSecretSizeInBytesInclusive;
   }
 
+  public boolean shouldValidateDatabase() {
+    return validateDatabase == null ? true : validateDatabase;
+  }
```
mmontgomery-square, line 254: Note that since we historically always performed database validation, we default to validation if the field is unspecified for backwards compatibility.

PR #1172, opened 2022-12-14
Adding expiry column and index to secrets table

mmontgomery-square:


spennymac (approved):

server/src/main/resources/db/mysql/migration/V17__secret_expiry.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE secrets ADD expiry BIGINT NULL;
```
rzio, line 1: I don't understand how the name correlates to the data type - what data will this column contain.

mmontgomery-square: "expiry" is the name of the corresponding column on the secrets content table, the name and type were chosen to match exactly. The question that Keywhiz is attempting to answer is, "What are all the non-deleted secrets which are expiring?" "non-deleted" and "expiring" live in two separate tables, which necessitates a join, and the join's performance is abysmal. Hoisting the expiration info up a level eliminates the need for the join.

PR #1174, opened 2022-12-19
Adding API to efficiently check whether a secret exists

mmontgomery-square:


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -78,6 +78,10 @@ private SecretSeriesDAO(
     this.rowHmacGenerator = rowHmacGenerator;
   }
 
+  public boolean secretSeriesExists(String name) {
+    return dslContext.fetchExists(SECRETS, SECRETS.NAME.eq(name));
+  }
```
mmontgomery-square, line 83: Fortunately, Jooq includes a handy fetchExists method which just returns true or false depending on whether the query returns any records or not.

PR #1177, opened 2022-12-21
Adding bookkeeping to keep secret series and secret content expiry in…

mmontgomery-square:
… sync

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
@@ -34,6 +34,7 @@ import org.jooq.DSLContext;
 import org.jooq.Field;
 import org.jooq.Record;
 import org.jooq.Record1;
+import org.jooq.Record2;
 import org.jooq.SelectQuery;
 import org.jooq.Table;
 import org.jooq.impl.DSL;
@@ -176,14 +177,36 @@ public class SecretSeriesDAO {
   }
 
   public int setExpiration(long secretContentId, Instant expiration) {
-    Field<Long> minExpiration = decode()
-        .when(SECRETS_CONTENT.EXPIRY.eq(0L), val(expiration.getEpochSecond()))
-        .otherwise(least(SECRETS_CONTENT.EXPIRY, val(expiration.getEpochSecond())));
+    return dslContext.transactionResult(configuration -> {
+      SecretsContentRecord content = dslContext.select(SECRETS_CONTENT.EXPIRY)
+          .from(SECRETS_CONTENT)
+          .where(SECRETS_CONTENT.ID.eq(secretContentId))
+          .forUpdate()
+          .fetchOneInto(SecretsContentRecord.class);
 
-    return dslContext.update(SECRETS_CONTENT)
-        .set(SECRETS_CONTENT.EXPIRY, minExpiration)
-        .where(SECRETS_CONTENT.ID.eq(secretContentId))
-        .execute();
+      if (content == null) {
+        return 0;
+      }
+
+      Long currentExpiry = content.getExpiry();
+      long epochSeconds = expiration.getEpochSecond();
+
+      Long updatedExpiry = (currentExpiry == null || currentExpiry == 0)
+          ? epochSeconds
+          : Math.min(currentExpiry, epochSeconds);
+
+      int contentsUpdated = dslContext.update(SECRETS_CONTENT)
+          .set(SECRETS_CONTENT.EXPIRY, updatedExpiry)
+          .where(SECRETS_CONTENT.ID.eq(secretContentId))
+          .execute();
+
+      int secretsUpdated = dslContext.update(SECRETS)
+          .set(SECRETS.EXPIRY, updatedExpiry)
+          .where(SECRETS.CURRENT.eq(secretContentId))
+          .execute();
+
+      return contentsUpdated + secretsUpdated;
+    });
   }
 
   public int setRowHmacByName(String secretName, String hmac) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -176,14 +170,36 @@ void updateSecretSeries(
   }
 
   public int setExpiration(long secretContentId, Instant expiration) {
-    Field<Long> minExpiration = decode()
-        .when(SECRETS_CONTENT.EXPIRY.eq(0L), val(expiration.getEpochSecond()))
-        .otherwise(least(SECRETS_CONTENT.EXPIRY, val(expiration.getEpochSecond())));
```
mmontgomery-square, line 181: decode() appears to be a method which allows you to insert conditional logic as a value in a Jooq query. As a consequence of the edits below (reading the record first rather than doing a blind update), this has been replaced with plain Java.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -176,14 +170,36 @@ void updateSecretSeries(
...
+
+      int contentsUpdated = dslContext.update(SECRETS_CONTENT)
+          .set(SECRETS_CONTENT.EXPIRY, updatedExpiry)
+          .where(SECRETS_CONTENT.ID.eq(secretContentId))
+          .execute();
+
+      int secretsUpdated = dslContext.update(SECRETS)
+          .set(SECRETS.EXPIRY, updatedExpiry)
+          .where(SECRETS.CURRENT.eq(secretContentId))
+          .execute();
```
mmontgomery-square, line 199: This is the block which keeps the series expiration and content expiration in sync. However, note that it only applies when the content being updated is the current version of a secret i.e. updating the expiration for non-current content has no effect on the series.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -201,18 +217,21 @@ public int setHmac(long secretContentId, String hmac) {
...
-    long checkId;
-    Record1<Long> r = dslContext.select(SECRETS_CONTENT.SECRETID)
+    SecretsContentRecord r = dslContext
+        .select(
+            SECRETS_CONTENT.SECRETID,
+            SECRETS_CONTENT.EXPIRY)
         .from(SECRETS_CONTENT)
         .where(SECRETS_CONTENT.ID.eq(secretContentId))
-        .fetchOne();
+        .fetchOneInto(SECRETS_CONTENT);
```
mmontgomery-square, line 226: Although it is possible to select arbitrary column values into a generic record, I think it's a lot more clear to always map the results into a properly typed record, even if not every field is populated.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -201,18 +217,21 @@ public int setHmac(long secretContentId, String hmac) {
...
+
     if (r == null) {
       throw new BadRequestException(
           String.format("The requested version %d is not a known version of this secret",
               secretContentId));
     }
 
-    checkId = r.value1();
+    long checkId = r.getSecretid();
     if (checkId != secretId) {
```
mmontgomery-square, line 235: Expanding on the above, I think getSecretId() is much more clear than value1().

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -221,18 +240,23 @@ public int setCurrentVersion(long secretId, long secretContentId, String updater
 
     return dslContext.update(SECRETS)
         .set(SECRETS.CURRENT, secretContentId)
+        .set(SECRETS.EXPIRY, r.getExpiry())
         .set(SECRETS.UPDATEDBY, updater)
         .set(SECRETS.UPDATEDAT, now)
         .where(SECRETS.ID.eq(secretId))
         .execute();
   }
```
mmontgomery-square, line 248: Because we're explicitly setting the current content version, we always update the series expiration to match.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -221,18 +240,23 @@ public int setCurrentVersion(long secretId, long secretContentId, String updater
...
-    SecretsRecord r =
-        dslContext.fetchOne(SECRETS, SECRETS.ID.eq(id).and(SECRETS.CURRENT.isNotNull()));
+    SecretsRecord r = getSecretSeriesRecordById(id);
     return Optional.ofNullable(r).map(secretSeriesMapper::map);
   }
 
+  @VisibleForTesting
+  SecretsRecord getSecretSeriesRecordById(long id) {
+    return dslContext.fetchOne(SECRETS, SECRETS.ID.eq(id).and(SECRETS.CURRENT.isNotNull()));
+  }
```
mmontgomery-square, line 258: So far I've chosen not to expose the series-level expiration in the Java model (SecretSeries), since it's more of an internal optimization for the database. As a result, in order to examine the series-level expiration we need access to the database model's record object, which we previously didn't have.

server/src/test/java/keywhiz/service/daos/SecretSeriesDAOTest.java
```diff
@@ -566,5 +649,28 @@ private long createSecretContent(long secretId) {
...
+        Instant.ofEpochSecond(expiration);
+      } catch (DateTimeException e) {
+        continue;
+      }
+
+      break;
+    }
+
+    return expiration;
+  }
```
mmontgomery-square, line 674: Because the Keywhiz APIs bounce back and forth between long and Instant, we have to ensure that the expiration we generate is valid for both (this came up during a test where the random long value was outside the bounds of what's allowed for Instant seconds).

spennymac (approved):

PR #1184, opened 2023-01-10
Switching expired secret series query to use new expiry column on sec…

mmontgomery-square:
…rets table

### DIFF
--- a/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
+++ b/server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
@@ -33,7 +33,6 @@ import org.jooq.Configuration;
 import org.jooq.DSLContext;
 import org.jooq.Record;
 import org.jooq.SelectQuery;
-import org.jooq.Table;
 import org.jooq.impl.DSL;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


cathy-yun-fu (approved):

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -282,38 +281,27 @@ public ImmutableList<SecretSeries> getSecretSeries(@Nullable Long expireMaxTime,
...
     if (expireMaxTime != null && expireMaxTime > 0) {
       // Set a lower bound of "now" on the expiration only if it isn't configured separately
       if (expireMinTime == null || expireMinTime == 0) {
         long now = System.currentTimeMillis() / 1000L;
-        select.addConditions(SECRETS_CONTENT.EXPIRY.greaterOrEqual(now));
+        select.addConditions(SECRETS.EXPIRY.greaterOrEqual(now));
       }
-      select.addConditions(SECRETS_CONTENT.EXPIRY.lessThan(expireMaxTime));
+      select.addConditions(SECRETS.EXPIRY.lessThan(expireMaxTime));
     }
```
mmontgomery-square, line 299: In essense, every reference to SECRETS_CONTENT.EXPIRY is now just SECRETS.EXPIRY (the value is the same), and we omit a no-longer-necessary join.

There is a very subtle difference in that, because we lack referential integrity (no foreign keys), it's possible that a secret that would have been excluded by the join (a "dangling" current secret content reference) would now be included. This should be a non-issue in practice.

PR #1188, opened 2023-01-19
Adding hard secret deletion support to CLI

mmontgomery-square:


api/src/main/java/keywhiz/api/model/SanitizedSecret.java
```diff
@@ -197,4 +199,30 @@ public static SanitizedSecret fromSecret(Secret secret) {
   public static String displayName(SanitizedSecret sanitizedSecret) {
     return sanitizedSecret.name();
   }
+
+  public static Builder builder() { return new AutoValue_SanitizedSecret.Builder();}
+
+  public abstract Builder toBuilder();
+
+  @AutoValue.Builder
+  public static abstract class Builder {
```
mmontgomery-square, line 208: I didn't end up using this functionality, but I've been adding builder support to the model classes one by one as I touch them because 15+ argument constructors are ridiculous.

cli/src/main/java/keywhiz/cli/commands/DeleteAction.java
```diff
@@ -102,7 +105,11 @@ public DeleteAction(DeleteActionConfig deleteActionConfig, KeywhizClient client)
               return;
             } else if (line.toUpperCase().startsWith("Y")) {
               logger.info("Deleting secret '{}'.", sanitizedSecret.name());
-              keywhizClient.deleteSecretWithId(sanitizedSecret.id());
+              if (deleteActionConfig.mode == null) {
+                keywhizClient.deleteSecretWithId(sanitizedSecret.id());
+              } else {
+                keywhizClient.deleteSecretWithId(sanitizedSecret.id(), deleteActionConfig.mode);
+              }
```
mmontgomery-square, line 112: Although we could consolidate to a single call by defaulting to soft deletion, I preserved the original invocation to maintain compatibility with existing test expectations. I'd be more than happy to do the consolidation, however.

cli/src/main/java/keywhiz/cli/commands/DeleteAction.java
```diff
@@ -116,4 +123,14 @@ public DeleteAction(DeleteActionConfig deleteActionConfig, KeywhizClient client)
...
+
+  private static void validateMode(String mode) {
+    if (mode != null && !SecretDeletionMode.values().contains(mode)) {
+      throw new IllegalArgumentException(
+          format(
+              "Invalid mode [%s], must be one of %s.",
+              mode,
+              SecretDeletionMode.values()));
+    }
+  }
```
mmontgomery-square, line 135: The method was starting to get a bit lengthy with multiple blocks, so I opted to do validation via a helper method instead in order to keep things concise.

client/pom.xml
```diff
@@ -29,5 +29,11 @@
       <groupId>io.dropwizard</groupId>
       <artifactId>dropwizard-client</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
```
mmontgomery-square, line 37: There were no unit tests in the client sub-project, so I had to add a JUnit dependency in order to add them.

client/src/main/java/keywhiz/client/KeywhizClient.java
```diff
@@ -272,6 +272,16 @@ public void deleteSecretWithId(long secretId) throws IOException {
...
 
+  public void deleteSecretWithId(long secretId, String mode) {
+    HttpUrl url = baseUrl.newBuilder()
+        .addPathSegment("admin")
+        .addPathSegment("secrets")
+        .addPathSegment(Long.toString(secretId))
+        .addQueryParameter("mode", mode)
+        .build();
+    httpDeleteQuietly(url);
+  }
```
mmontgomery-square, line 283: Rather than String.format(), which does no proper escaping whatsoever, we're building up our URL properly.

client/src/main/java/keywhiz/client/KeywhizClient.java
```diff
@@ -419,4 +429,12 @@ private String httpDelete(HttpUrl url) throws IOException {
...
     return makeCall(request);
   }
+
+  private String httpDeleteQuietly(HttpUrl url) {
+    try {
+      return httpDelete(url);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
mmontgomery-square, line 439: We're also bucking the trend of having every method throw IOException (which is checked and thus annoying to deal with) in favor of wrapping it.

client/src/test/java/keywhiz/client/KeywhizClientTest.java
```diff
@@ -0,0 +1,23 @@
...
+    HttpUrl viaResolve = baseUrl.resolve(String.format("/admin/secrets/%d", secretId));
+    HttpUrl viaBuilder = baseUrl.newBuilder()
+        .addPathSegment("admin")
+        .addPathSegment("secrets")
+        .addPathSegment(Long.toString(secretId))
+        .build();
+
+    assertEquals(viaResolve, viaBuilder);
+  }
+}
```
mmontgomery-square, line 23: This isn't a unit test in the normal sense, but I found it useful to verify that I could construct an identical URL using the proper builder method.

mmontgomery-square:
> LGTM! Though in the future, it would be helpful to include a PR description that summarizes the changes, explains the reason for the change, and links to previous discussions in Slack threads, Jira issues, and design docs (if applicable). Those things tend to be very helpful for future developers looking at old PRs! 👍

Historically we have not included JIRA/etc. links since this repo is public, but obviously we can follow whatever procedure we'd like.

PR #1189, opened 2023-01-20
Adding CLI option to the "describe secrets" action to include deleted secrets

chloebarker:
Usage:

```
$ keywhiz.cli describe secret --name="foo" --include-deleted
```

Added unit tests to test the new behavior.

### DIFF
--- a/cli/src/main/java/keywhiz/cli/Printing.java
+++ b/cli/src/main/java/keywhiz/cli/Printing.java
@@ -154,6 +154,15 @@ public class Printing {
     System.out.println(DOUBLE_INDENT + DateFormat.getDateTimeInstance().format(d));
   }
 
+  public void printDeletedSecretsWithDetails(List<SanitizedSecret> deletedSecrets) {
+    System.out.println("Deleted Secrets:");
+    if (deletedSecrets.isEmpty()) {
+      System.out.println("None found");
+    } else {
+      deletedSecrets.forEach(this::printSanitizedSecretWithDetails);
+    }
+  }
+
   public void printSanitizedSecretWithDetails(SanitizedSecret secret) {
     System.out.println(SanitizedSecret.displayName(secret));
     SecretDetailResponse secretDetails;
@@ -162,6 +171,9 @@ public class Printing {
     } catch (IOException e) {
       throw Throwables.propagate(e);
     }
+
+    System.out.println(INDENT + "Internal Identifier:");
+    System.out.println(DOUBLE_INDENT + secret.id());
 
     System.out.println(INDENT + "Owner:");
     if (secret.owner() != null) {
--- a/cli/src/main/java/keywhiz/cli/Printing.java
+++ b/cli/src/main/java/keywhiz/cli/Printing.java
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import keywhiz.api.ClientDetailResponse;
 import keywhiz.api.GroupDetailResponse;
@@ -154,12 +155,19 @@ public class Printing {
     System.out.println(DOUBLE_INDENT + DateFormat.getDateTimeInstance().format(d));
   }
 
-  public void printDeletedSecretsWithDetails(List<SanitizedSecret> deletedSecrets) {
-    System.out.println("Deleted Secrets:");
-    if (deletedSecrets.isEmpty()) {
-      System.out.println("None found");
-    } else {
+  public void printDeletedSecretsWithDetails(@Nullable List<SanitizedSecret> deletedSecrets) {
+    int deletedCount = deletedSecrets == null ? 0 : deletedSecrets.size();
+    System.out.println(String.format("Deleted Secrets: found %d", deletedCount));
+    if (deletedSecrets != null) {
       deletedSecrets.forEach(this::printSanitizedSecretWithDetails);
+    }
+  }
+
+  public void printNonDeletedSecretWithDetails(@Nullable SanitizedSecret secret) {
+    if (secret == null) {
+      System.out.println("No non-deleted secret found.");
+    } else {
+      printSanitizedSecretWithDetails(secret);
     }
   }
 
@@ -172,7 +180,7 @@ public class Printing {
       throw Throwables.propagate(e);
     }
 
-    System.out.println(INDENT + "Internal Identifier:");
+    System.out.println(INDENT + "Internal ID:");
     System.out.println(DOUBLE_INDENT + secret.id());
 
     System.out.println(INDENT + "Owner:");
--- a/cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
+++ b/cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
@@ -93,35 +93,38 @@ public class DescribeAction implements Runnable {
   }
 
   private void describeSecret() {
-    SanitizedSecret sanitizedSecret = getSanitizedSecret();
-    List<SanitizedSecret> deletedSecrets = describeActionConfig.includeDeleted
-        ? getDeletedSecrets()
-        : null;
-
-    if (sanitizedSecret != null) {
-      printing.printSanitizedSecretWithDetails(sanitizedSecret);
-    }
-
-    if (deletedSecrets != null) {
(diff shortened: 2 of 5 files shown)


cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -154,6 +154,15 @@ public void printGroupWithDetails(Group group) {
...
   }
 
+  public void printDeletedSecretsWithDetails(List<SanitizedSecret> deletedSecrets) {
+    System.out.println("Deleted Secrets:");
+    if (deletedSecrets.isEmpty()) {
+      System.out.println("None found");
+    } else {
+      deletedSecrets.forEach(this::printSanitizedSecretWithDetails);
+    }
+  }
```
mmontgomery-square, line 164: I wonder if it would be helpful to include the number of deleted secrets somewhere in the output (maybe at the end of line 158?).

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
```diff
@@ -83,20 +84,49 @@ public DescribeAction(DescribeActionConfig describeActionConfig, KeywhizClient c
...
-          sanitizedSecret = keywhizClient.getSanitizedSecretByName(name);
-
-          printing.printSanitizedSecretWithDetails(sanitizedSecret);
-        } catch (NotFoundException e) {
-          throw new AssertionError("Secret not found.");
-        } catch (IOException e) {
-          throw Throwables.propagate(e);
-        }
+        describeSecret();
         break;
```
mmontgomery-square, line 88: Yay helper functions

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
```diff
@@ -83,20 +84,49 @@ public DescribeAction(DescribeActionConfig describeActionConfig, KeywhizClient c
...
+        : null;
+
+    if (sanitizedSecret != null) {
+      printing.printSanitizedSecretWithDetails(sanitizedSecret);
+    }
+
+    if (deletedSecrets != null) {
+      printing.printDeletedSecretsWithDetails(deletedSecrets);
+    }
+  }
```
mmontgomery-square, line 108: We might want to push the null checks down a level so that the individual methods are more resilient/defensive and shorten this method up a bit. 

Another thought is that we may want to print out the main secret's details before we attempt to fetch the deleted secrets in case that fetch call is slow. That would allow you to move all of the deleted-related logic within an if check around the includeDeleted flag, keeping it closer together.

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
```diff
@@ -83,20 +84,49 @@ public DescribeAction(DescribeActionConfig describeActionConfig, KeywhizClient c
...
+    } catch (NotFoundException e) {
+      if (!describeActionConfig.includeDeleted) {
+        throw new AssertionError("Secret not found.");
+      }
+      // If we're including deleted secrets, it's ok if no non-deleted secret was found
+      return null;
+    } catch (IOException e) {
+      throw Throwables.propagate(e);
+    }
+  }
```
mmontgomery-square, line 123: Since the result is possibly null, we can better represent that by using an Optional<SanitizedSecret> and not have to worry about null checks. That also allows us to do something fancy in the calling method like

```
maybeSanitizedSecret.ifPresent(printing::printSanitizedSecretWithDetails);
```

The original code throws AssertionError, which is almost certainly wrong. I'd argue we just rethrow the NotFoundException as-is, but that might be annoying if it's checked (and also violates our current API contract).

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
```diff
@@ -83,20 +84,49 @@ public DescribeAction(DescribeActionConfig describeActionConfig, KeywhizClient c
...
+
+  private List<SanitizedSecret> getDeletedSecrets() {
+    try {
+      return keywhizClient.getDeletedSecretsByName(describeActionConfig.name);
+    } catch (IOException e) {
+      throw Throwables.propagate(e);
+    }
+  }
```
mmontgomery-square, line 131: I don't know if getDeletedSecretsByName() can return a null collection, but if so we could hide that by instead returning an empty collection and eliminate the need for null checks in the calling code.

cli/src/main/java/keywhiz/cli/configs/DescribeActionConfig.java
```diff
@@ -28,4 +28,7 @@ public class DescribeActionConfig {
 
   @Parameter(names = "--name", description = "Name of the item to describe", required = true)
   public String name;
+
+  @Parameter(names = { "-d", "--include-deleted" }, description = "Whether to include any deleted secret(s) that match the provided name")
+  public Boolean includeDeleted = false;
```
mmontgomery-square, line 33: Since we are providing a non-null default, I believe this should be a lowercase-b boolean (there's precedent in the existing config classes for this style). And while technically we wouldn't need to specify a default in that case, that is the existing practice, so I'd stick with it.

client/src/main/java/keywhiz/client/KeywhizClient.java
```diff
@@ -350,6 +350,14 @@ public SanitizedSecret getSanitizedSecretByName(String name) throws IOException
     return mapper.readValue(response, SanitizedSecret.class);
   }
 
+  public List<SanitizedSecret> getDeletedSecretsByName(String name) throws IOException {
+    checkArgument(!name.isEmpty());
+    String response =
+        httpGet(baseUrl.resolve(format("/admin/secrets/deleted/%s", name)).newBuilder().build());
+    return mapper.readValue(response, new TypeReference<>() {
+    });
+  }
```
mmontgomery-square, line 359: This class mostly provides numerous bad examples, but we should opt for the HttpUrl.builder()-based approach (which a couple of the methods in the class do use) for building up the URL as opposed to String.format(), which does not perform any of the necessary escaping.

cli/src/main/java/keywhiz/cli/configs/DescribeActionConfig.java
chloebarker: sounds good. looks like I need to learn what the difference between `Boolean` and `boolean` is in Java

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
chloebarker: you mean handling null in the print methods? that would allow us to print something when there's no result, which might be helpful

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
chloebarker: I agree throwing the NotFoundException as is makes sense. the worst case is that the CLI barfs when there's no secret found, which isn't that bad, right?

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
chloebarker: this "everything might be null" thing is so weird to me in a statically typed language

chloebarker: I'm going to try your other suggestion of making the lower-level methods (the print methods) handle nulls, so it's fine if this returns null

client/src/main/java/keywhiz/client/KeywhizClient.java
chloebarker: I think I get what you're saying here. this is what I'm changing it to:

```
String response =
        httpGet(baseUrl.newBuilder()
            .addPathSegment("admin")
            .addPathSegment("secrets")
            .addPathSegment("deleted")
            .addPathSegment(name)
            .build());
```

chloebarker:
Updated based on feedback from @mmontgomery-square

cli/src/main/java/keywhiz/cli/commands/DescribeAction.java
mmontgomery-square: Yeah, the typing in, say, TypeScript where IIRC you can explicitly model the possibility of null via a union type is nice. I think that's one reason why Guava (and Java best practices in general) try to get you to avoid "producing" nulls (whether by replace a null collection with an emtpy one, or with Optional, etc.) so that the code that consumes those values doesn't have to constantly do null checks.

PR #1193, opened 2023-02-01
Fix --include-deleted

chloebarker:
There was an issue with the `--include-deleted` flag I introduced in #1189 where an exception was thrown when a deleted secret was present. This was because the client was expecting a `List<SanitizedSecret>` from the `/admin/secrets/deleted/{name}` endpoint, but the server was sending `List<SecretSeries>`.

This change fixes this by expecting `List<SecretSeries>` and handling it appropriately on the client. Notably, this means that the print method no longer expects a `SanitizedSecret`. Instead, it accepts the `secretID`, and fetches all the information it needs to print using `/admin/secrets/{id}`. It was making this API call anyway, but I had to update the object returned by that call (`SecretDetailResponse`) to include some additional fields that were previously being accessed via the `SanitizedSecret` object. Namely, `checksum`, `expiry`, `contentCreatedAt`, and `contentCreatedBy`.


cli/src/main/java/keywhiz/cli/Printing.java
```diff
@@ -199,64 +202,69 @@ public void printSanitizedSecretWithDetails(SanitizedSecret secret) {
...
-    if (secret.contentCreatedAt().isPresent()) {
+    if (secretDetails.contentCreatedAt().isPresent()) {
       System.out.println(INDENT + "Content created at:");
-      d = new Date(secret.contentCreatedAt().get().toEpochSecond() * 1000);
-      System.out.println(DOUBLE_INDENT + DateFormat.getDateTimeInstance().format(d));
+      System.out.println(DOUBLE_INDENT + apiDateToString(secretDetails.contentCreatedAt().get()));
     }
   }
 
+  private String apiDateToString(ApiDate apiDate) {
```
graysonchao, line 259: :+1:

graysonchao (approved):
nice work, thanks!

PR #1197, opened 2023-02-15
Deleted Secrets - Read from new tables

chloebarker:
### Overview

See [design doc](https://docs.google.com/document/d/1Qfr8kWYVYNKaKcJVC3YjgYRRLTRhH9wvsC4Is9D44IA/edit#). This is Step 3:

Update all places where deleted secrets are read to support both the new and old ways of storing deleted secrets. 

### Details

In the following document, I did a search of keywhiz-server for "delete" and read through all the matches to make sure I wasn't missing anything. You can see the results of that search and my notes on all of the search results here: [Keywhiz code relating to deletion](https://docs.google.com/document/d/1qr7L3hFrZAvAKT_y9bI3y3QkIQGE6SOFoKNbMIo9b60/edit).

Once I identified all the places that needed to be changed, I changed them and updated the unit tests accordingly.


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -119,21 +125,78 @@ long createSecretSeries(
...
-      r.setOptions("{}");
-    }
+    r.setOptions(getOptionsField(generationOptions));
+
     r.store();
 
     return r.getId();
   }
 
+  public long createDeletedSecretSeries(
```
chloebarker, line 135: This only exists for testing, but I can remove it in my next PR. It's helpful right now because there's no other method create a row in the deleted_secrets table. After my next PR, I can put a row in the deleted_secrets table by just creating a secret and then deleting it, so I won't need this method.

I could refactor this so that this creation function is in test code, but that would be a lot more complicated because this function depends on member variables that are instantiated via dependency injection, and I would have to set all that up in my test class. Since I'm about to remove this function anyway, I don't think it's worth it.

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -22,12 +22,17 @@
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import keywhiz.api.model.Group;
+import keywhiz.api.model.Secret;
```
spennymac, line 28: nit: unused import

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -22,12 +22,17 @@
...
 import keywhiz.api.model.SecretSeries;
+import keywhiz.jooq.tables.records.DeletedSecretsRecord;
 import keywhiz.jooq.tables.records.SecretsContentRecord;
 import keywhiz.jooq.tables.records.SecretsRecord;
 import keywhiz.service.config.Readonly;
 import keywhiz.service.crypto.RowHmacGenerator;
+import org.jetbrains.annotations.NotNull;
```
spennymac, line 35: nit: unused import

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
spennymac, line 135:
```suggestion
  @VisibleForTesting
  public long createDeletedSecretSeries(
```

server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -119,21 +125,78 @@ long createSecretSeries(
...
 
+  public long createDeletedSecretSeries(
+      String name,
+      Long ownerId,
+      String creator,
+      String description,
+      @Nullable String type,
+      @Nullable Map<String, String> generationOptions,
+      long now) {
+    long generatedId = rowHmacGenerator.getNextLongSecure();
```
spennymac, line 143: Curious why we aren't just using an  mysql auto-incremented for this?

spennymac: Ah, because we need to know the id before we add the `rowhmac`.

chloebarker: oh yeah I had no idea why it was being done that way!

chloebarker:
@spennymac found a bug, so this is no longer ready for review. I'll update more about the bug and design updates soon. Thanks @spennymac!

chloebarker:
The bug is this:

1. A secret with ID `101` is deleted from the `secrets` table and moved into the `deleted_secrets` table
2. A new secret is created, and the ID `101` is reused. This is unlikely, but possible. The probability for each secret creation is `number_of_deleted_secrets * Math.pow(2, -64)`. As the number of deleted secrets increases and as the number of secret creations over time increases, it becomes more and more likely. 
3. A user attempts to un-delete the secret with ID `101`. There is a key collision.

My solution is outlined [here](https://docs.google.com/presentation/d/1V-yGQ5ZkPuSuHiM_EpbUpdhrndf9So7xPOydYd8OYt4/edit#slide=id.g20e518955da_0_699). Essentially, we will continue to "back up" data from the `secrets` table in `deleted_secrets`, but we will leave the "deleted" row in the `secrets` table as well (as is currently being done anyway), in order to "reserve" the ID.

This has some consequences for how deleted secrets should be read, since newly deleted ones will be present in both tables, and oldly deleted ones will be present only in the `secrets` table. Therefore, I will need to make some changes to the code in this PR.

graysonchao (approved):

PR #1215, opened 2023-04-19
add automation API to get secrets by name+version

graysonchao:
This adds an API that allows getting secret contents for a specific version of the secret. Useful for recovering lost data.

### DIFF
--- a/api/src/main/java/keywhiz/api/automation/v2/SecretContentsAtVersionResponseV2.java
+++ b/api/src/main/java/keywhiz/api/automation/v2/SecretContentsAtVersionResponseV2.java
@@ -1,0 +1,34 @@
+package keywhiz.api.automation.v2;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.List;
+import java.util.Map;
+
+@AutoValue public abstract class SecretContentsAtVersionResponseV2 {
+  SecretContentsAtVersionResponseV2() {} // prevent sub-classing
+
+
+  public static Builder builder() {
+    return new AutoValue_SecretContentsAtVersionResponseV2.Builder().secret("");
+  }
+
+  @AutoValue.Builder public abstract static class Builder {
+    public abstract SecretContentsAtVersionResponseV2.Builder secret(String secret);
+    public abstract SecretContentsAtVersionResponseV2 build();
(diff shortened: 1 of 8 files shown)


cathy-yun-fu (approved):

server/src/main/java/keywhiz/service/daos/SecretDAO.java
```diff
@@ -395,6 +395,28 @@ public Optional<SecretSeriesAndContent> getSecretByName(DSLContext dslContext, S
     return Optional.empty();
   }
 
+  /**
+   * @param name of secret series to look up secrets by.
```
lavanyaharinarayan, line 399: nit: missing param version (it's also not `checkArgument`'d below)

api/src/main/java/keywhiz/api/automation/v2/SecretContentsAtVersionResponseV2.java
```diff
@@ -0,0 +1,35 @@
+package keywhiz.api.automation.v2;
```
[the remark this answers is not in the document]
graysonchao: Nope. At least, these ones aren't... maybe there's a way to autogenerate them but i couldn't find it <_<

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
```diff
@@ -448,6 +450,34 @@ public void partialUpdateSecret_notFound() throws Exception {
   }
 
 
+  //---------------------------------------------------------------------------------------
+  // secretContentsAtVersion
+  //---------------------------------------------------------------------------------------
+
+  @Test public void secretContentsAtVersion_success() throws Exception {
```
lavanyaharinarayan, line 457: Wdyt about a test case that checks that `Optional.empty()` is returned when the secret/version isn't found?

lavanyaharinarayan (approved):

server/src/main/java/keywhiz/service/daos/SecretDAO.java
graysonchao: Added! I went to add a `checkArgument` but I wasn't sure what might constitute an invalid/empty `version`, since it's just a Long.

server/src/test/java/keywhiz/service/resources/automation/v2/SecretResourceTest.java
graysonchao: Can't hurt - I'll add one real quick

graysonchao:
Seems like the docker build has been broken for at least 10 months, so I'm pretty confident this change is not the culprit.

PR #1216, opened 2023-04-21
Implement keywhiz.cli clone

graysonchao:
Adds a new command, `keywhiz.cli clone`, that allows cloning a secret to a new name. This is intended to help resolve an issue where a user needs to recover an old version of their secret without rolling back the existing secret.

### DIFF
--- a/api/src/main/java/keywhiz/api/automation/v2/CloneSecretRequestV2.java
+++ b/api/src/main/java/keywhiz/api/automation/v2/CloneSecretRequestV2.java
@@ -1,0 +1,62 @@
+package keywhiz.api.automation.v2;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import keywhiz.api.validation.ValidBase64;
+
+import javax.annotation.Nullable;
+import java.util.Base64;
+import java.util.Map;
+
+@AutoValue public abstract class CloneSecretRequestV2 {
+  CloneSecretRequestV2() {} // prevent sub-classing
+
+  public static Builder builder() {
+    return new AutoValue_CloneSecretRequestV2.Builder()
+        .name("")
+        .newName("");
+  }
+
+  @AutoValue.Builder public abstract static class Builder {
+    // intended to be package-private
+    abstract CloneSecretRequestV2 autoBuild();
+
+    public abstract Builder name(String name);
+    public abstract Builder newName(String newName);
+
+    /**
+     * @throws IllegalArgumentException if builder data is invalid.
+     */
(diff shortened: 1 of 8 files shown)


api/src/main/java/keywhiz/api/automation/v2/CloneSecretRequestV2.java
```diff
@@ -0,0 +1,62 @@
...
+    public abstract Builder name(String name);
+    public abstract Builder newName(String newName);
+
+    /**
+     * @throws IllegalArgumentException if builder data is invalid.
+     */
+    public CloneSecretRequestV2 build() {
+      // throws IllegalArgumentException if content not valid base64.
+      return autoBuild();
+    }
```
mmontgomery-square, line 37: Is this comment still applicable?

api/src/main/java/keywhiz/api/automation/v2/CloneSecretRequestV2.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  @SuppressWarnings("unused")
+  @JsonCreator public static CloneSecretRequestV2 fromParts(
+      @JsonProperty("name") String name,
+      @JsonProperty("newName") String newName) {
+    return builder()
+        .name(name)
+        .newName(newName)
+        .build();
+  }
```
mmontgomery-square, line 51: I guess while we're on the subject of comments, since @JsonCreator literally means "static factory method for Jackson", we could probably ditch the comment.

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
```diff
@@ -696,6 +697,74 @@ public Response updateCurrentSecretVersion(@Auth User user, @PathParam("secretId
...
+              existingSecret.getSecret(),
+              existingSecret.getCreatedBy(),
+              existingSecret.getExpiry())
+              .withDescription(existingSecret.getDescription())
+              .withMetadata(existingSecret.getMetadata())
+              .withOwnerName(existingSecret.getOwner())
+              .withType(existingSecret.getType().orElse(""))
+              .create();
+      newId = newSecret.getId();
+    } catch (DataAccessException e) {
```
mmontgomery-square, line 736: What about the secret contents? I'm worried that we'll have a new top-level secret with no actual contents/content history pointing to it. Or is this handled magically by the SecretController logic? Or do we not actually want the whole history cloned?

api/src/main/java/keywhiz/api/automation/v2/CloneSecretRequestV2.java
graysonchao: Nope, good call

server/src/main/java/keywhiz/service/resources/admin/SecretsResource.java
graysonchao: Actually you're right, the content history is lost. I had another implementation that used the SecretDAO directly and I think we need to do that + add a DAO method to clone the actual history.

PR #1218, opened 2023-04-27
Removing unnecessary join for expiring secrets listing V4

mmontgomery-square:


server/src/main/java/keywhiz/service/daos/SecretSeriesDAO.java
```diff
@@ -416,18 +416,7 @@ public ImmutableList<SecretSeries> getSecretSeries(@Nullable Long expireMaxTime,
...
-      // examines a large number of rows, causing significant performance
-      // degradation.
-      secretsContentTable = secretsContentTable.useIndexForJoin("secrets_content_expiry");
-    }
-
     SelectQuery<Record> select = baseSelect()
-          .join(secretsContentTable)
-          .on(SECRETS.CURRENT.equal(SECRETS_CONTENT.ID))
           .where(SECRETS.CURRENT.isNotNull())
           .getQuery();
```
mmontgomery-square, line 421: No columns were selected from the secrets_content table, and the join is unnecessary since filtering out secret.current nulls (where current serves as a foreign key to secrets_content) achieves the same result-filtering effect.

PR #1227, opened 2023-07-31
Add support for reserved prefixes

chloebarker:
This allows operators of a keywhiz service to reserve secret names that begin with a specified prefix for a specified owner. For example, we can enforce that a secret beginning with `sp:` can only be created if the ownerName is `specialOwner`.


spennymac:
Is there any concern for existing secrets that match the reserved pattern?

spennymac (approved):

chloebarker:
Good question. The "reserved prefixes" will only be checked at create time, so any existing secrets that match the reserved pattern will not be subject to the reserved prefix rules.

